PR #10, opened 2026-07-31
Absorption phase-01 (Basket B): variables/mode stdlib, variant-matrix component-set

### LINKED ISSUE #1: Absorption phase-01 — Basket B: variables/mode CRUD stdlib, variant-matrix component-set, knowledge recipes
Spec: monorepo plans/260731-1054-capability-absorption/phase-01. Zero new wire commands; ui.* stdlib + knowledge only. Includes fixing our own swallowed CREATE_VARIABLE --mode failure (executor-variables.ts). MIT attribution: THIRD-PARTY.md, one derivation header. Source study: /Users/jang/Products/research/figma-console-mcp.

jangtrinh:
Closes #1.

## Summary
- `ui.vars.*` (rename/remove/describe a variable, add/rename/remove a mode, setModeValue) — every helper verifies its own write, throws E_EVAL on mismatch.
- `ui.componentSet(...)` — variant-matrix or combine-existing COMPONENT_SET builder, adapted from figma-console-mcp's `figma_create_component_set` (MIT, see THIRD-PARTY.md). Deliberately does NOT port the fork's manual rollback bookkeeping — this repo's own `--undo-group` bracket already gives atomic rollback.
- Fixes a pre-existing swallowed failure: `CREATE_VARIABLE --mode <unknown>` used to silently no-op and report success; now throws, naming the modes that exist.
- Two new knowledge recipes (`variables-and-modes.md`, `component-sets.md`) — the analysis recipe explicitly documents and avoids two bugs found in the studied fork (a case-sensitive `state=default` match, and a "guess the main interactive child" heuristic).
- `THIRD-PARTY.md` — MIT attribution table.

## Test plan
- [x] Test-first: the `CREATE_VARIABLE` fix and the component-set verify-mismatch path each have a dedicated regression test, proven to fail against the pre-fix code (revert → confirm fail → restore → confirm green).
- [x] 34 new tests (14 `exec-stdlib-variables`, 17 `exec-stdlib-component-set` + pure-matrix, 3 `executor-variables` regression).
- [x] `npm run build` — clean (cli 228.0kb, plugin 140.0kb)
- [x] `npm test` — 79 files / 1013 tests green (was 79/979 before this branch; +34 explained above, no unexplained drift)
- [x] Panel gate (`kernel/design-os` submodule) — 43/43 green

## NOT done — flagging, not faking
The plan's own Definition of Done asks for a real-canvas live run (build a variant matrix, exercise `ui.vars.*` end-to-end, confirm the `state=Default` case-sensitivity fix on a real set, confirm `--undo-group` rollback on a thrown script). I could not do this from this session: `figma-agent status` on this machine currently shows a broker with **no plugin connected** (`plugins: []`), and driving the Figma Desktop GUI to open/connect one is outside what I can do (no foreground GUI control). Every `[re-verify]` tag in the two new knowledge docs is worded as "unverified, from the fork" per the plan's own fallback instruction (§7) rather than claimed as confirmed. Needs an owner or a session with an actual connected canvas to close out.

Holding for review.

### DIFF
--- a/THIRD-PARTY.md
+++ b/THIRD-PARTY.md
@@ -1,0 +1,64 @@
+# Third-party attribution
+
+This repo absorbs specific capabilities studied from a fork of an existing open-source
+Figma tool, under its MIT license. This file lists exactly what was adapted, from where,
+and how — the citations in `plugin/src/main/` doc comments and `knowledge/` point back
+here.
+
+## Source
+
+- Project: `figma-console-mcp` (Figma Desktop Bridge)
+- Read at: `southleft/figma-console-mcp` (studied locally as
+  `/Users/jang/Products/research/figma-console-mcp`)
+- Version read: **v1.38.2**
+- Read date: **2026-07-31**
+- License: MIT
+
+Verbatim MIT notice from that project's `LICENSE`:
+
+```
+MIT License
+
+Copyright (c) 2025 Figma Console MCP Contributors
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+```
+
+## Derivations
+
+| Our file | Derives from | Nature of derivation |
+|---|---|---|
+| `plugin/src/main/exec-stdlib-component-set.ts` + `exec-stdlib-component-matrix.ts` | `src/core/write-tools.ts:2825-3050` (`figma_create_component_set` tool contract) + `figma-desktop-bridge/code.js:2121-2411` (`CREATE_COMPONENT_SET` handler) | Adapted algorithm + constraint list: the two-mode contract (base+axes vs. combine-existing), the `=`/`,` name-token ban, the 100-combination hard cap and 40-variant size-warning threshold, and the cartesian-product-then-`combineAsVariants` approach. NOT ported: the handler's manual clone/rename rollback bookkeeping (`rollbackComponentSetWork`) — this repo's own `--undo-group` bracket (`executor-exec-js.ts`) already gives atomic rollback over the whole script via Figma's real undo stack, so a second hand-rolled mechanism would duplicate it. Also not ported: `autoArrange`/grid-layout — out of scope for this phase. |
+
+## Facts cited, not derived expression
+
+`knowledge/component-sets.md`'s state→CSS-selector table and its case-sensitivity
+warning are **facts about Figma's own naming convention and about a bug observed in the
+source project's analysis code** (`code.js:1318`, `code.js:1176-1197`), cited for
+re-verifiability — so a reader can go check the claim against the source — not
+reproduced as licensed expression. The table itself is rewritten in this repo's own
+words per `knowledge/component-sets.md` §2.3.
+
+## Attribution already carried in `knowledge/figma-agent-hand.md`
+
+The broker/relay design (websocket-server pending-request correlation, heartbeat
+approach) and the `scripts/probe/visual-diff.mjs` pixel-diff approach were adapted from
+other MIT-licensed projects prior to this repo's own extraction from the `ease-design`
+monorepo — see the top-level `README.md`'s Attribution section for those (pre-existing,
+unrelated to this phase's absorption work).
--- a/knowledge/component-sets.md
+++ b/knowledge/component-sets.md
@@ -32,6 +32,14 @@ return await ui.componentSet({
 });
 ```
 
+- **Run this behind `--undo-group`.** `ui.componentSet` cleans up its OWN mutations
+  (restores a renamed base, removes the clones it created) if one of ITS OWN later
+  steps throws — that is a narrow, self-scoped safety net, not a substitute for the
+  real undo bracket. If a DIFFERENT step in the same script fails after
+  `ui.componentSet` already returned successfully, `--undo-group` is what rolls the
+  whole script back; without it, a successfully-built set stays on the canvas even
+  if a later line in the same script throws. Always wrap a script that calls this
+  in `exec-js --undo-group` unless you specifically want partial results to survive.
 - **Exactly one of `base`+`axes` or `components`** — both or neither throws, naming the
   two valid shapes.
 - **Axis and value names must not contain `=` or `,`** — Figma parses variant properties
--- a/knowledge/component-sets.md
+++ b/knowledge/component-sets.md
@@ -1,0 +1,127 @@
+# Component sets — build from a matrix, analyse an existing one
+
+Two recipes (absorption phase-01, Basket B): `ui.componentSet(...)` builds a
+COMPONENT_SET from a variant matrix or from existing components; the analysis recipe
+below composes tools this repo already has (`figma-agent scan-node`, `ui.q`,
+`ui.componentSet`'s own `propertyDefinitions`) to answer the same question the fork's
+`figma_analyze_component_set` does — **without** its case-sensitivity bug or its
+"guess the main interactive child" heuristic (both explained below, so neither comes
+back by accident).
+
+All snippets are `figma-agent exec-js` ready (async body, `ui`/`figma` globals, `return`
+a JSON-safe value). Companion: `knowledge/figma-craft/components-variables-styles.md`
+§1.3 (the `combineAsVariants` naming contract this builds on).
+
+---
+
+## 1. Build a set from a variant matrix
+
+```js
+// Mode 1 — generate from a base COMPONENT + axes (cartesian product):
+return await ui.componentSet({
+  base: '12:34',
+  axes: { State: ['default', 'hover', 'disabled'], Size: ['sm', 'lg'] },
+});
+// → 6 variants. The BASE keeps its node id (variants[0].id === '12:34'), so any
+//   existing instance of the base survives as an instance of that first variant.
+
+// Mode 2 — combine existing COMPONENT nodes, naming them via variantProps:
+return await ui.componentSet({
+  components: ['1:1', '1:2', '1:3'],
+  variantProps: [{ State: 'default' }, { State: 'hover' }, { State: 'disabled' }],
+});
+```
+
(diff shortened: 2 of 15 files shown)


jangtrinh (commented):
**Review verdict: REQUEST CHANGES** (posted as a review comment — GitHub refuses a formal request-changes on a self-authored PR.)

Reviewed on a fresh clone of `absorption-phase-01-basket-b`; re-ran the three new test files locally (34/34 green) and spot-checked the upstream citations in `THIRD-PARTY.md` against the referenced source.

The shape of this work is right: every `ui.vars.*` helper re-reads its own write, the `CREATE_VARIABLE --mode` swallowed failure is fixed with a regression test, the variant matrix is deterministic, and both bugs found in the studied project are documented in the recipe *and* absent from the ported code. Two issues need a change before merge.

## Request changes

**1. `setModeValue` cannot set a variable alias — it throws `E_EVAL` on a write that actually succeeded.** `plugin/src/main/exec-stdlib-variables.ts:158-166`

```ts
const matches = variable.resolvedType === 'COLOR'
  ? valuesEqual(actual, value as VariableValue)
  : actual === value;
```

For non-`COLOR` types this is reference equality, and `valuesEqual` (`plugin/src/main/executor-variables.ts:28-36`) also falls back to `a === b` for any object that is not an RGBA. `setValueForMode(modeId, { type: 'VARIABLE_ALIAS', id })` — the documented way to point one variable at another, i.e. the whole semantic→primitive token layer — reads back as a *new* object, so the verification fails and the helper throws even though the canvas is correct.

This is invisible in tests because the mock stores the caller's object by reference (`tests/helpers/mock-figma.ts:751-753`), so `actual === value` is trivially true there, and no test passes an alias value. Fix: teach the comparator about `VARIABLE_ALIAS` (compare `type` + `id`), have the mock store a structural copy of the value, and add an alias round-trip test.

**2. The "no manual rollback, `--undo-group` covers it" divergence has no caller-facing contract — and the helper's own verification is what creates the orphans.** `plugin/src/main/exec-stdlib-component-set.ts:60-84` renames the base component and clones it N-1 times; `:170-183` then throws `E_EVAL` if the combined children do not parse back to the intended axes. Without `--undo-group`, that throw leaves the clones on the canvas and the base renamed to `Axis=Value, …`, with nothing to clean it up.

The requirement is stated only in a source comment (`:1-10`). `knowledge/component-sets.md` — the surface an agent actually reads before calling this — does not mention `--undo-group` anywhere. Worth noting that the studied project ports a rollback for exactly this window (its bridge defines `rollbackComponentSetWork` and invokes it when the combine did not complete), so the divergence is only sound if the bracket is guaranteed.

Suggested fix, in two parts: (a) state the `--undo-group` requirement in the recipe and in the helper's own doc comment/result; (b) on the helper's *own* verification failure, reverse the helper's *own* mutations before throwing — remove the clones it created and restore the base name. That is a compensating action for a self-inflicted failure, not a second undo system, and it keeps the divergence honest for callers who forget the flag.

## Minor

3. **The mock encodes refusals for `removeMode`/`renameMode` but not for `addMode` or `combineAsVariants`.** `tests/helpers/mock-figma.ts:718-724` never throws, so the plan-cap catch at `plugin/src/main/exec-stdlib-variables.ts:96-101` is never exercised; `combineAsVariants` (`tests/helpers/mock-figma.ts:833-859`) likewise accepts an empty list, a non-`COMPONENT` node, or a node already inside a set. Encoding the refusals you depend on is what makes this mock trustworthy — `removeMode` refusing the last mode is exactly the right precedent to extend.

4. **Dead field in a public result shape.** `remove()` declares `boundReferences?: number` (`plugin/src/main/exec-stdlib-variables.ts:70`, mirrored at `:180`) but nothing can ever populate it. `boundReferencesChecked: false` is the honest signal; drop the optional count until a file-wide reference reader exists, so the shape does not imply a capability.

5. **Silent truncation in an error message.** `resolveCollection` (`plugin/src/main/exec-stdlib-variables.ts:20-22`) lists at most 20 candidates with no "+N more", so on a large file the caller cannot tell the list is partial. Same for the mode lists, which are unbounded — one of the two conventions should win.

6. **`resolveParent` is too permissive.** `plugin/src/main/exec-stdlib-component-set.ts:44-50` accepts any node with `appendChild`, so a `COMPONENT` or `INSTANCE` id passes the guard and fails later inside `combineAsVariants` — i.e. after the clones already exist (issue 2's window). Restrict to the container types that can actually hold a component set.

## Verified sound

- Every `ui.vars.*` helper re-reads its own write and throws `E_EVAL` on mismatch; `renameMode` re-reads by `modeId` and `setModeValue` indexes `valuesByMode` by the resolved mode's own id, so the mode-targeting is correct (only the *comparison* is wrong — issue 1).
- `remove()` reporting `boundReferencesChecked: false` instead of a page-scoped count that would under-report on a multi-page file: the right call, documented with its upgrade trigger.
- The `CREATE_VARIABLE --mode` fix (`plugin/src/main/executor-variables.ts:181-194`) throws and names the modes that exist, matching `setModeValue`'s contract; the paired regression test is real.
- Matrix core (`plugin/src/main/exec-stdlib-component-matrix.ts`): deterministic cartesian order (axis insertion order, values in given order), `=`/`,` rejected *before* anything is created, `> 100` rejects and `> 40` warns with no off-by-one at either boundary.
- Post-combine verification compares child count *and* each parsed axis map, and mode-2-without-`variantProps` deliberately asserts only survival rather than a fabricated axis map — the right restraint.
- Both bugs attributed to the studied project are documented in `knowledge/component-sets.md:100-127` (case-sensitive `state=default` match; the "guess the main interactive child" heuristic) and neither pattern appears in the ported code: nothing lowercases against a literal template, and no such heuristic exists.
- `THIRD-PARTY.md` citations line up with the referenced source, including the rollback function the table says was deliberately not ported, and the facts-vs-expression distinction is drawn correctly.
- The `[re-verify]` tags read as unverified rather than confirmed, and the missing live-canvas run is flagged in the PR body instead of quietly claimed — that is the correct way to leave this open.

## Question

`ui.componentSet` mode 1 mutates the base component's name before any clone exists. If `combineAsVariants` fails for an unrelated reason (bad parent, plan limit), is leaving the base renamed acceptable, or should the helper snapshot and restore it independently of issue 2's fix?

jangtrinh:
Pushed `fa5b948` addressing the review.

## Fixed

**1. `setModeValue` couldn't set a variable alias.** The equality check fell back to reference comparison for anything that wasn't a color, and the test mock happened to store values by reference — so no test could ever see the bug, even though a real Figma write would fail every time. Both are fixed: `valuesEqual` now compares alias objects by their target id, and the mock stores a real copy of whatever you pass it (matching how Figma actually persists a value). Added a test that sets a variable to reference another variable and confirmed it fails against the old code before the fix.

**2. `componentSet` could leave a half-built mess if something failed partway through.** Building a variant set from a single base component works by renaming that component and cloning it — and until this fix, if anything went wrong afterward (a bad parent, a naming mismatch), the base stayed renamed and the clones stayed on the canvas. Now the function tracks exactly what it changed and undoes just that if it fails, independent of *why* it failed. This is deliberately narrow — it's not a general-purpose undo system, and the docs now say plainly that wrapping the call in `--undo-group` is still how you protect the rest of a script.

## Also cleaned up

- Removed a result field (`boundReferences`) that could never actually hold a value.
- Error messages listing available collections/modes now say "+N more" instead of silently cutting off after 20.
- Fixed a spot where passing the wrong kind of node as a parent would fail late (after mutations already happened) instead of upfront.
- The test mock now enforces a few more real Figma constraints (component-set membership, a configurable mode-count cap) so the code paths that depend on them actually get exercised.

## Verification
- 9 new tests (1022 total, up from 1013), including 3 written first and confirmed to fail against the pre-fix code before the fix landed.
- typecheck, compile, and the full suite (incl. the 43-test panel gate) all pass.

Ready for re-review.

jangtrinh (commented):
**Review verdict: APPROVE** — both issues from the previous pass are closed and independently verified. (Posted as a review comment: GitHub refuses a formal approval on a self-authored PR.)

Re-reviewed at `fa5b948` on a fresh clone. Both issues from the first pass are properly closed, and I verified the alias fix the hard way rather than trusting the test name.

## Verified

**Issue 1 — alias write (fixed, revert-proven independently).** `valuesEqual` now compares every real `VariableValue` shape structurally (`plugin/src/main/executor-variables.ts:36-55`): RGBA with epsilon, `VARIABLE_ALIAS` by target id, primitives strict. `setModeValue` uses it for all types, not just `COLOR` (`plugin/src/main/exec-stdlib-variables.ts:174-178`). To confirm the new test actually pins the behaviour, I replaced `valuesEqual` with its pre-fix body in my clone and re-ran: exactly the new alias test fails (1 failed / 16 passed), then passes again once restored. The mock's `setValueForMode` now stores a structural clone (`tests/helpers/mock-figma.ts:771-778`), which is what makes a reference-equality bug catchable at all — the right lesson from the earlier permissive-mock problem, applied to the mock's own write path.

**Issue 2 — self-cleaning build (fixed).** `buildModeA`/`buildModeB` return a `cleanup()` that reverses exactly their own mutations (`plugin/src/main/exec-stdlib-component-build.ts:44-63`, `:107-113`), and `componentSet` wraps parent resolution → combine → verification in one `try`, calling `cleanup()` and rethrowing (`plugin/src/main/exec-stdlib-component-set.ts` catch block). The rethrow is `throw err` on the original object, so the `code` and stack survive — I checked that specifically. Both failure shapes are covered by observable tests through a real frame parent (verify-mismatch and an unrelated `combineAsVariants` throw), and mode 2 restores every `variantProps` rename. `knowledge/component-sets.md:35-42` now states the `--undo-group` requirement and why the self-cleanup is not a substitute.

**Minors** — `boundReferences` dropped; `listWithMore()` applied to both collection and mode lists with an explicit `+N more`; `resolveParent` restricted to a type allowlist with a test proving a `COMPONENT` parent is refused; the mock encodes `combineAsVariants` refusals (empty list, non-`COMPONENT`, already in a set) and `setMockModeCap` exercises the `addMode` plan-limit catch that previously had no path through it.

Test files re-run in my clone: 43/43 green across the three changed suites, and the split leaves `exec-stdlib-component-set.ts` at 127 lines with the build half at 115.

## Non-blocking follow-ups

1. **`cleanup()` can mask the original error.** The catch does `build.cleanup(); throw err;` with no guard, so if cleanup itself throws, the caller sees the cleanup failure instead of the real one. Real Figma throws when `.remove()` is called on an already-removed node, and when a property is set on one — while the mock's `remove()` (`tests/helpers/mock-figma.ts:218-226`) tolerates both, so `BuildResult.cleanup`'s "safe to call multiple times" (`plugin/src/main/exec-stdlib-component-build.ts:19`) is a claim about the mock rather than the canvas. Suggest a per-action `try`/`catch` inside `cleanup()` (and either a `node.removed` guard or a narrower comment).

2. **The cleanup boundary is right, but the recipe overstates it.** On the verification-mismatch path, `combineAsVariants` has already succeeded — so the new COMPONENT_SET survives with the base inside it, name restored but parent changed. `tests/exec-stdlib-component-set.test.ts:19-24` documents that precisely and I agree with the scope (un-combining is genuinely `--undo-group`'s job). But `knowledge/component-sets.md:35-42` describes the net as "restores a renamed base, removes the clones it created", which reads as full restoration to someone deciding whether they need the bracket. One sentence — "a set that was already combined stays; only the bracket removes that" — closes it.

3. **Failures *inside* the build still leave renames.** `buildModeB` resolves, validates and renames per index (`plugin/src/main/exec-stdlib-component-build.ts:78-104`), so `components[3]` failing validation leaves 0-2 renamed with no cleanup closure in existence yet. A first pass that resolves and type-checks every id before the first rename removes the window entirely.

4. **A cited convention that isn't written down.** Both new headers reference "this repo's 200-line cap", but neither `CLAUDE.md` nor `README.md` states it, and `exec-stdlib-variables.ts` is now 205 lines. Either record the rule or drop the citation, so the next split isn't justified by a number nobody can look up.

*Informational:* `valuesEqual`'s final fallback is `JSON.stringify` equality, which is key-order sensitive. Unreachable for the shapes a `VariableValue` can actually take — the comment says as much — but worth remembering if that union ever widens.

The live-canvas run remains the open item from the PR body, and it is still flagged rather than claimed. Approving on the code; that run is the thing to close before relying on the `[re-verify]` claims.

PR #12, opened 2026-07-31
fix(broker): atomic advertisement writes, guarded self-heal, probe before kill

### LINKED ISSUE #5: Broker hardening — atomic port-file, heartbeat self-heal, probe-before-kill
From mining report (plans/reports/scout-260731-1047): temp-file+rename advertisement writes, self-healing heartbeat re-advertise, liveness probe before orphan kill; CHECK our probe for the IPv6 localhost-vs-127.0.0.1 bug they fixed in v1.38.2. Backlog 1.7.

jangtrinh:
Closes #5.

## What changed

The broker advertises itself at `/tmp/figma-agent-broker.json` so every CLI
invocation and the Figma plugin can find it. Three related hardening fixes,
from the mining report on southleft/figma-console-mcp's port-discovery
module (`plans/reports/scout-260731-1047-figma-console-mcp-mining.md` in the
opah workspace):

**1. Atomic advertisement writes.** The file was written with a plain
`writeFileSync`, which truncates before writing. Any reader racing a rewrite
(the 30s heartbeat, `ensureBroker`, `figma-agent status`) could observe an
empty or partial file mid-write and misread a perfectly healthy broker as
dead/corrupt. Fixed by writing to a pid-scoped temp file and `renameSync`-ing
onto the target — atomic on POSIX, so a reader only ever sees the complete
old file or the complete new one.

**2. Guarded heartbeat self-heal.** The refresh interval already rewrote the
advertisement unconditionally every tick, which incidentally self-healed the
file if it vanished while the broker was still alive. But nothing stopped
that same interval from resurrecting the file *after* the broker had
deliberately shut down (relevant in the test harness, where the injected
`exit` stub throws rather than truly halting the process — a stand-in for
any future path where shutdown doesn't immediately kill the event loop).
Added an in-process `ownsAdvertisement` flag, flipped false before
`shutdown()` unlinks the file, checked at the top of the refresh tick.

**3. Probe before kill.** Two paths decided an advertised broker was
dead/stale without probing liveness first:
   - `ensureBroker()` fell straight through to spawning a competing broker
     on any stale heartbeat, even when the owning pid was still alive (just
     slow — GC pause, a big EXEC_JS job).
   - `runCommand()` deleted the advertisement file after a single failed WS
     connect, with no check for whether the pid was actually dead.

   Both are a split-brain risk: a second broker starts while the first is
   still alive, the newer one takes over the advertisement, and the older
   one is silently orphaned while still holding the Figma plugin's live WS
   connection. Added `probeStaleButAlive` (pid check, then a handshake
   connect, before trusting a stale-looking advertisement) and
   `isConfirmedDeadAfterFailedConnect` (pid check before deleting the file;
   a live pid gets one retry instead of an immediate teardown).

**IPv6 host-agreement check.** Verified the bind and connect/probe host
strings already agree — both used the `127.0.0.1` literal, never the
ambiguous `'localhost'` that caused the fork's actual v1.38.2 bug (Node
resolves `'localhost'` to the IPv6 loopback for some socket calls and IPv4
for others on a dual-stack macOS box, so a bind and a probe can silently
land on different address families). Centralized the literal behind a
`LOOPBACK_HOST` constant + `loopbackWsUrl()` helper and added a
regression-lock test that fails if any transport file reintroduces a bare
`'localhost'` string.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run build`
- [x] `npm test` — 1039/1039 passing, including 3 new test files and one
      extended describe block in `tests/broker-daemon-harness.test.ts`
- [x] Test-first verified: stashed the source changes and re-ran the suite —
      all 12 new/extended test cases failed against pre-fix code (see PR
      description commits/CI for the failure log), confirming they exercise
      real behavior differences, not tautologies.
- [x] Re-ran the full suite 3x locally to check for flakiness in the new
      timing-sensitive self-heal test — stable each time.

## Deviations from the issue text (flagged, not improvised)

- The advertisement-refresh interval's cadence was hardcoded to the raw
  `HEARTBEAT_INTERVAL_MS` constant (30s, un-overridable), unlike every other
  interval in the daemon which already reads an env-overridable local
  const. This made the self-heal behavior untestable in reasonable time, so
  I switched it to the same `FIGMA_AGENT_HEARTBEAT_MS`-derived `HEARTBEAT_MS`
  the WS-ping interval already uses. Default production cadence is
  unchanged (falls back to the same 30s constant); only test runs override
  it.
- Introduced two new pure predicate functions (`probeStaleButAlive` in
  `broker-discovery.ts`, `isConfirmedDeadAfterFailedConnect` in
  `broker-client.ts`) rather than testing `ensureBroker()`/`runCommand()`
  directly — both of those already hardcode the real, shared
  `/tmp/figma-agent-broker.json` path and can spawn real detached
  processes, so they aren't safely test-injectable. This follows the
  existing convention in this codebase (`refusesReadOnlyAssertion` in
  `broker-client.ts` is the same pattern: extract the decision as a pure,
  DI'd predicate; keep the impure wrapper thin).

## Not verified

- No live Figma plugin round-trip (no Figma Desktop instance available in
  this environment). The change is broker/CLI-side only and does not touch
  `plugin/src` or the wire protocol shape, so the existing panel-gate tests
  (which do exercise the plugin bundle) are the closest available signal.
- Real split-brain / IPv6 dual-stack behavior was not reproduced live (would
  require deliberately racing two broker processes or a genuinely
  dual-stack-ambiguous host) — covered instead by the pure-predicate and
  regression-lock tests described above.

### DIFF
--- a/cli/src/transport/broker-client.ts
+++ b/cli/src/transport/broker-client.ts
@@ -30,6 +30,15 @@ import {
 } from './protocol-helpers.ts';
 
 const CONNECT_TIMEOUT_MS = 4_000;
+// A failed connect that isn't confirmed-dead (see isConfirmedDeadAfterFailedConnect)
+// is treated as "broker was mid-accept or busy on another handshake" — retrying
+// IMMEDIATELY re-attempts inside the exact window that caused the first failure.
+// This delay is what actually gives that busy state room to clear.
+const AMBIGUOUS_CONNECT_RETRY_DELAY_MS = 200;
+
+function sleep(ms: number): Promise<void> {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
 let requestCounter = 0;
 
 let expectedFile: string | undefined;
@@ -223,6 +232,27 @@ export function isConfirmedDeadAfterFailedConnect(
 }
 
 /**
+ * After an ambiguous (not confirmed-dead) connect failure, wait a short backoff
+ * then retry the connect once. Stage-4 review round (minor) — the retry used to
+ * fire immediately, re-attempting inside the exact "broker mid-accept / busy on
+ * another handshake" window that caused the first failure; the backoff is what
+ * actually gives that window room to clear. Exported with injectable `connect`/
+ * `sleep` (defaulting to the real ones) so the backoff-then-retry ordering is
+ * unit-testable without a real broker or the real shared advertisement file.
+ */
+export async function retryAmbiguousConnect(
+  port: number,
+  deps: { connect: (port: number) => Promise<WebSocket>; sleep: (ms: number) => Promise<void>; delayMs?: number } = {
+    connect: connectWs,
+    sleep,
+    delayMs: AMBIGUOUS_CONNECT_RETRY_DELAY_MS,
+  },
+): Promise<WebSocket> {
+  await deps.sleep(deps.delayMs ?? AMBIGUOUS_CONNECT_RETRY_DELAY_MS);
+  return deps.connect(port);
+}
+
+/**
  * Run one command through the broker. Connect failure after a valid
  * advertisement forces one rediscovery (broker may have just died).
  *
@@ -259,9 +289,9 @@ export async function runCommand(
     if (!isConfirmedDeadAfterFailedConnect(ad)) {
       // Probe before kill: the pid is still alive, so the first failure is
       // ambiguous rather than confirmed-dead — give the handshake one more try
-      // before concluding the advertisement lied.
+      // (after a short backoff) before concluding the advertisement lied.
       try {
-        ws = await connectWs(ad.port);
+        ws = await retryAmbiguousConnect(ad.port);
       } catch {
         throw new CliError(
           'E_NO_BROKER',
--- a/cli/src/transport/broker-client.ts
+++ b/cli/src/transport/broker-client.ts
@@ -10,12 +10,13 @@ import {
   PROTOCOL_VERSION,
   makeRequestFrame,
   makeRequestId,
+  type BrokerAdvertisement,
   type CommandName,
   type FileContext,
   type JobInfo,
   type WireError,
 } from '../../../shared/protocol.ts';
-import { ensureBroker } from './broker-discovery.ts';
+import { ensureBroker, isPidAlive, loopbackWsUrl } from './broker-discovery.ts';
 import { MUTATING_COMMANDS } from '../../../shared/mutating-commands.ts';
 import {
   ChunkAssembler,
@@ -73,7 +74,7 @@ export function refusesReadOnlyAssertion(wireCmd: CommandName, readOnly: boolean
 
 function connectWs(port: number): Promise<WebSocket> {
   return new Promise((resolve, reject) => {
-    const ws = new WebSocket(`ws://127.0.0.1:${port}`);
+    const ws = new WebSocket(loopbackWsUrl(port));
     const timer = setTimeout(() => {
       ws.terminate();
       reject(new CliError('E_NO_BROKER', `broker connect timed out on port ${port}`));
@@ -182,7 +183,7 @@ export function exchange(
  */
 export function fetchBrokerHello(port: number): Promise<Record<string, unknown>> {
   return new Promise((resolve, reject) => {
-    const ws = new WebSocket(`ws://127.0.0.1:${port}`);
+    const ws = new WebSocket(loopbackWsUrl(port));
     const timer = setTimeout(() => {
       ws.terminate();
       reject(new CliError('E_NO_BROKER', `broker hello timed out on port ${port}`));
@@ -199,6 +200,26 @@ export function fetchBrokerHello(port: number): Promise<Record<string, unknown>>
     ws.once('error', (err) => done(() => reject(new CliError('E_NO_BROKER', `broker hello failed: ${err.message}`))));
     ws.once('close', () => { clearTimeout(timer); reject(new CliError('E_NO_BROKER', 'broker closed before hello')); });
   });
+}
+
+/**
+ * Whether a WS connect failure against an advertised broker justifies deleting its
+ * advertisement file and spawning a replacement. A single failed connect is not
+ * proof of death — it can also mean the broker was mid-accept, busy on another
+ * handshake, or hit a transient loopback hiccup. Deleting the file and spawning a
+ * SECOND broker while the first is still alive is a split-brain: two live daemons,
+ * only the newer one advertised, the older one silently orphaned while still
+ * holding the Figma plugin's live WS connection. `isPidAliveFn` is the cheap,
+ * network-independent signal used to gate that: a dead pid can never have caused
+ * the failure by being "merely busy", so deletion is safe; a live pid means the
+ * failure is ambiguous, so `runCommand` retries once before giving up rather than
+ * tearing the broker down.
+ */
+export function isConfirmedDeadAfterFailedConnect(
+  ad: BrokerAdvertisement,
+  isPidAliveFn: typeof isPidAlive = isPidAlive,
+): boolean {
+  return !isPidAliveFn(ad.pid);
 }
 
 /**
@@ -235,13 +256,27 @@ export async function runCommand(
   try {
     ws = await connectWs(ad.port);
   } catch {
-    try {
-      unlinkSync(BROKER_FILE); // advertisement lied — drop it and respawn
-    } catch {
-      /* already gone */
+    if (!isConfirmedDeadAfterFailedConnect(ad)) {
+      // Probe before kill: the pid is still alive, so the first failure is
+      // ambiguous rather than confirmed-dead — give the handshake one more try
+      // before concluding the advertisement lied.
+      try {
+        ws = await connectWs(ad.port);
+      } catch {
+        throw new CliError(
+          'E_NO_BROKER',
+          `broker (pid ${ad.pid}) is alive but not answering on port ${ad.port} — see /tmp/figma-agent-broker.log`,
+        );
+      }
+    } else {
+      try {
+        unlinkSync(BROKER_FILE); // pid confirmed dead — advertisement lied, drop it and respawn
+      } catch {
+        /* already gone */
+      }
+      ad = await ensureBroker();
+      ws = await connectWs(ad.port);
     }
-    ad = await ensureBroker();
-    ws = await connectWs(ad.port);
   }
 
   const timeoutMs = opts?.timeoutMs ?? COMMAND_TIMEOUTS[wireCmd] ?? DEFAULT_TIMEOUT_MS;
--- a/cli/src/transport/broker-daemon.ts
+++ b/cli/src/transport/broker-daemon.ts
@@ -15,9 +15,11 @@ import {
   type BrokerAdvertisement, type ErrorCode, type EventMsg, type JobInfo, type ReplyErr, type ReplyOk, type RequestMsg,
   type WireMsg,
 } from '../../../shared/protocol.ts';
-import { isPidAlive, readAdvertisement, selfBuildMtime, writeAdvertisement } from './broker-discovery.ts';
 import {
-  ChunkAssembler, deleteConnectionChunk, getConnectionChunks, isChunkMsg, isEventMsg, isReplyMsg, isRequestMsg,
+  cleanupOrphanedAdvertisementTempFiles, isPidAlive, readAdvertisement, selfBuildMtime, writeAdvertisement,
+} from './broker-discovery.ts';
+import {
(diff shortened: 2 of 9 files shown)


jangtrinh (commented):
**Review verdict: REQUEST CHANGES** — one regression to fix (the new stale-but-alive path skips the build/protocol gate); everything else in the PR is sound and verified. (Posted as a review comment: GitHub refuses a formal request-changes state on a self-authored PR.)

Reviewed `ea79bd8` on a fresh clone; ran the four affected test files twice (23/23 both times) and checked the claims against the call sites rather than the descriptions.

The three fixes are real and the reasoning behind each is sound — atomic writes with no truncate path left, the ownership flag flipped before the unlink rather than after, and both destructive paths genuinely gated by their new predicates. One regression needs fixing before merge.

## Request changes

**The new stale-but-alive path bypasses the protocol/build gate.** `cli/src/transport/broker-discovery.ts:219-229`

```ts
if (ad && isAdvertisementLive(ad)) {
  const outdatedBuild = myMtime - ad.buildMtime > 1;
  if (ad.protocolV === PROTOCOL_VERSION && !outdatedBuild) return ad;
  await requestBrokerShutdown(ad); // stale build / protocol → replace
} else if (ad && (await probeStaleButAlive(ad))) {
  return ad;                       // ← no version check on this path
}
```

The live path replaces a broker running an outdated build or a mismatched protocol; the new path returns it unconditionally. Pre-fix, a heartbeat-stale advertisement fell through to `spawnBroker()`, so a rebuilt CLI always got a matching broker. Post-fix it reuses whatever is listening.

This is reachable through the most ordinary path there is. A healthy broker refreshes `lastSeen` every 30s regardless of traffic, so "stale" means the refresh interval was starved past ~120s — i.e. the machine slept, or a long `EXEC_JS` blocked the loop. Close the lid, rebuild while it is closed, reopen: the pid is alive, the handshake answers, and the CLI now talks to the pre-rebuild broker with no signal that it did. Split-brain avoidance is the right goal, but it should not cost the hot-replace guarantee.

Fix: apply the same gate inside the new branch — reuse only when `protocolV` matches and the build is not outdated, otherwise `await requestBrokerShutdown(ad)` and fall through to `spawnBroker()`. That keeps the probe's benefit (no competing spawn on top of a live, current broker) without silently pinning callers to a stale one.

## Non-blocking

1. **The retry has no backoff.** `cli/src/transport/broker-client.ts:259-270` — when the pid is alive, the code immediately re-attempts `connectWs(ad.port)`. The scenario this retry exists for is "the broker was mid-accept or busy on another handshake", which is exactly the state where a second connect issued microseconds later also fails. A short delay (100-250ms) before the retry would make it do the work its own comment describes.

2. **Temp files can orphan on a hard kill.** `cli/src/transport/broker-discovery.ts:87-97` writes `${path}.${process.pid}.tmp` and unlinks it on a thrown write, but a `SIGKILL` between the write and the `renameSync` leaves it behind. Bounded (pid-scoped, so a recycled pid overwrites it) and harmless, but nothing ever sweeps them — a one-line unlink of `${BROKER_FILE}.*.tmp` at daemon startup would close it.

3. **`isAdvertisementLive` still hardcodes the old cadence.** `broker-discovery.ts:73` allows `HEARTBEAT_STALE_MS + 30_000`, where the `30_000` is the refresh interval that this PR just made env-overridable. Today's default keeps them in agreement, but the slack no longer derives from the cadence it is meant to tolerate. Deriving it from the same source (or a comment pinning the assumption) keeps the two from drifting apart later.

4. **The regression lock's file list is hand-maintained.** `tests/broker-loopback-host.test.ts:22-27` enumerates three transport files, so a new one that builds a socket URL is not covered. Also worth a sentence in that test: the plugin side deliberately connects to a `.localhost` hostname (`plugin/src/ui/ui-relay.ts:353-361`, because Figma Desktop rejects a literal IP in the manifest), which is an intentional exception rather than an oversight — otherwise the next reader may "fix" it into the very bug this test guards.

## Verified

- **No truncate path left.** `writeAdvertisement` is the only writer and it goes through `writeFileAtomic`; grep finds no other `writeFileSync` against the advertisement.
- **Atomic rename does not break any reader.** `readAdvertisement`, `figma-agent status`, and `shutdown()`'s own ownership read all take whole files; with `rename(2)` they see either the complete old or the complete new one, which is precisely the window that previously misread a healthy broker as dead.
- **`ownsAdvertisement` is placed correctly.** Flipped false at the top of `shutdown()` *before* the unlink (`broker-daemon.ts:446-452`) and checked as the first statement of the refresh tick (`:1330`), so a cleanly-exited broker cannot resurrect its own file, while a live broker whose file vanished still re-advertises on the next tick. The harness test proves both directions on one run.
- **Both predicates gate the destructive action, not just exist.** `probeStaleButAlive` sits on the branch that decides whether to spawn a competitor, and checks `isPidAlive` before spending a network probe; `isConfirmedDeadAfterFailedConnect` guards the `unlinkSync(BROKER_FILE)` and diverts a live-pid failure into a retry plus a named error (`pid`, port, log path) rather than a teardown.
- **Cadence deviation is genuinely inert in production.** `HEARTBEAT_MS = envMs('FIGMA_AGENT_HEARTBEAT_MS', HEARTBEAT_INTERVAL_MS)` keeps the 30s default, and nothing else reads `HEARTBEAT_INTERVAL_MS` for cadence — the plugin has its own `PLUGIN_HEARTBEAT_INTERVAL_MS`.
- **Host agreement holds.** No quoted `'localhost'` remains in the three transport files (the only hits are prose in comments, which the lock's regex correctly ignores), and bind, connect, probe and the daemon's log line all read `LOOPBACK_HOST`.
- **The probe tests discriminate both ways** — dead pid never reaches the network probe, live-and-answering is trusted, live-but-unresponsive is not, and both directions of `isConfirmedDeadAfterFailedConnect` are asserted, with two real-socket cases for listening vs refused.
- **Simultaneous starts are safe:** two brokers bind different ports, the last rename wins, and the loser's next tick sees a different live pid and shuts down — with the ownership checks preventing it from deleting the winner's file. Worth knowing that the loser can keep serving for up to one cadence, but that is pre-existing and unrelated to this change.

**On the timing-sensitive test:** I judge it stable rather than the load-sensitive class. The heal wait is a 1s budget against a 150ms cadence (needs 1 of ~6 ticks), and the post-shutdown assertion is that the file stays *absent* — under a starved event loop, fewer ticks fire, which pushes that assertion toward passing rather than flaking. No guard needed.

jangtrinh:
Addressed in `4cd004c`.

**Regression (blocking) — fixed.** `ensureBroker()`'s stale-but-alive branch now goes through the identical protocol and build-mtime gate as the live path, with the identical consequence. Extracted the whole decision into a pure `decideBrokerAction(ad, myMtime, deps)` so it's unit-testable without a real broker: `tests/broker-discovery-probe.test.ts` — REGRESSION LOCK case is "stale-but-alive + outdated build → replace, never reuse", plus 6 sibling cases covering the other combinations (null ad, genuinely dead, matching build, mismatched protocol, and parity with the live-heartbeat path).

**Minors — all four addressed:**
1. `retryAmbiguousConnect` — the retry now backs off (200ms) before its one attempt, extracted as an injectable function (`connect`/`sleep` deps) so the backoff-before-retry ordering is unit-tested with fakes, not timers.
2. `cleanupOrphanedAdvertisementTempFiles` — sweeps `${path}.<dead-pid>.tmp` leftovers from a hard kill, called once at daemon startup before the first `writeAdvertisement`. 5 tests: dead pid swept, alive pid left alone, unrelated files untouched, missing directory is a no-op, mixed sweep leaves only the alive one.
3. `isAdvertisementLive`'s staleness slack now reads the same `FIGMA_AGENT_HEARTBEAT_MS`-derived cadence the refresh interval runs on (extracted a shared `envMs` helper into `protocol-helpers.ts`, used by both files) instead of a second hardcoded 30s value. Locked with a test that discriminates old vs. new behavior directly (fails under the old hardcoded slack, passes under the derived one).
4. Added the sentence to `broker-loopback-host.test.ts` naming `plugin/src/ui/ui-relay.ts`'s deliberate `figma-agent.localhost` connection (Figma Desktop's manifest rule, not Node's dual-stack resolution) as an intentional, unrelated exception — explicitly telling a future reader not to "fix" it into this test's guard.

Gates: typecheck / build / test all green (1055/1055, up from 1039), re-run 3x for flakiness on the timing-sensitive tests. Test-first re-verified for this round too: stashed the source changes, confirmed 14 of 15 new/changed assertions fail against pre-fix code (the 15th is a non-discriminating sanity companion, expected to hold both ways), restored the fix.

Holding for re-review.

jangtrinh (commented):
**Review verdict: APPROVE** — the regression is fixed, its lock provably discriminates, and all four minors are closed. (Posted as a review comment: GitHub refuses a formal approval state on a self-authored PR.)

Re-reviewed `4cd004c` on a fresh clone. All five items are closed, and I checked the regression lock the hard way rather than trusting its name.

## Verified

**The regression is fixed, and the lock genuinely discriminates.** `decideBrokerAction` (`cli/src/transport/broker-discovery.ts`) collapses both liveness signals into one and feeds them the identical gate:

```ts
const alive = deps.isAdvertisementLive(ad) || (await deps.probeStaleButAlive(ad));
if (!alive) return 'spawn';
const outdatedBuild = myMtime - ad.buildMtime > 1;
return ad.protocolV === PROTOCOL_VERSION && !outdatedBuild ? 'reuse' : 'replace';
```

There is no longer a path that returns an outdated advertisement. To confirm the lock pins that, I reintroduced the pre-fix bypass in my clone (stale-but-alive returning `'reuse'` directly): exactly two tests fail — the `REGRESSION LOCK: stale-but-alive + OUTDATED build → replace` case and its mismatched-protocol sibling — and nothing else. Restored, 24/24 green.

**The extraction is behaviour-preserving elsewhere**, which I checked case by case against the previous head: live + current → `reuse`; live + outdated/mismatched → `requestBrokerShutdown` then spawn; nothing advertised → spawn; not live and probe fails → spawn. Two things worth naming: the `||` short-circuit means the 1.5s handshake probe still only runs when the heartbeat already looks stale, so the happy path adds no latency; and stale-but-alive + outdated now takes the shutdown-then-respawn path rather than spawning on top of a live broker, which is the behaviour the previous round asked for.

**m1 — retry backoff.** `retryAmbiguousConnect` sleeps `AMBIGUOUS_CONNECT_RETRY_DELAY_MS` (200ms) before the second connect, with `connect`/`sleep`/`delayMs` injectable so the delay is asserted without waiting on a real timer, and the call site is the not-confirmed-dead branch only.

**m2 — orphan sweep is correctly narrow.** `cleanupOrphanedAdvertisementTempFiles` builds its matcher from the advertisement's own basename with the regex metacharacters escaped and both ends anchored (`^…\.(\d+)\.tmp$`), so it can only ever match this repo's own naming; it skips any entry whose owning pid is still alive (so a sibling broker mid-write, or a recycled pid, is left alone — the safe direction), tolerates an unreadable directory, and per-entry failures are swallowed. It runs once at startup before this process writes its own temp file, with a log line when it actually sweeps something, and the comment explains why it is not on the heartbeat tick.

**m3 — one source for the slack.** `envMs` now lives in `protocol-helpers.ts` and both the daemon and `broker-discovery.ts` read `FIGMA_AGENT_HEARTBEAT_MS` with the same `HEARTBEAT_INTERVAL_MS` fallback; `isAdvertisementLive` is `HEARTBEAT_STALE_MS + HEARTBEAT_MS`, so tolerance shrinks in lockstep with the cadence and the daemon's former local copy of `envMs` is gone. No second hardcode remains.

**m4 — the `.localhost` exception is documented where it will be read.** The added note in `tests/broker-loopback-host.test.ts` explains that the plugin's hostname is a Figma Desktop manifest constraint rather than Node's dual-stack resolution, and tells the next reader not to "fix" it into `LOOPBACK_HOST` or add `plugin/src` to the file list — which is exactly the failure mode a source-text lock invites.

Four discovery/daemon test files re-run twice in my clone: green both times, 24/24 in the probe suite.

## One nit, non-blocking

`ensureBroker` asserts the narrowing the decision function already knows: `if (action === 'reuse') return ad as BrokerAdvertisement;` and the same cast on the `replace` branch. The invariant ("any action other than `spawn` implies a non-null `ad`") is real but lives in another function, so the cast is doing the work the type could. Returning `{ action: 'reuse' | 'replace', ad } | { action: 'spawn' }` would let the compiler carry it and remove both casts. Purely cosmetic — the current code is correct.

*Informational:* the startup sweep lists `dirname(BROKER_FILE)` (i.e. `/tmp`), which on a long-lived machine can be a large directory. Startup-only is the right scoping and the comment says so; just worth remembering if broker start latency ever gets measured.

Nothing else outstanding from my side on this PR.

PR #13, opened 2026-07-31
feat: absorption phase-02 — Slots + annotations, editor-surface guard

### LINKED ISSUE #2: Absorption phase-02 — Slots + annotations (design editor)
Spec: plans/260731-1054-capability-absorption/phase-02. Slots via typeof createSlot host-check (maturity conflict GA-vs-beta noted in spec concerns); shared refusal builder for wrong-editorType. stdlib + recipes, no wire commands.

jangtrinh:
Closes #2.

## Summary
- `ui.slot.*` (create/list/append/reset/addProperty) — Figma Slots (GA June 2026), adapted from figma-console-mcp's slot handlers (MIT, see THIRD-PARTY.md). Every helper verifies its own write.
- `ui.annotate.*` (get/set/categories) — Figma Annotations, adapted from the fork's GET/SET_ANNOTATIONS handlers. `set` re-reads and throws E_EVAL on a count mismatch; `get` counts (never silently drops) children that throw on property access.
- `shared/editor-surface.ts` — pure `editorRefusal()` builder, the one place a wrong-editor-type refusal message is composed. No file in this phase needs to call it yet (Slots/annotations both work in Figma design mode, which is all this plugin currently runs in); it's built here, ahead of the phases that need it (03 FigJam, 04 Slides), per the plan's own rationale for landing it early.
- `STATUS.editorType` (additive) — reads `figma.editorType` directly, `null` (never the fork's silent `'figma'` default) when an older host reports nothing.
- Two genuine `@figma/plugin-typings` gaps found via `tsc --noEmit`, each worked around with a commented cast: `componentPropertyReferences.slotContentId` is untyped (the "Slots maturity conflict" this phase was scoped around), and `AnnotationCategory` exposes `.label`, not `.name`.
- Deliberate deviations from the fork, each commented at the site: ambiguous slot-name resolution throws instead of silently taking `[0]` (matches `byPath`'s precedent); `addProperty`'s "nested inside another slot" check is implemented for real (the fork's own tool description claims it, its code never does); `append`'s `content.nodeType` covers RECTANGLE/FRAME/TEXT only, not the fork's full 8-type set.
- Two new knowledge recipes (`slots.md`, `annotations.md`), every `[re-verify]` item worded as unverified-from-the-fork.
- `THIRD-PARTY.md` — two new attribution rows (slot files, annotate file).
- Small companion doc fix in the monorepo (`ease-design/opah`'s `knowledge/figma-agent-hand.md`): added the `editorType` status-shape note and pointers to all four `design-os-figma-plugin/knowledge/*.md` recipes — phase-01's own two pointers had never actually landed there, closing that gap too.

## Test plan
- [x] Test-first / diagnosed-first: a pre-existing `mock-figma.ts` gap surfaced by the new slot tests (`getNodeByIdAsync` couldn't resolve nodes created via plain `createFrame`/`createRectangle`/`createComponent`/`createSlot`) was root-caused via the actual failure messages before being fixed — every creation path now registers into `nodesById`, the same map `setMockComponents` already used.
- [x] 42 new tests (6 `editor-surface`, 25 `exec-stdlib-slot`, 11 `exec-stdlib-annotate`).
- [x] `npm run typecheck` — clean (`tsc -p cli && tsc -p plugin`, both `--noEmit`)
- [x] `npm run build` — clean (cli 231.3kb, plugin 158.5kb)
- [x] `npm test` — 85 files / 1097 tests green (was 82/1064 before the annotate suite landed within this branch; the 82/1064 baseline itself already included this branch's editor-surface + slot suites — no unexplained drift once the mock-registration fix was in)
- [x] No `npm run lint` script exists in this repo (same as phase-01) — skipped, consistent with the established gate.

## NOT done — flagging, not faking (consolidated with phase-01's still-open item)
Neither this phase's nor phase-01's (#10, merged) live-canvas checklist has been run yet — one connected session should run both, since #2 shares the same "no connected plugin, no foreground GUI" blocker phase-01 reported. Consolidated list for the owner to run once:

**From phase-01 (`ui.vars.*`, `ui.componentSet`):**
1. Build a variant matrix end-to-end via `ui.componentSet`, confirm `state=Default` case-sensitivity fix on a real set.
2. Exercise `ui.vars.*` (rename/remove/describe a variable, add/rename/remove a mode, setModeValue) on a real file.
3. Confirm `--undo-group` rollback on a script that throws partway through.

**From this phase (`ui.slot.*`, `ui.annotate.*`, editor guard):**
4. `ui.slot.create` → `list` → `append` (clone an instance in) → `list` again (child present) → `reset` → `list` (child gone), on a real COMPONENT.
5. `ui.slot.create` with `layoutMode:'GRID'` — expect refusal, verify no stray SLOT was created.
6. `ui.slot.append` with a raw COMPONENT as `sourceNodeId` — expect refusal, verify no orphan component was left.
7. `ui.slot.append` with a bad `sourceNodeId` **and** `clearExisting:true` on a populated slot — expect refusal with existing children still present.
8. `ui.annotate.set` with a markdown label + two pinned properties → `get` → confirm read-back, then `set(mode:'append')` on that node → confirm the pre-existing annotation survived.
9. `ui.annotate.set` with an invalid property type — expect refusal naming valid values.
10. `figma-agent status` → confirm `editorType: "figma"`.
11. Editor guard smoke — no FigJam file needed: covered by `tests/editor-surface.test.ts` (unit); a temporary guarded probe on canvas is optional per the plan's own fallback (§8 step 8) — skipped here, relying on the unit test.

Every `[re-verify]` tag in `knowledge/slots.md` and `knowledge/annotations.md` is worded "unverified, from the fork" rather than claimed as confirmed, per the plan's own instruction.

Holding for review.


jangtrinh (commented):
**Review verdict: REQUEST CHANGES** — two issues (a swallowed category lookup reported as "no categories", and a half-built slot property left behind when the helper's own verification fails); everything else verified sound. (Posted as a review comment: GitHub refuses a formal request-changes state on a self-authored PR.)

Reviewed `cbf6188` on a fresh clone. Suite reproduces exactly as claimed — 85 files / 1097 tests green once `kernel/design-os` is initialized (without it the panel gate fails with an explicit "run `git submodule update --init`" message, which is the right way for that dependency to announce itself).

The absorbed facts genuinely landed as guards rather than prose, the two typings gaps are handled honestly, and the unwired guard is the phase plan's own instruction rather than an omission — I checked that against the plan and the manifest, not just the PR description. Two issues need a change.

## Request changes

**1. A failed category lookup is reported as "this file has no categories".** `plugin/src/main/exec-stdlib-annotate.ts:55-60`

```ts
async function getCategories(): Promise<{ id: string; name: string }[]> {
  try { … } catch { return []; }
}
```

Every consumer treats `[]` as fact: `ui.annotate.categories()` returns `{ categories: [] }`, `get`/`set` report `availableCategories: []`, and `toOutput` then maps every `categoryId` to `categoryName: null` because the lookup map is empty. A transient failure, an older host, or a permissions problem is therefore indistinguishable from a file that genuinely has no annotation categories — and the caller has no signal to tell them apart.

This is the same shape this phase explicitly refuses one file over: `STATUS.editorType` deliberately emits `null` instead of the fork's `|| 'figma'` guess (`plugin/src/main/executor-ops.ts:93-96`, with a comment saying exactly why). The same standard should apply here. Suggest `availableCategories: null` (or a `categoriesUnavailable: true` flag) on failure, and not letting a categories failure quietly null out every `categoryName` — a name that could not be resolved *because the lookup failed* is a different fact from one whose id has no match.

**2. `ui.slot.addProperty` leaves a half-built property when its own verification fails.** `plugin/src/main/exec-stdlib-slot-property.ts:60-84`

`addComponentProperty` mints the SLOT property first; the frame link is written after; then two verification checks can throw. If either check fails — or the `componentPropertyReferences` merge throws — the component keeps a brand-new SLOT property with nothing linked to it, the call reports failure, and nothing tells the caller that the property survived. On a design system that is a real artifact someone has to find and delete by hand.

This repo already set the pattern for precisely this shape one PR ago: `ui.componentSet` reverses its *own* mutations when its *own* verification throws (`exec-stdlib-component-build.ts`'s `cleanup()`), on the argument that a self-inflicted failure should not leave debris. Same argument applies. Either delete the property (`deleteComponentProperty(propertyKey)`) before rethrowing, or name it in the error message so the state is visible rather than silent.

## Non-blocking

3. **The old-host guard gates `create` but not `addProperty`.** `exec-stdlib-slot.ts:36-39` refuses with "update Figma Desktop" when `typeof createSlot !== 'function'`, and the test proves it by deleting the method off the node. `addProperty` has no equivalent, so on an old Desktop `addComponentProperty(name, 'SLOT', …)` surfaces a raw platform error instead of the actionable refusal. `list`/`append`/`reset` degrade acceptably (no SLOT nodes can exist on such a host, so they fail with "slot not found" / an empty list), which leaves `addProperty` as the single gap.

4. **`set`'s verification only compares the count.** `exec-stdlib-annotate.ts:165-169` re-reads and compares `readBack.length` against what it wrote. The comment says "never trust the input we just built", but a dropped `properties` array or `categoryId` passes a count check. The returned payload *is* the read-back, so nothing is fabricated — this is a strictness gap rather than a lie. Comparing the fields actually written (label/markdown, property types, categoryId) would match the standard the rest of this stdlib holds.

5. **One mock creation path still doesn't register.** `tests/helpers/mock-figma.ts` — `mk()`, `createSlot()`, the instance graft, `setMockComponents` and `combineAsVariants` all populate `nodesById`, but `FakeNode.clone()` does not, so a clone appended into a slot via `content.clone: true` is not resolvable by `getNodeByIdAsync`. The PR's "every creation path now registers" is slightly wider than the code; either register there too or narrow the claim.

6. *Nit:* the 13-line Levenshtein in `exec-stdlib-annotate.ts:20-30` computes "nearest valid type" for an error message, where this codebase's established style is simply to name the candidates (`byPath`, and `resolveCollection`'s `listWithMore`). Thirty-three closed values fit in one line; the extra algorithm is upkeep for a message.

## Verified

- **The unwired guard is correct for this phase, not an oversight.** The plan states the guard lands "in the phase whose own capabilities do not need it", and `plugin/manifest.json` is still `"editorType": ["figma"]` — so there is no current FigJam or Slides path that could reach a slot/annotation helper, which is the question worth asking before accepting an unwired abstraction. `shared/editor-surface.ts` stays pure (no `figma` access) and is unit-tested across allowed / refused / unknown-`null` / multi-required.
- **`STATUS.editorType`** is additive at the end of the reply, changes no existing key or order, and reads `figma.editorType` directly with `?? null` rather than the fork's silent `'figma'` default.
- **The Slots facts survived as guards:** `createSlot()` takes no arguments (the mock mirrors that by not accepting one, and naming is rename-after-create); GRID is rejected before anything is created; the old-host refusal is real and actually exercised; ambiguous slot names throw with the direct-vs-nested distinction instead of silently taking `[0]`; `append` guards a raw `COMPONENT` *before* cloning and resolves content before honouring `clearExisting`, so a bad `sourceNodeId` cannot empty a populated slot; `addProperty` merges `componentPropertyReferences` instead of assigning it, and implements the nested-inside-another-slot check the fork's own description claims but its code never performs.
- **Both typings gaps are handled honestly.** `slotContentId` goes through one commented, localized cast in a single named reader (`exec-stdlib-slot-resolve.ts:9-14`) rather than an ambient widening or a bare `any`; and `AnnotationCategory.label` needed no cast at all — the code reads the real property and maps it to `name`, which is the better outcome than working around it.
- **`get` is honest about what it could not read:** children whose property access throws increment `skippedChildren` instead of vanishing, and an unmatched `categoryId` reports `categoryName: null` rather than echoing the id.
- **Mock fidelity improved in the right direction:** `annotations` exists only on node types that really have it (so `'annotations' in node` is a genuine capability test), `findAllWithCriteria` walks deep so the ambiguity rule has real structure to disagree with, and `addComponentProperty` mints a namespaced key and merges into existing definitions.
- Knowledge recipes carry `[re-verify]` wording as unverified-from-the-fork, THIRD-PARTY.md gains the two attribution rows, and the consolidated live-canvas checklist is flagged rather than quietly dropped.

jangtrinh:
Fix round pushed (ec15783), addressing the stage-4 review:

- **BLOCKER 1** (category-lookup swallow): `getCategories()` no longer catches a thrown `getAnnotationCategoriesAsync()` into `[]`. A real failure now surfaces as `E_EVAL` through `get`/`set`/`categories` alike — an honest empty list and a swallowed throw no longer look the same.
- **BLOCKER 2** (orphaned property on verify failure): `addSlotProperty` now deletes the property it just minted if its own post-mint verify fails — safe because the mint happened in this same synchronous call (single-threaded main thread; nothing else could reference it yet), scoped so it never touches a pre-existing property. If the rollback deletion itself fails, the property is left in place and named in the thrown error alongside the original failure.
- **m1**: `addProperty` gets the same old-host `typeof` guard as `create`.
- **m2**: `set`'s read-back verify now compares the annotations actually written, not just the count.
- **m3**: `FakeNode.clone()` registers into `nodesById` (closes the mock's "every creation path resolves by id" claim for real).
- **m4**: dropped the bespoke Levenshtein ranking for invalid annotation property types — lists all 32 valid values instead, matching this codebase's existing `q()`/`byPath` error-message convention.

Test-first: all 6 new/changed tests confirmed to fail against the pre-fix code before landing (stashed the fix, ran red, restored, ran green). Gate green: typecheck clean, compile clean, 85 files / 1103 tests (was 1097 — +6 explained above).

Holding for re-review.

jangtrinh (commented):
**Review verdict: APPROVE** — both blockers closed, each proven to discriminate by reverting the fix in isolation; three non-blocking notes below. (Posted as a review comment: GitHub refuses a formal approval state on a self-authored PR.)

Re-reviewed `ec15783` on a fresh clone: 85 files / 1103 tests green, matching the reported gates exactly. Both blockers are closed and I confirmed their tests discriminate by reverting each fix in isolation rather than trusting the stash-verification note.

## Verified

**BLOCKER 1 — a thrown category lookup now surfaces.** `plugin/src/main/exec-stdlib-annotate.ts:46-62` separates the two facts it previously conflated: an empty array from a *successful* call stays an honest answer, while a throw becomes `E_EVAL` with the underlying message. Reverting to `catch { return []; }` fails exactly the two new tests — the `categories()` case and the `get()` propagation case — and nothing else. The distinction that mattered is preserved on the read side too: a `categoryId` present with `categoryName: null` is still visible as an unresolved id rather than being flattened into "no category".

**BLOCKER 2 — the mint is now reversed on a failed post-mint verify.** `plugin/src/main/exec-stdlib-slot-property.ts:74-115` wraps everything after `addComponentProperty` in a `try`, deletes the property on failure, and rethrows `original` — the same object, so code and stack survive the ordinary rollback path. Removing the deletion fails both new tests (the rollback case and the rollback-failure case) and nothing else.

**The cleanup cannot mask the original error** — the concern I raised on the broker cleanup in #12. Here the fallback composes rather than replaces: the thrown message carries the original failure *and* names the leftover key, and the code is taken from the original (`original.code ?? 'E_EVAL'`). Both paths are tested, which is the part that makes the guarantee real.

**The boundary comment is accurate.** The rollback deletes only `propertyKey`, minted in the same synchronous call, and the comment says plainly that it must never be generalised to a pre-existing property another variant might depend on. That is the correct scope — deleting a property that other variants reference would be a materially different and more dangerous action, and nothing in this code path can reach one.

**Minors all landed.** `addProperty` gained the old-host `typeof` refusal, matching `create`'s (`exec-stdlib-slot-property.ts:64-70`); `set`'s verification now deep-compares the projected annotations rather than counting them (`exec-stdlib-annotate.ts:166-181`), with a test where the count matches but the content does not; `FakeNode.clone()` registers into `nodesById`, so "every creation path resolves by id" is now true as stated; and the Levenshtein ranking is gone in favour of naming all 32 valid types, which is this codebase's existing error style.

## Non-blocking

1. **The deletion path rests on two platform assumptions the mock cannot test.** `deleteComponentProperty` is called with the *key* returned by `addComponentProperty` (`name#id`), and it is declared inside the same hand-written cast as `addComponentProperty` (`exec-stdlib-slot-property.ts:59-62`). The mock deletes by that same key, so it necessarily agrees with the code rather than checking it — if the real API expects the bare property name, or exposes a different signature, the rollback throws and the fallback path fires (property survives, error names it — a safe degradation, but not the intended one). Two small things would close it: narrow the cast to the member that actually needs asserting (the `'SLOT'`-typed `addComponentProperty`) so the compiler keeps checking the rest, and add "confirm `addProperty` rollback actually removes the property" to the live-canvas checklist, next to the existing slot items.

2. *Informational:* on the rollback-failure path the thrown value is a new `Error`, so the original stack is replaced (message and code are preserved). Only reachable when two failures coincide; worth knowing when reading such a report.

3. *Informational:* `get` is a read path that now fails hard when the category API throws — which is the ruling, and the message is actionable. Worth being explicit that this also means a host *missing* `getAnnotationCategoriesAsync` entirely could not read annotations at all. If that is a real user state, the consistent treatment is the one m1 just added for Slots: a `typeof … !== 'function'` refusal naming the fix, rather than the generic unavailable-message.

Nothing else outstanding from my side. The remaining open item on this PR is unchanged: the consolidated live-canvas checklist (phase-01's items plus this phase's), still flagged rather than claimed.

PR #14, opened 2026-07-31
Audit fix: reply sender-verification in routeFromPlugin + op-scaled BATCH timeout

### LINKED ISSUE #6: Audit — reply sender-verification in routeFromPlugin + op-scaled timeouts
Backlog 2.10: verify a reply from a DIFFERENT plugin instance than dispatched cannot be accepted (esp. around reconnect with same instanceId, socket changed). Then per-op timeout scaling (component-set by variant count) instead of one fixed PLUGIN_WAIT.

jangtrinh:
Closes #6.

## Audit finding (phase 1, reported before any fix)

**The hole exists.** `routeFromPlugin` (`cli/src/transport/broker-daemon.ts`) routed a
reply frame to the waiting CLI keyed on the wire `id` alone. It never checked which
socket actually sent the frame against `job.targetInstanceId` — the plugin instance a
job was pinned to at admission (`JobRecord.targetInstanceId`, already present in this
codebase's job table). Any OTHER currently-connected plugin instance sending a reply
carrying the same `id` was accepted and forwarded to the CLI as though it were the real
dispatched plugin's answer, with zero signal that it wasn't.

**How it's reachable.** Request ids are minted CLI-process-side as `c_<counter>_<ts>`
(`shared/protocol.ts`'s `makeRequestId`), and `requestCounter` resets to 0 on every fresh
`figma-agent` invocation. Two concurrent CLI invocations (two terminals, a script
launching parallel commands) issuing their first command in the same millisecond produce
identical ids. If those two commands dispatch to two different plugin instances (e.g. two
open Figma files), whichever admission call landed second silently overwrote `pending`/
`dispatchedTo` for that id — and the FIRST plugin's real reply then got routed to the
SECOND CLI caller as if it were the answer that CLI was waiting for.

**Reconnect nuance (the issue's specific callout).** A naive fix comparing raw
`WebSocket` object references would have broken a legitimate case: when a plugin
reconnects (Figma iframe reload), `PluginRegistry.register()` replaces the socket for the
SAME `instanceId` — any job dispatched before the reconnect still has the OLD ws in
`dispatchedTo`. A ws-reference check would misclassify the reconnected instance's own
reply as "a different instance". The fix verifies by **instance identity**
(`instanceId`), never socket reference, so this stays correct.

**Proof.** `tests/broker-daemon-harness.test.ts` — two genuinely distinct, simultaneously
connected plugin instances (X, Y); dispatch a job to X; Y sends a spoofed reply carrying
X's dispatched id; assert the CLI never receives Y's payload, only X's real one. This
test fails against pre-fix code (Y's spoofed `{spoofed: true, from: 'Y'}` was what the CLI
actually received).

## Fix (phase 1)

`isReplyFromDispatchedInstance` (`cli/src/transport/job-table.ts`) — a pure predicate
comparing the sender's registry-resolved `instanceId` against `job.targetInstanceId`.
Wired into `routeFromPlugin`, which now takes the sender's own `ws` at both of its call
sites (chunk and reply). A mismatch is counted + logged and discarded outright — never
touches `pending`/job state, so the real dispatched plugin's reply (or the existing
watchdog timeout) is still what resolves the job. Unit-tested directly in
`tests/job-table-sender-verification.test.ts` (dispatched-instance match, cross-instance
mismatch, unrecognised/superseded sender, reconnect-identity resilience, unknown-job
no-op) — deterministic, since racing a real WebSocket reconnect against the unrelated
pre-existing `handleClose` disconnect-triggered job failure is not a reliable integration
test (documented in the harness test file).

## Phase 2 — per-op timeout scaling

**Deviation flagged:** the issue's phrasing ("component-set-class operations scale by
variant count") assumes a `CREATE_COMPONENT_SET`-equivalent command. This codebase has no
such command — the structurally closest analog to "many variants built in one
uncancellable pass" is `BATCH` (N arbitrary ops executed sequentially in one round trip).
`batchTimeoutMs` (`cli/src/commands/batch.ts`) scales `BATCH`'s per-attempt timeout by op
count, adapting `componentSetTimeoutMs`'s formula (per-unit budget, floor, cap, hop
buffer) from southleft/figma-console-mcp — same 1.2s-per-unit budget and 120s cap, but
the floor is pinned to today's existing `COMMAND_TIMEOUTS.BATCH` (60s) so an ordinary
batch keeps its exact current timeout; only a batch large enough that `1.2s × opCount +
5s` would exceed 60s gets more room.

`batch.ts` gained an `execute(filePath, stopOnError, runner)` + injectable `Runner` type,
matching `scan-design-system.ts`'s existing DI pattern for testability without a live
broker.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run build` (no drift in committed `plugin/code.js`/`cli/dist` — this PR never
      touches `plugin/src`)
- [x] `npm test` — 1068/1068 (up from 1055), re-run 3x for stability
- [x] Test-first verified honestly: stashed all source changes, confirmed all 13
      new/changed test cases fail against pre-fix code (the daemon-harness spoof test
      literally receives Y's spoofed payload instead of X's real one), restored the fix

## Not verified

- No live Figma plugin round-trip (no Figma Desktop instance in this environment).
- The full real-world reconnect race (superseded socket closing, `handleClose` racing a
  legitimately reconnected instance's reply) is not exercised end-to-end — see the note
  in `tests/broker-daemon-harness.test.ts` for why, and `tests/job-table-sender-verification.test.ts`
  for the deterministic identity-level proof instead.
- Did not add a `BATCH` size ceiling/warning (the fork's write-tools.ts advises splitting
  above ~40 variants) — out of scope for this backlog item as phrased; flagging in case a
  follow-up wants it.

### DIFF
--- a/cli/src/commands/batch.ts
+++ b/cli/src/commands/batch.ts
@@ -2,7 +2,7 @@
 // BATCH request (single round-trip; plugin executes sequentially).
 import { readFileSync } from 'node:fs';
 import { resolve } from 'node:path';
-import { COMMANDS, type CommandName } from '../../../shared/protocol.ts';
+import { COMMAND_TIMEOUTS, COMMANDS, DEFAULT_TIMEOUT_MS, type CommandName } from '../../../shared/protocol.ts';
 import type { CommandArgs } from '../figma-agent.ts';
 import { CliError } from '../transport/protocol-helpers.ts';
 import { runCommand } from '../transport/broker-client.ts';
@@ -31,9 +31,51 @@ function loadOps(filePath: string): BatchOp[] {
   });
 }
 
+/**
+ * Audit backlog 2.10, phase 2 — BATCH executes every op sequentially in ONE
+ * uncancellable pass (the plugin sandbox cannot be interrupted mid-sequence), so a
+ * fixed timeout that fires early doesn't stop the work — it just makes the CLI's
+ * report contradict the file state (ops keep landing after the reported "failure")
+ * and invites a duplicate retry that double-applies. Mirrors
+ * southleft/figma-console-mcp's `componentSetTimeoutMs` (per-unit budget, floor,
+ * cap, hop buffer over the same base formula) — adapted per-OP rather than
+ * per-variant: this codebase has no CREATE_COMPONENT_SET-equivalent single
+ * command, so BATCH (N ops, one round trip, one uncancellable pass) is the
+ * structurally closest analog available to scale.
+ *
+ * Never scales BELOW `COMMAND_TIMEOUTS.BATCH` (today's fixed default): an ordinary
+ * batch keeps today's exact timeout unchanged. Only once `opCount * PER_OP_MS +
+ * HOP_BUFFER_MS` would exceed that default does the wait actually grow, capped at
+ * `CAP_MS` — mirroring the fork's own cap.
+ */
+export function batchTimeoutMs(opCount: number): number {
+  const PER_OP_MS = 1_200; // mirrors componentSetTimeoutMs's per-unit budget
+  const HOP_BUFFER_MS = 5_000; // mirrors the fork's buffer over its own hop
+  const CAP_MS = 120_000; // same cap the fork uses
+  const base = COMMAND_TIMEOUTS.BATCH ?? DEFAULT_TIMEOUT_MS;
+  const scaled = opCount * PER_OP_MS + HOP_BUFFER_MS;
+  return Math.min(CAP_MS, Math.max(base, scaled));
+}
+
+/** A command runner (the BATCH transport call), injectable for tests. */
+export type Runner = (cmd: string, params: unknown, opts?: { timeoutMs?: number }) => Promise<unknown>;
+
+/**
+ * Decoupled from `CommandArgs` + the real transport so it is unit-testable with a
+ * stub runner and a fixture file, matching `scan-design-system.ts`'s `execute`
+ * pattern.
+ */
+export async function execute(
+  filePath: string,
+  stopOnError: boolean,
+  runner: Runner = runCommand,
+): Promise<unknown> {
+  const ops = loadOps(resolve(filePath));
+  return runner('BATCH', { ops, stopOnError }, { timeoutMs: batchTimeoutMs(ops.length) });
+}
+
 export async function run(args: CommandArgs): Promise<unknown> {
   const fileArg = args.positionals[0];
   if (!fileArg) throw new CliError('E_INVALID_ARGS', 'usage: figma-agent batch <file.json> [--stop-on-error]');
-  const ops = loadOps(resolve(fileArg));
-  return runCommand('BATCH', { ops, stopOnError: args.bool('stop-on-error') });
+  return execute(fileArg, args.bool('stop-on-error'));
 }
--- a/cli/src/transport/broker-daemon.ts
+++ b/cli/src/transport/broker-daemon.ts
@@ -35,7 +35,7 @@ import {
   resolveProjectDir, writeBindCache, writeBindMarker, type Binding,
 } from './project-bind.ts';
 import {
-  JobTable, isFinishedState, toJobInfo, type JobRecord,
+  JobTable, isFinishedState, isReplyFromDispatchedInstance, toJobInfo, type JobRecord,
 } from './job-table.ts';
 import {
   completeSkippingStale, emptyQueue, enqueue as enqueueJob, queuePosition, remove as removeFromQueue,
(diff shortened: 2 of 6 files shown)


jangtrinh (commented):
**Review verdict: APPROVE** — the vulnerability and its fix are both reproduced end to end (disabling the new check makes the waiting CLI accept the spoofed reply); three non-blocking items below. (Posted as a review comment: GitHub refuses a formal approval state on a self-authored PR.)

Reviewed `8c99eac` on a fresh clone; 20/20 green across the three affected suites, and I verified the headline test discriminates rather than trusting the description.

The audit finding is real and the fix addresses it at the right layer. Approving with three non-blocking items.

## Verified

**The vulnerability and the fix are both proven.** Disabling the new check in `routeFromPlugin` (`cli/src/transport/broker-daemon.ts`) makes the harness test fail with exactly the pre-fix symptom — the waiting CLI receives `{ spoofed: true, from: 'Y' }` instead of X's real reply. Restored, green. That is the whole vulnerability reproduced end to end, not a predicate asserted in isolation.

**Identity-based, not socket-based — and that distinction matters here.** `isReplyFromDispatchedInstance` (`cli/src/transport/job-table.ts:31-61`) compares the registry-resolved `instanceId` against `job.targetInstanceId`, so a plugin that reconnects (new socket, same instance) still completes its own job, while a genuinely different instance is rejected. A `WebSocket`-equality check would have broken the honest reconnect, which is the trap the issue called out.

**It composes correctly with the terminal-job guard from the concurrency work.** The sender check runs *first*, so a wrong-instance frame arriving against an already-terminal job increments the sender-mismatch counter rather than `lateReplyCount` — which is the right attribution: a foreign frame is not "the dispatched target answering late", and folding it into that counter would corrupt the one signal that means exactly that. Equally important, the CLI forward (`client.send(rawText)`) now sits *after* verification; pre-fix the spoofed frame reached the caller before any job logic ran at all.

**The pre-registration window cannot cause a legitimate discard.** Both call sites are gated behind `isPlugin = st.registry.touch(ws)`, and `touch` returns false for any socket the registry does not hold (`cli/src/transport/plugin-registry.ts:85-90`). A socket that has not yet sent `PLUGIN_HELLO` therefore never reaches `routeFromPlugin` at all — it takes the CLI branch, exactly as before this PR. No behaviour change there.

**Un-dispatched replies are unchanged.** `!job → true`, then the existing `pending` lookup and forward. Nothing about the no-job path moved.

**The BATCH deviation reasoning holds.** There is no `CREATE_COMPONENT_SET`-equivalent command in this codebase, and `BATCH` genuinely is the "N units executed in one uncancellable pass" analog, which is the property that makes a fixed timeout harmful. Pinning the floor to today's `COMMAND_TIMEOUTS.BATCH` means an ordinary batch (up to ~45 ops) keeps its exact current timeout, so the change is inert where it should be. Worth noting that the concurrency wave's `shouldWarmRetry` already refuses to retry a `jobId`-carrying `E_TIMEOUT`, so a scaled-but-still-exceeded timeout cannot turn into a double-applied batch.

**The reconnect-race scoping is acceptable**, and for a reason worth writing down: a superseded pre-reconnect socket fails `touch`, so its frame is dropped one level above this check, and the job then resolves honestly through `handleClose`'s `E_NO_PLUGIN` rather than by accepting a frame the registry no longer recognises. That is a better end state than the pre-fix accidental completion, which is what makes the predicate-level coverage sufficient here.

## Non-blocking

1. **The mismatch counter never reaches an operator.** `senderMismatchCount` (`broker-daemon.ts:381-386`) is logged and nothing more. This repo has an established pattern for exactly this class of counter — `resultDropped` on a job poll, `orphan_shards` and `skip_history_truncated` in the reconcile envelope, `overflowed`/`edgeEvictedUnresolved` from `sync-corrections` — all machine-readable, not log-only. A discarded cross-instance reply is arguably the most security-relevant of the set: surfacing it additively in `BROKER_HELLO`/`status` (default `0`) would let an agent notice it without tailing `/tmp/figma-agent-broker.log`.

2. **The 120s cap and the missing size ceiling interact.** `batchTimeoutMs` is `min(120_000, max(60_000, 1_200·n + 5_000))`, so from roughly 96 ops upward the computed budget is clamped below what the formula says the pass needs — which reintroduces precisely the failure this phase exists to remove, just at a higher op count. The flagged "no BATCH size ceiling" gap therefore isn't independent of the cap; the two decisions only make sense together. Cheapest honest option: refuse a batch whose scaled budget would exceed the cap, naming the op count and suggesting a split (the pass is uncancellable, so an early timeout is worse than a refusal). Raising the cap for `BATCH` specifically is the alternative. Either way this is worth a board issue rather than a comment.

3. **The predicate's documented `null`-sender case is unreachable from the real call sites.** The doc describes `senderInstanceId == null` as covering "a stale reference to an already-superseded pre-reconnect socket", but such a socket fails `touch` and never gets here, so that branch is defence in depth only and its counter will never increment for that reason. Worth wording it as defensive, otherwise the next reader goes looking for a counter that cannot move.

*Informational:* `!job → true` means verification is skipped whenever the job record is absent. That is narrow today — live jobs are never evicted (the finished-cap only touches terminal records), and a restarted broker has neither job nor `pending`, so nothing is forwarded — but it is the one door left, and it would widen if the job table ever evicted live records.

PR #17, opened 2026-07-31
fix(broker): surface senderMismatchCount in the reply/status envelope

### LINKED ISSUE #15: Surface senderMismatchCount in the reply envelope (not log-only)
From PR #14 review (non-blocking). A discarded cross-instance reply increments senderMismatchCount but only logs it. This repo surfaces exactly this counter class machine-readably (resultDropped, orphan_shards, skip_history_truncated) and a discarded cross-instance reply is the most security-relevant of the set — surface it in the reply/status envelope so an operator/agent can see a spoof/misroute happened.

jangtrinh:
## Summary
- PR #14's sender-verification fix discards a reply from a plugin instance that doesn't match the job's dispatched target, and counts it in `senderMismatchCount` -- but that counter was log-only, invisible to any caller.
- This repo's own convention (`resultDropped`, `lateReplyCount`) already surfaces "a thing got discarded" as a machine-readable counter in envelopes. A discarded cross-instance reply (a spoof/misroute) is the most security-relevant member of that class, so it gets the same treatment.
- `senderMismatchCount` is daemon-scoped (not per-job), so it surfaces in `BROKER_HELLO` (`buildBrokerHelloData`/`BrokerMeta`) -- the same status payload `figma-agent status` already reads -- and is threaded through into `status.ts`'s own `broker` block.
- Follows the `resultDropped`/`lateReplyCount` convention exactly: the key is present only once a mismatch has actually happened, so the payload stays byte-identical to before this field existed when the count is zero.

Closes #15.

## Test plan
- [x] `tests/broker-status.test.ts` -- unit coverage on `buildBrokerHelloData`: zero mismatches omits the key (byte-identical), a nonzero count surfaces verbatim.
- [x] `tests/broker-daemon-harness.test.ts` -- full daemon integration test: two real plugin instances, one spoofs a dispatched job's reply id, a brand-new connection's `BROKER_HELLO` greeting reports `senderMismatchCount === 1`.
- [x] Both new tests verified to fail against pre-fix code (test-first).
- [x] `npm run typecheck && npm run build && npm test` all green (87 files, 1119 tests), including the panel gate against the pinned kernel submodule. Re-ran the full suite 3x to confirm no flakiness.

### DIFF
--- a/cli/src/commands/status.ts
+++ b/cli/src/commands/status.ts
@@ -32,6 +32,10 @@ export async function run(args: CommandArgs): Promise<unknown> {
     pid: ad.pid,
     uptimeMs: (hello.uptimeMs as number | undefined) ?? null,
     protocolVersion: (hello.protocolV as number | undefined) ?? PROTOCOL_VERSION,
+    // Sender-verification counter (backlog 2.10 / issue #15) — mirrors BROKER_HELLO's
+    // own byte-identical-when-zero contract: present only once a cross-instance reply
+    // has actually been discarded, so the common (zero) case stays unchanged here too.
+    ...(typeof hello.senderMismatchCount === 'number' && { senderMismatchCount: hello.senderMismatchCount }),
   };
   const all = Array.isArray(hello.plugins) ? (hello.plugins as PluginStatusEntryWithJob[]) : [];
 
--- a/cli/src/transport/broker-daemon.ts
+++ b/cli/src/transport/broker-daemon.ts
@@ -1268,7 +1268,10 @@ export async function runBrokerDaemon(options?: BrokerDaemonOptions): Promise<vo
     type: 'BROKER_HELLO',
     data: buildBrokerHelloData(
       st.registry,
-      { port, pid: process.pid, protocolV: PROTOCOL_VERSION, buildMtime: selfBuildMtime(), uptimeMs: Date.now() - startedAt },
+      {
+        port, pid: process.pid, protocolV: PROTOCOL_VERSION, buildMtime: selfBuildMtime(),
+        uptimeMs: Date.now() - startedAt, senderMismatchCount,
+      },
       currentFilter(),
       Date.now,
       jobStatusFor,
--- a/cli/src/transport/broker-status.ts
+++ b/cli/src/transport/broker-status.ts
@@ -25,6 +25,11 @@ export interface BrokerMeta {
   protocolV: number;
   buildMtime: number;
   uptimeMs: number;
+  /** Sender-verification counter (backlog 2.10 / issue #15) — how many reply frames
+   *  `routeFromPlugin` has discarded this daemon run because the sending socket's
+   *  plugin instance did not match the job's `targetInstanceId` (a spoofed/misrouted
+   *  cross-instance reply). Daemon-scoped, not per-job. */
+  senderMismatchCount: number;
 }
 
 /**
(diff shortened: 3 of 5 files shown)


jangtrinh (commented):
**Review verdict: APPROVE** — byte-identical-when-zero confirmed, the surfaced counter is the live one (no drift), the end-to-end test discriminates, and the harness helper is test-only: production CLIs cannot miss the first BROKER_HELLO. (Posted as a review comment: GitHub refuses a formal approval state on a self-authored PR.)

Reviewed `c9a7926` on a fresh clone; 21/21 across the two affected suites, and both new tests discriminate (removing the surface line in `broker-status.ts` fails the unit nonzero-surface case and the end-to-end harness case, nothing else). This is a clean, correctly-scoped change. Approving.

## Verified

**Byte-identical-when-zero is real.** `buildBrokerHelloData` gates the field behind `meta.senderMismatchCount > 0` (`cli/src/transport/broker-status.ts:73`), exactly the `resultDropped`/`orphan_shards` pattern, so a broker that has never discarded a spoofed reply emits the same payload as before the field existed. The unit test asserts the zero case omits the key entirely, not that it is `0`.

**The surfaced counter is the live one, not a copy.** `senderMismatchCount` is a single closure `let` in `runBrokerDaemon` that `routeFromPlugin` increments; `brokerHello()` is rebuilt on every call and reads that same variable through the `meta` object it constructs at call time (`broker-daemon.ts:1267-1276`). There is no second field or snapshot that could drift — the greeting reflects the count as of the moment it is sent.

**The integration test is genuinely end to end.** It stands up two real instances, dispatches a job to X, has Y send a reply carrying X's dispatched id, drains X's real reply, then opens a *fresh* connection and reads `senderMismatchCount: 1` off that new BROKER_HELLO greeting — the same frame `figma-agent status` consumes via `fetchBrokerHello`. That exercises the increment, the daemon-scoped visibility (any connection, not just the original CLI), and the envelope surfacing in one path.

**`status.ts` does not fabricate a zero.** The passthrough is `...(typeof hello.senderMismatchCount === 'number' && { senderMismatchCount: hello.senderMismatchCount })` (`cli/src/commands/status.ts:35`), so an absent field stays absent in the CLI output rather than becoming `0` — the absent-vs-zero distinction is preserved end to end, matching the envelope's own contract.

## Production-ordering question (the deviation) — answered: TEST-ONLY, not masking a real bug

`connectAndAwaitBrokerHello` is confined to `tests/broker-daemon-harness.test.ts`; no production file was touched for it, and the daemon's synchronous `ws.send(brokerHello())` in the connection handler is unchanged.

More importantly, the synchronous-on-connect greeting is **correct for real CLIs**, and here is the evidence rather than an assumption: the only production consumer of that greeting, `fetchBrokerHello` (`cli/src/transport/broker-client.ts`), attaches its `message` listener **synchronously in the Promise executor, at socket construction — before `open`**. So the listener provably exists before any I/O can occur, and a frame the broker pushes the instant the connection completes cannot be dropped. The bug the helper fixes was purely in the *test's* older `connectSocket` → `await 'open'` → *then attach* sequence, which opens a real window where a frame sent between `open` and the listener attach is lost. The new helper simply adopts the pattern production already uses.

Put plainly: real CLIs cannot miss the first BROKER_HELLO, because the code path that reads it never waits for `open` to attach its listener. The deviation papers over a test-harness sequencing artifact, not a production first-frame-drop. (I checked `fetchBrokerHello` specifically because it is the sole reader of the greeting; the command path `connectWs`/`exchange` does not depend on catching it.)

## Non-blocking

1. `BrokerMeta.senderMismatchCount` is a **required** field now, not optional. Correct for this repo — there is exactly one production caller and TypeScript would flag any other — but it is a slightly stronger contract than the optional-arg style the job-status fields used (`jobStatusFor?`). No change needed; noting it because the PR description framed this as "following the optional-arg pattern", and the field itself is required (its *surfacing* is what is conditional).

2. *Informational:* the counter is daemon-scoped and resets on restart, which the doc comment states. That is the right scope for this signal, but it means an operator reading a nonzero value cannot tell whether the discards were one burst or spread over a long-lived broker — fine for "did this ever happen", worth a timestamp only if it ever needs triage depth.

Nothing else outstanding.

Issue #7, opened 2026-07-31
Feed routing hygiene — cwd-independent staging root (5.6) + error-log binding (5.9)


jangtrinh:
Backlog 5.6+5.9: unbound staging roots are broker-cwd-relative (orphan on cwd change; tracking comments at both staging helpers); errorLogPath resolved once at startup with no binding awareness. One small wave: route BOTH through a cwd-independent root + binding identity, mirror DOC_CHANGE/EDIT_FEED pattern.

jangtrinh:
PR: #19 (holding for review). Blast radius + 3 forks were sent to team-lead directly before implementation (rulings: migrate old staging not orphan; full parity for errors including staging+migrate; fix the FileContext.fileKey comment; fold in the appendErrorFrame rotation gap too) — implemented per those rulings.

PR #18, opened 2026-07-31
fix(cli): refuse BATCH admission above the 120s scaled-timeout ceiling

### LINKED ISSUE #16: BATCH op-count hard refusal above the 120s scaled-timeout ceiling
From PR #14 review + team-lead ruling. The per-op timeout scaling (PR #14) clamps at 120s, so from ~96 ops up the budget is below what the formula says an uncancellable BATCH pass needs — reintroducing the timeout failure at higher n. RULING: HARD REFUSAL, not a raised cap — refuse a BATCH whose scaled budget would exceed the 120s ceiling BEFORE dispatch, with an actionable message ('N ops exceeds the single-pass budget; split into batches of ≤M'). An uncancellable pass must never be dispatched knowing it will time out. Cheap, loud, no double-apply risk. Test the boundary (M ops accepted, M+1 refused).

jangtrinh:
## Summary

Closes #16. PR #14's op-scaled BATCH timeout clamps at 120s (`CAP_MS`), so from
~96 ops up the uncapped formula (`opCount * PER_OP_MS + HOP_BUFFER_MS`) already
exceeds the cap: `batchTimeoutMs` silently returns 120s regardless, which is now
*less* than what the formula itself says the pass needs. Dispatching anyway
sends an uncancellable, sequential plugin-side pass that keeps running past its
own reported timeout — the CLI reports failure while ops keep landing, inviting
a double-apply retry.

Ruling from the issue: **hard refusal before dispatch, not a raised cap.**

- `maxBatchOps()` — the largest op count whose uncapped budget still fits inside
  `CAP_MS` (95: `95*1200+5000=119000 <= 120000`, `96*1200+5000=120200 > 120000`).
- `assertBatchAdmissible(opCount)` — throws `E_INVALID_ARGS` naming the op count,
  the 120s ceiling, and the max ops that fit, e.g.:
  `96 ops exceeds the single-pass budget of 120s; split into batches of ≤95`
- Wired into `execute()` before the runner is ever invoked — a batch over the
  boundary never reaches the broker; no job is created for a pass already known
  to time out mid-flight.
- `PER_OP_MS`/`HOP_BUFFER_MS`/`CAP_MS` hoisted to module scope so `batchTimeoutMs`
  and the new admission check share one formula instead of two constants that
  could drift apart.

## Test plan

- [x] New assertions in `tests/batch-op-ceiling.test.ts` confirmed to **fail
      against pre-fix `batch.ts`** (`assertBatchAdmissible`/`maxBatchOps` did not
      exist; `execute()` dispatched any op count unconditionally) — proven via
      `git stash` before implementing, re-run, `git stash pop`.
- [x] Boundary test: M (95) ops accepted and dispatched; M+1 (96) ops refused,
      runner never invoked.
- [x] `npm run build` — clean
- [x] `npm test` — 88 test files, 1127 tests passed (includes the panel gate /
      submodule)

### DIFF
--- a/cli/src/commands/batch.ts
+++ b/cli/src/commands/batch.ts
@@ -31,6 +31,15 @@ function loadOps(filePath: string): BatchOp[] {
   });
 }
 
+// Mirrors southleft/figma-console-mcp's componentSetTimeoutMs's per-unit budget,
+// hop buffer, and cap — shared between `batchTimeoutMs` (scales the wait) and
+// `maxBatchOps`/`assertBatchAdmissible` (issue #16: refuses what the SAME formula
+// says the cap can no longer cover). Hoisted to module scope so both readings of
+// the formula can never drift apart.
+const PER_OP_MS = 1_200; // mirrors componentSetTimeoutMs's per-unit budget
+const HOP_BUFFER_MS = 5_000; // mirrors the fork's buffer over its own hop
+const CAP_MS = 120_000; // same cap the fork uses
+
 /**
  * Audit backlog 2.10, phase 2 — BATCH executes every op sequentially in ONE
  * uncancellable pass (the plugin sandbox cannot be interrupted mid-sequence), so a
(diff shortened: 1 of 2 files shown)


jangtrinh (commented):
**Review verdict: APPROVE** — the ceiling is computed from the shared PER_OP_MS/HOP_BUFFER/CAP constants (desync trap closed, proven), the refusal fires before any dispatch (runner never invoked), and empty/1-op batches are unaffected. (Posted as a review comment: GitHub refuses a formal approval state on a self-authored PR.)

Reviewed `4c745ee` on a fresh clone; 15/15 across the two batch suites. This is the hard refusal I asked for as a follow-up on #14, implemented the way that avoids the trap it was guarding against. Approving.

## Verified

**The ceiling is COMPUTED from the same constants the timeout uses, not re-hardcoded.** `maxBatchOps()` returns `Math.floor((CAP_MS - HOP_BUFFER_MS) / PER_OP_MS)` (`cli/src/commands/batch.ts`), and those three constants are hoisted to module scope and shared with `batchTimeoutMs`. I confirmed the desync trap is actually closed: bumping `PER_OP_MS` to `2_400` in my clone recomputes the ceiling to 47 automatically, so the refusal boundary and the scaled-timeout formula can never drift apart — the whole point of the finding.

**The `95` regression lock is the right kind of hardcode.** The *function* derives the value; the *test* pins `maxBatchOps()).toBe(95)` so a future constant change can't silently move the documented boundary without a human re-confirming it. Production auto-tracks, the test forces the conversation — that is the correct division, not a drift risk.

**Refusal fires at admission, before any side effect.** `execute()` calls `assertBatchAdmissible(ops.length)` *between* `loadOps` and the `runner(...)` dispatch, and the integration test proves it: an M+1 batch rejects with the runner's `called` flag still false — no job created, no frame sent, so there is no double-apply exposure (the exact risk a mid-flight wire timeout carries).

**`E_INVALID_ARGS` is the stable, correct code and the message is actionable.** It names all three numbers — the offending op count, the `120s` ceiling, and the max that fits (`≤95`) — with a concrete next step ("split into batches of ≤95"). The test asserts each of the three substrings, not just that it threw.

**The boundary is asserted at the exact edge, off the live constant.** 95 (`maxBatchOps()`) is accepted and dispatches; `maxBatchOps() + 1` is refused. The tests read `maxBatchOps()` rather than a literal, so they exercise whatever the constants currently produce; and the arithmetic checks out independently — `95×1200+5000 = 119000 ≤ 120000`, `96×1200+5000 = 120200 > 120000`.

**The scaled-timeout cap is unchanged for admitted batches.** `batchTimeoutMs` still `Math.min(CAP_MS, …)`, so a ≤95 batch gets its correctly-scaled wait exactly as PR #14 shipped it — and by construction an admitted batch's scaled value (≤119000) never reaches the clamp, so the belt-and-suspenders cap is inert on the admitted path rather than silently truncating anything.

**Empty and tiny batches behave sanely — no off-by-one.** An empty array is rejected earlier by `loadOps` ("non-empty JSON array"), so it never reaches the ceiling check. A 1-op batch passes `1 > 95 → false` and gets `max(60000, 6200) = 60000` — today's floor, unchanged. The floor never refuses a legitimate small batch; the ceiling never fires below 96.

**Scope is disjoint as claimed** — `cli/src/commands/batch.ts` plus one test file, no `broker-daemon.ts` change, so it cannot collide with the separately-tracked issue #7 work.

## Non-blocking

*Informational only:* `maxBatchOps` floors the division, so at the boundary the admitted budget is `119000`ms against a `120000`ms cap — a 1000ms slack that simply falls out of the floor. Intended and harmless (erring toward admitting slightly less than the cap), noted only so the 1s gap isn't mistaken for an oversight.

Nothing outstanding.

PR #19, opened 2026-07-31
feat: feed routing hygiene — cwd-independent staging + error-log binding

### LINKED ISSUE #7: Feed routing hygiene — cwd-independent staging root (5.6) + error-log binding (5.9)

jangtrinh:
Closes #7.

## Summary
This closes the cwd-misattribution class across all four broker write surfaces — DOC_CHANGE and EDIT_FEED already routed through the binding index; this closes the last two (backlog 5.6, 5.9).

- **5.6 — cwd-independent unbound staging root.** `unboundStagingPath`/`unboundEditStagingPath` used to root at `changeLogDir()` (the broker's own spawn cwd) — a restart into a different cwd orphaned whatever was staged there. Deferred at RI-P1 fix#1 time specifically because it meant moving an already-shipped, review-verified mechanism ("its own small spec/review, not a rider on whichever wave happens to touch this file next" — that review is this PR). New `unboundStagingRoot()`: `FIGMA_AGENT_UNBOUND_DIR` env override, else a fixed `/tmp/figma-agent-unbound` default, same convention as `project-bind.ts`'s `bindCacheFile()`. **One-time startup migration** (ruling: migrate, never silently orphan): a broker upgrading from before this fix scans the OLD location once and migrates anything found into the new root, reusing `migrateStagedChanges` as-is, guarded by a breadcrumb so future restarts skip the scan.
- **5.9 — error log gets full binding-aware routing.** `errorLogPath()` used to be resolved ONCE at broker startup and shared by every file/project the broker ever relayed for — no per-event resolution at all, unlike DOC_CHANGE/EDIT_FEED. Now: `errorLogPathFor(projectDir)` / `unboundErrorStagingPath(slug)`, resolved fresh per `ReplyErr` (`fileIdentity` + `resolveProjectDir`, same shape as the other two), with staging + migrate-on-bind parity (`migratedErrorCount`).
- **Minor:** `appendErrorFrame` now calls `rotateIfNeeded` — `log-rotate.ts`'s own module header always named the error log as one of the three feeds its policy covers, but the call site was never actually wired, leaving it the one unbounded-growth feed of the three.
- `shared/protocol.ts`: `FileContext.fileKey`'s comment corrected (was "carried, never used for routing" — now used for routing on the error-log path specifically; one-line comment fix, no schema change).

## Test plan
- [x] Test-first where it mattered: every existing unbound-staging test that hardcoded the old root's path was run RED against the new source first (confirmed failing for the expected reason), then updated.
- [x] 26 new/changed tests: `unboundStagingRoot` (2), `migrateLegacyUnboundChanges` (3), `migrateLegacyUnboundEdits` (2), `errorLogPathFor`/`unboundErrorStagingPath` (3), the rotation-wiring integration test (1), plus root-relocation fixes across the existing staging-path assertions, and 2 new end-to-end `broker-daemon-harness.test.ts` scenarios (error routing stage→bind→migrate→direct, and the one-time legacy-migration-on-startup).
- [x] Every test file that exercises unbound staging now isolates `FIGMA_AGENT_UNBOUND_DIR` via a dedicated scratch dir in `beforeEach`/`afterEach`, matching this repo's existing env-swap convention — confirmed no bleed into the real `/tmp/figma-agent-unbound` default (checked directly after each local run).
- [x] `npm run build` — clean (cli 235.4kb, plugin 159.0kb — plugin unchanged, no plugin-side source touched)
- [x] `npm test` — 87 files / 1129 tests green (was 85/1103 on latest main before this branch; +26 explained above, no unexplained drift)
- [x] No `npm run lint` script exists in this repo — consistent with prior gates, skipped.

No live-canvas verification needed — this phase is entirely broker/CLI-side (no plugin change), and every scenario (staging, migration, rotation, routing) is exercised end-to-end via the real in-process broker + real `ws` sockets in `broker-daemon-harness.test.ts`, not simulated.

Holding for review.

### DIFF
--- a/cli/src/commands/status.ts
+++ b/cli/src/commands/status.ts
@@ -36,6 +36,10 @@ export async function run(args: CommandArgs): Promise<unknown> {
     // own byte-identical-when-zero contract: present only once a cross-instance reply
     // has actually been discarded, so the common (zero) case stays unchanged here too.
     ...(typeof hello.senderMismatchCount === 'number' && { senderMismatchCount: hello.senderMismatchCount }),
+    // Issue #7 fix round — same mirror-only-when-true contract: an operator running
+    // `status` sees this the moment the one-time legacy-staging migration deferred,
+    // without needing to grep the broker log.
+    ...(hello.legacyMigrationDeferred === true && { legacyMigrationDeferred: true }),
   };
   const all = Array.isArray(hello.plugins) ? (hello.plugins as PluginStatusEntryWithJob[]) : [];
 
--- a/cli/src/transport/broker-daemon.ts
+++ b/cli/src/transport/broker-daemon.ts
@@ -367,15 +367,31 @@ export async function runBrokerDaemon(options?: BrokerDaemonOptions): Promise<vo
   // the WS server accepts any connection (below), so no live DOC_CHANGE/EDIT_FEED write
   // can race it. Guarded by a breadcrumb under the NEW root so a broker that already
   // ran this once doesn't pay a readdir cost on every future restart.
+  //
+  // Stage-4 fix (reviewer finding 1) — START-AND-DEFER, with loud visibility, never
+  // ABORT: this is a best-effort move of already-safe, already-retryable UNBOUND staging
+  // data (copy-before-delete in migrateStagedChanges), not a reason to keep the whole
+  // relay from ever accepting a connection. A readdir/append/mkdir failure here (a full
+  // or unwritable /tmp) is caught, logged clearly, and the breadcrumb is deliberately
+  // NOT written — the next restart retries. `legacyMigrationDeferred` is surfaced in
+  // BROKER_HELLO/status (same "discarded thing gets a machine-readable counter" pattern
+  // as `senderMismatchCount`, backlog 2.10/issue #15) so an operator can't miss that
+  // unbound data is still stuck, even though the broker itself came up fine.
+  let legacyMigrationDeferred = false;
   const legacyMigratedMarker = join(unboundStagingRoot(), '.legacy-migrated');
   if (!existsSync(legacyMigratedMarker)) {
-    const migratedLegacyChangeFiles = migrateLegacyUnboundChanges();
-    const migratedLegacyEditFiles = migrateLegacyUnboundEdits();
-    if (migratedLegacyChangeFiles > 0 || migratedLegacyEditFiles > 0) {
-      log(`legacy unbound staging migrated to the new root: ${migratedLegacyChangeFiles} change file(s), ${migratedLegacyEditFiles} edit file(s)`);
+    try {
+      const migratedLegacyChangeFiles = migrateLegacyUnboundChanges();
+      const migratedLegacyEditFiles = migrateLegacyUnboundEdits();
+      if (migratedLegacyChangeFiles > 0 || migratedLegacyEditFiles > 0) {
+        log(`legacy unbound staging migrated to the new root: ${migratedLegacyChangeFiles} change file(s), ${migratedLegacyEditFiles} edit file(s)`);
+      }
+      mkdirSync(unboundStagingRoot(), { recursive: true });
+      writeFileSync(legacyMigratedMarker, JSON.stringify({ at: Date.now() }), 'utf8');
+    } catch (err) {
+      legacyMigrationDeferred = true;
+      log(`LEGACY UNBOUND STAGING MIGRATION FAILED, deferred to next restart (data untouched, not lost): ${(err as Error).message}`);
     }
-    mkdirSync(unboundStagingRoot(), { recursive: true });
-    writeFileSync(legacyMigratedMarker, JSON.stringify({ at: Date.now() }), 'utf8');
   }
 
   const st: BrokerState = {
@@ -1309,7 +1325,7 @@ export async function runBrokerDaemon(options?: BrokerDaemonOptions): Promise<vo
       st.registry,
       {
         port, pid: process.pid, protocolV: PROTOCOL_VERSION, buildMtime: selfBuildMtime(),
-        uptimeMs: Date.now() - startedAt, senderMismatchCount,
+        uptimeMs: Date.now() - startedAt, senderMismatchCount, legacyMigrationDeferred,
       },
       currentFilter(),
       Date.now,
--- a/cli/src/transport/broker-daemon.ts
+++ b/cli/src/transport/broker-daemon.ts
@@ -7,7 +7,8 @@
 // persistent broker-daemon pattern (one long-lived relay process, hot-swappable
 // across CLI rebuilds), adapted from southleft/figma-console-mcp's websocket-server
 // pending-request correlation (347-360) / heartbeat (672-685).
-import { appendFileSync, readFileSync, unlinkSync } from 'node:fs';
+import { appendFileSync, existsSync, mkdirSync, readFileSync, unlinkSync, writeFileSync } from 'node:fs';
+import { join } from 'node:path';
 import WebSocket, { WebSocketServer } from 'ws';
 import {
   BROKER_FILE, BROKER_IDLE_SHUTDOWN_MS, EXEC_JS_MAX_TIMEOUT_MS, HEARTBEAT_INTERVAL_MS, LOOPBACK_HOST, PLUGIN_WAIT_MS,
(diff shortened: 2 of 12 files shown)


jangtrinh (commented):
**Review verdict: APPROVE** — the cwd-misattribution WRITE class is genuinely closed across all four surfaces through one shared root + one reused migration; I grepped for a fifth surface and none remains. Migration crash-safety, 'unknown'-bucket containment, reader coherence, and test isolation all verified. One non-blocking robustness note (unguarded startup migration can abort broker start). (Posted as a review comment: GitHub refuses a formal approval state on a self-authored PR.)

Reviewed `c341e0a` on a fresh clone; 71/71 across the four affected suites, and I confirmed the real `/tmp/figma-agent-unbound` was never created by the run (test isolation is genuine). I went looking specifically for the surface this did NOT close, since it claims to close a whole bug class. I could not find a fifth write surface — the thesis holds. Approving, with non-blocking notes.

## The anti-patchwork claim is real

All three staging paths now resolve through one `unboundStagingRoot()` (`FIGMA_AGENT_UNBOUND_DIR` else a fixed `/tmp/figma-agent-unbound`), keeping their own disjoint sub-paths (`unbound/`, `changes/unbound/`, `errors/unbound/`) so the three logs still can't cross-contaminate. The error log gains true parity: `errorLogPathFor`/`unboundErrorStagingPath` mirror the change/edit shape exactly, the `!msg.ok` branch resolves `fileIdentity → resolveProjectDir` **per message** (the once-cached `errorsPath` is gone), and `handleProjectBind` migrates staged errors with a `migratedErrorCount`. `appendErrorFrame` now calls `rotateIfNeeded`, closing the one unbounded feed. One root, one migration function reused three times — this is consolidation, not another patch.

**I grepped every `appendFileSync`/`writeFileSync` in `cli/src/transport/` and every `changeLogDir()`/`process.cwd()` consumer.** The only capture-write surfaces are the three feeds, all now routed; `SYNC_REQUEST` already resolves its project via `resolveProjectDir` and refuses when unbound; the advertisement, broker log, bind marker/cache, and mirror-capture temp file are not per-project capture data. No fifth misattribution surface exists.

## Focus findings

**Migration crash-safety holds across the legacy path (item 1+2).** `migrateStagedChanges` is copy-before-delete with a hash marker AND a `boundLogEndsWith` belt check, so a crash after the new-root append but before the marker/cleanup re-reads the old file, sees the content already present at the destination, and skips — no double-append. Because startup migration runs before the WS server accepts connections, nothing can append to the new-root staging between a crash and the restart, so the belt check is always decisive. The `.legacy-migrated` breadcrumb is keyed to the NEW root and the legacy *scan* reads the OLD cwd-relative dir, whose files are deleted on first successful move — so a changed `FIGMA_AGENT_UNBOUND_DIR` cannot cause a re-migration (old dir is already empty) and the marker dedup covers it even if it weren't. Neither silent-skip nor double-append is reachable.

**The 'unknown' bucket stays contained (item 3).** A `!msg.ok` with no fileKey and no fileName resolves to `fileIdentity(null,null) = 'unknown'`; `resolveProjectDir('unknown', …)` can only match a binding literally keyed `unknown`, which no real project has — so it falls to `unboundErrorStagingPath('unknown')` and can never reach a real project's error log. Same containment the registry wave established, and the same per-message shape the other two feeds use.

**Reader coherence is preserved (item 5).** The `errors` reader intentionally stays on `errorLogPath()` = `changeLogDir()/figma-errors.jsonl` = `<cwd>/design/…`, which equals the writer's `errorLogPathFor(projectDir)` whenever the operator runs `figma-agent errors` from the project dir — exactly the cwd==projectDir coherence the `changes`/`edit` readers already rely on, and the doc comment states the reader default is deliberately unchanged. Writer and reader coincide; they did not diverge.

**Test isolation is genuine (item 6).** Every new/changed suite sets `FIGMA_AGENT_UNBOUND_DIR` to a scratch dir in `beforeEach` and restores it in `afterEach`; the harness threads it into the in-process daemon before startup; the one default-value test only reads the string, never writes. The real `/tmp` path stayed absent through the whole run.

## Non-blocking

1. **The startup migration is unguarded — a staging-move I/O error aborts broker startup.** `broker-daemon.ts:369-380`: `migrateLegacyUnboundChanges()` / `migrateLegacyUnboundEdits()` run inline with no `try`/`catch`, before the WS server binds. A `readdirSync`/`appendFileSync`/`mkdirSync` failure on the new root (a full or unwritable `/tmp`) therefore propagates out of daemon startup, so the broker never begins accepting connections — over a best-effort move of *unbound* staging data. Copy-before-delete already guarantees the data is safe and retryable, so the loud failure buys nothing here that a logged, deferred retry wouldn't. Suggest wrapping the two calls: on failure, log and continue starting the broker, and do **not** write the breadcrumb (so the next start retries). That decouples broker availability from a staging-migration hiccup — the same "a log write must never disrupt the relay" contract the append paths themselves already hold.

2. *Informational:* `figma-sync-config.ts`'s `syncConfigPath()`/`projectDir()` is the one remaining cwd-relative resolution in the broker, reading `design/figma-sync.json` (the idle window) from `changeLogDir()`. It is a read of a low-stakes config value, not a capture-write surface, and its own comment documents keeping it — so it is genuinely outside this PR's write-misattribution class, but it is the last cwd dependency in the broker and may deserve its own tracking note.

3. *Informational:* the error path's staging KEY is `safeSlug(fileName)` while its resolve identity is `fileIdentity(fileKey, fileName)` (fileKey-preferring) — so a hypothetical fileKey-set/fileName-null unbound error would stage under `unknown` and not be found by a later name-slug-keyed bind migration. This asymmetry is identical across all three feeds (parity is the goal and is achieved), and the case is near-unreachable since a connected plugin's scene almost always carries `figma.root.name`. Noted for completeness, not as a regression.

4. *Test coverage:* `migrateLegacyUnboundChanges/Edits` are unit-tested directly and the harness asserts the `.legacy-migrated` breadcrumb lands, but no test pre-seeds the OLD legacy dir and then boots the broker to prove startup migration carries that data into the new root end to end. Worth one test, given this is the load-bearing migration.

Verdict below.

jangtrinh:
Fix round pushed (1bca6ca), addressing the blocking finding + fold-ins:

- **Finding 1** (unguarded migration could abort startup): wrapped in try/catch. On failure, logged clearly, the breadcrumb is deliberately NOT written (next restart retries), and the broker proceeds to accept connections — never aborts over a best-effort move of already-safe, already-retryable staging data. `legacyMigrationDeferred` surfaces in BROKER_HELLO/status, same "present only when true" contract issue #15 established for `senderMismatchCount`.
- **Fold-in (informational note 3):** the legacy-migration harness test now pre-seeds BOTH the change-log's and the edit feed's old locations in one boot (was change-log only).
- Merged latest main (through PR #18) to pick up #15's surfacing pattern directly rather than re-deriving it.

Test-first: the new failure-resilience test (a plain file where the legacy dir is expected — `readdirSync` throws `ENOTDIR` for real, no mocking) confirmed to abort startup against the pre-fix code, then confirmed green after the fix.

Gate: typecheck clean, compile clean, 88 files / 1143 tests (delta vs. 1129 includes tests brought in by the main merge, not just my own 4 new tests — no unexplained drift).

`syncConfigPath`'s remaining cwd-relative read: filing as a separate low-stakes board issue now, not folding into this PR.

Holding for re-review.

jangtrinh (commented):
**Follow-up verdict: APPROVE** — finding 1 is closed. Start-and-defer verified: the catch is scoped to only the migration, the broker completes a full plugin handshake after a real ENOTDIR migration failure, the breadcrumb is withheld so the next boot retries (copy-before-delete + hash-marker still gate double-append), the deferral flag is byte-identical-absent on the happy path, and both feeds are proven to land in the new root end to end. The deferral test discriminates (rethrow → red, defer → green). Whole cwd-misattribution write class is shut. (Posted as a review comment: GitHub refuses a formal approval state on a self-authored PR.)

Re-reviewed the finding-1 fix at `1bca6ca` (with main merged through #18) on a fresh clone: 26/26 across the harness + status suites, and the real `/tmp/figma-agent-unbound` stayed absent through the run. The fix is exactly the start-and-defer shape I asked for, and its two new tests genuinely discriminate. Closing my finding.

## Verified

**The catch wraps only the migration, not unrelated init.** The `try` in `broker-daemon.ts:381-393` spans exactly `migrateLegacyUnboundChanges()`, `migrateLegacyUnboundEdits()`, the success log, the `mkdirSync(root)`, and the breadcrumb `writeFileSync` — nothing else. `st` construction, the WS bind, and every other startup step are outside it, so an unrelated startup failure still surfaces normally rather than being mislabeled a deferred migration.

**Defer-and-retry holds, and the deferral test discriminates.** The ENOTDIR test plants a plain file where `legacyUnboundStagingDir()` expects a directory — a real `readdirSync` failure, no mocking — then asserts `legacyMigrationDeferred: true` on the greeting AND that the broker completes a full plugin handshake (`helloPlugin` after `connectSocket`), i.e. genuinely serving, not merely answering HELLO. I confirmed it isn't a tautology: rethrowing the caught error instead of deferring turns that exact test red (startup aborts), and restoring makes it green. The breadcrumb is asserted absent on failure, so the next boot retries; and because the deferral leaves `migrateStagedChanges` untouched, its copy-before-delete + hash-marker + `boundLogEndsWith` guards still gate any re-run against a double-append — the crash-mid-migration case I cleared last round is unchanged.

**`legacyMigrationDeferred` is byte-identical-absent on the happy path.** Both spreads are conditional on the boolean being true (`...(meta.legacyMigrationDeferred && …)` in the envelope, `...(hello.legacyMigrationDeferred === true && …)` in `status.ts`), so a broker that migrated cleanly or had nothing to migrate — the overwhelming common case — emits the same payload as before the field existed. It follows issue #15's `senderMismatchCount` present-only-when-set pattern exactly.

**Both feeds land in the new root, proven end to end.** The pre-seed-both test writes leftover staging at both old locations (`<changes>/unbound/…` and `<changes>/changes/unbound/…`) before boot, then asserts both files appear under the new root with their exact content, both old files are cleaned up, and the breadcrumb is written — one boot, both feeds, real filesystem.

That closes finding 1. The whole cwd-misattribution write class is now genuinely shut across all four surfaces, the one-time migration is unified and crash-safe, and its one operational failure mode is non-fatal and visible. No further concerns from me — good to merge and close the class in main.

The earlier informational notes remain informational and unchanged (the read-only `figma-sync-config` cwd dependency; the near-unreachable fileKey-only staging-key asymmetry shared across all three feeds). Neither blocks.

PR #22, opened 2026-07-31
fix: route SYNC_CONFIG idleMs through the file's project binding

### LINKED ISSUE #20: syncConfigPath: last cwd-relative resolution in the broker
From PR #19 review (informational, out of that PR's class). figma-sync-config.ts's syncConfigPath() is the last cwd-relative path resolution left in the broker after the feed-routing-hygiene wave closed the cwd-misattribution class across the four capture-WRITE surfaces (DOC_CHANGE, EDIT_FEED, unbound staging, error log). This one is a low-stakes config READ, not a per-project capture-write, so it was correctly left out of #7 — but it's the last cwd-relative resolver and deserves its own audit: does any multi-project/cwd-restart scenario make it read the wrong project's sync config? If yes, route it through the binding like the rest; if it's genuinely single-project-only by contract, document that in-code so the next reader doesn't file it as a bug.

jangtrinh:
## Audit finding (issue #20)

`syncConfigPath()`/`readIdleMs()` was the last cwd-relative path resolver left in the
broker after #19 closed the cwd-misattribution class across the four capture-WRITE
surfaces. The hole is real, not a documented single-project contract:

- The broker is a multi-project daemon by design (`project-bind.ts`'s `bindIndex`
  routes many bound projects through one long-lived process — see `loadBindIndex`,
  the `/tmp/figma-agent-binds.json` restart-survival cache, `PROJECT_BIND`).
- `idleMs` was resolved ONCE at broker startup from the broker's own spawn cwd (or
  `FIGMA_AGENT_CHANGES_DIR`) and sent unconditionally to every `PLUGIN_HELLO`,
  regardless of which project that plugin's file was bound to.
- Concretely: a broker spawned from project A's cwd (its own `design/figma-sync.json`)
  serving a plugin bound to project B hands B's plugin A's idle window — and even an
  already-connected plugin never picks up its own project's config after a
  `PROJECT_BIND` resolves it, since `SYNC_CONFIG` is only ever sent once, at HELLO.

Two discriminating tests in `tests/broker-daemon-harness.test.ts` reproduce this
against pre-fix code (both fail: one asserts `222_000`, gets the broker-cwd default
`111_000`; the other times out because no second `SYNC_CONFIG` is ever sent after
bind).

## Fix

- `cli/src/transport/figma-sync-config.ts`: add `syncConfigPathFor(projectDir)` /
  `readIdleMsFor(projectDir)`, mirroring `change-log.ts`'s `changeLogPathFor()`
  convention. `syncConfigPath()`/`readIdleMs()` stay as the documented GLOBAL fallback
  for a genuinely unbound file.
- `cli/src/transport/broker-daemon.ts`:
  - `PLUGIN_HELLO` now resolves the file's binding first (same
    `fileIdentity`→`resolveProjectDir` shape the error-log/edit-feed paths already use)
    and sends that project's own idle window, falling back to the broker-cwd default
    only when unbound.
  - `handleProjectBind` pushes a fresh `SYNC_CONFIG` to an already-connected plugin the
    moment its file is bound, so the correct idle window takes effect without a
    reconnect.

No changes to `plugin/` or the pinned `kernel/design-os` submodule.

## Gates (in the worktree)

- `npm run typecheck` — pass
- `npm run build` — pass
- `npm test` — 88 files / 1145 tests pass (includes the panel gate)

Closes #20

### DIFF
--- a/cli/src/transport/broker-daemon.ts
+++ b/cli/src/transport/broker-daemon.ts
@@ -34,7 +34,7 @@ import {
   appendEditFrames, editFeedPathForIdentity, migrateLegacyUnboundEdits, safeSlug, unboundEditStagingPath,
 } from './edit-feed-log.ts';
 import { appendErrorFrame, buildErrorLogFrame, errorLogPathFor, unboundErrorStagingPath } from './error-log.ts';
-import { readIdleMs } from './figma-sync-config.ts';
+import { readIdleMs, readIdleMsFor } from './figma-sync-config.ts';
 import { spawnReconcileApply } from './figma-sync-apply.ts';
 import {
   fileIdentity, loadBindIndex, needsAliasPromotion, readBindMarker, recordBinding, removeBinding,
@@ -642,6 +642,11 @@ export async function runBrokerDaemon(options?: BrokerDaemonOptions): Promise<vo
     // unbound staging, giving it full parity with the other two feeds.
     const migratedErrorCount = migrateStagedChanges(unboundErrorStagingPath(slug), errorLogPathFor(projectDir));
     log(`BIND recorded: "${fileName}" → ${projectDir}${fileKey ? ` (fileKey ${fileKey})` : ' (pending fileKey)'}${migratedCount > 0 ? `, migrated ${migratedCount} staged change frame(s)` : ''}${migratedEditCount > 0 ? `, migrated ${migratedEditCount} staged edit frame(s)` : ''}${migratedErrorCount > 0 ? `, migrated ${migratedErrorCount} staged error frame(s)` : ''}`);
+    // Issue #20 fix: an ALREADY-CONNECTED plugin for this file received SYNC_CONFIG at
+    // HELLO time, before this binding existed — its idle timer is still running on the
+    // broker's cwd default (or a stale project's). Push the newly-bound project's own
+    // idle window now, so it takes effect without waiting for a reconnect.
+    if (hit) sendEvent(hit.ws, 'SYNC_CONFIG', { idleMs: readIdleMsFor(projectDir) });
     reply({ fileName, projectDir, fileKey, pendingKey: fileKey === null, migratedCount, migratedEditCount, migratedErrorCount });
   };
 
@@ -1208,7 +1213,18 @@ export async function runBrokerDaemon(options?: BrokerDaemonOptions): Promise<vo
         log(`plugin registered [${instanceId}]${replaced ? ' (replaced — same instance re-hello)' : ''}: ${JSON.stringify(msg.data)}`);
         // Live-sync (spec 004 P4): hand this plugin the idle window so its debounce
         // timer matches the project's design/figma-sync.json.
-        sendEvent(ws, 'SYNC_CONFIG', { idleMs });
+        //
+        // Issue #20 fix: resolved from THIS file's binding, same fileIdentity→
+        // resolveProjectDir shape the error-log/edit-feed paths already use — never the
+        // once-cached, broker-cwd-derived `idleMs` unconditionally, or a broker spawned
+        // from one project's cwd hands every OTHER bound project's plugin the wrong idle
+        // window. A genuinely unbound file (no PROJECT_BIND yet) still falls back to the
+        // broker's own cwd default — there is nowhere else to read it from.
+        const helloScene = st.registry.getByWs(ws)?.scene;
+        const helloFileName = (helloScene?.fileName as string | undefined) ?? null;
+        const helloFileKey = (helloScene?.fileKey as string | null | undefined) ?? null;
+        const helloBound = resolveProjectDir(fileIdentity(helloFileKey, helloFileName), st.bindIndex);
+        sendEvent(ws, 'SYNC_CONFIG', { idleMs: helloBound ? readIdleMsFor(helloBound) : idleMs });
         flushWaiting(); // deliver any requests parked during the reconnect gap
         broadcastPeers();
       } else if (msg.type === 'PING') {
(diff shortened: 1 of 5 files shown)


jangtrinh (commented):
**Review verdict: APPROVE** — the audit is legitimate (a per-plugin READ member of the same cwd-misattribution class), the fix routes idleMs through the file's binding correctly (unbound→fallback, unknown→fallback, targeted post-bind push, last-wins double-send safe), and it closes the genuinely-last cwd resolver in the broker's per-plugin path. Both tests discriminate (111000-not-222000, and timeout). On #19: the WRITE-class claim still holds; the asterisk is on my own review's side-note that called this READ surface 'outside the class' — it wasn't. (Posted as a review comment: GitHub refuses a formal approval state on a self-authored PR.)

Reviewed `64cab1d` on a fresh clone; the harness suite is green and both new tests discriminate exactly as reported — reverting the HELLO change yields `111000` instead of `222000`, and dropping the post-bind push makes the second test time out. The audit is right, the fix is correct, and it closes the surface my own #19 review misjudged. Approving.

## The finding is legitimate — and it IS the same class

This is not a stretch to fit the theme. `readIdleMs()` was resolved ONCE at startup from `changeLogDir()` (the broker's spawn cwd) into a closure `idleMs`, then sent to EVERY plugin at HELLO regardless of which project its file was bound to — and never re-sent after a later `PROJECT_BIND`. On a multi-project broker (which `bindIndex`/`PROJECT_BIND` make real), a daemon spawned from project A's cwd handed project B's plugin project A's idle window. That is the identical cwd-misattribution shape as the write surfaces, just on a config READ that fans out per-plugin. My #19 note called this "low-stakes config-read outside the class" — that was wrong on the "outside the class" part, and the audit correctly caught it.

## Verified

**Unbound file at HELLO gets the fallback, never a throw or a wrong project.** `resolveProjectDir(fileIdentity(helloFileKey, helloFileName), st.bindIndex)` returns null for an unbound file, and the send falls back to the startup `idleMs` — `sendEvent(ws, 'SYNC_CONFIG', { idleMs: helloBound ? readIdleMsFor(helloBound) : idleMs })`. No throw, no cross-project read.

**The post-bind push is targeted, not broadcast.** `if (hit) sendEvent(hit.ws, 'SYNC_CONFIG', { idleMs: readIdleMsFor(projectDir) })` sends only to `hit` — the plugin whose fileName the bind resolved (`st.registry.matching(fileName, { exact: true })[0]`). A bind for project B pushes to B's socket alone; project A's live plugin is untouched. (The second test proves the bound plugin receives the update; see the one hardening note below on making the non-broadcast explicit.)

**HELLO uses the same identity chain as the write surfaces.** `fileIdentity(fileKey, fileName)` — fileKey wins, else name-slug, else `unknown` — and `resolveProjectDir('unknown', …)` matches only a literal `unknown` binding, which no real project has, so an unknown file takes the fallback and can never read a real project's config. Same containment the error-log/edit-feed paths use.

**No double-send hazard; the second event is authoritative.** In the HELLO-after-bind path the file is already bound, so HELLO sends the correct value directly (one event). In the connect-then-bind path HELLO sends the fallback and the bind sends the project value second. On a single socket these are strictly ordered — a bind can only match a plugin that already completed HELLO/registration, so its push always follows the HELLO push — and the plugin applies SYNC_CONFIG as a plain last-wins assignment (`main.ts:415-417`, floored at `MIN_IDLE_MS`), so receiving two is fine and the later (project) value wins.

**This is genuinely the last cwd-relative resolver in the broker.** I grepped every `changeLogDir()`/`process.cwd()`/`projectDir()` consumer. What remains is: the reader defaults (`errorLogPath`/`editFeedDir`/`changeLogPath`) — cwd-relative by design so a CLI run inside the project reads its own feed, the coherence #19 established and the writers now bypass; `legacyUnboundStagingDir` — migration source only; the CLI-side cwd reads in `figma-agent.ts`/`sync-corrections.ts`/`capture.ts` — correct, those run in the operator's project; and `readIdleMs()`/`syncConfigPath()`, now documented and used solely as the unbound fallback. `figma-sync-config.ts`'s `projectDir()` now has NO live caller at all (only a comment reference in `project-bind.ts`) — dead code this PR could delete. There is no config-read #6 in the broker's per-plugin path.

## Non-blocking

1. **Strengthen the non-broadcast assertion.** The post-bind test has a single plugin, so it proves the bound plugin *receives* the update but not that an *unrelated* project's plugin does *not*. The code is correct by construction (`sendEvent(hit.ws, …)`), but a second connected plugin for a different file, asserted to receive no SYNC_CONFIG on B's bind, would lock the targeting against a future refactor to a broadcast helper.

2. *Cleanup:* `projectDir()` in `figma-sync-config.ts` is now unreferenced (grep finds only the `project-bind.ts` comment). Worth deleting so the next reader doesn't mistake it for a live per-project resolver.

## On the #19 "anti-patchwork" claim — does it need an asterisk?

Partly. #19's own words were "closes the cwd-misattribution WRITE class across all four surfaces," and that specific claim still holds — the four capture-WRITE feeds are shut. What was over-broad was my review's side-note dismissing `syncConfigPath` as "outside the class": it was a READ surface of the *same* class, and the class was defined by cwd-vs-binding misattribution, not by write-vs-read. So the honest framing is: #19 closed the write members and correctly scoped its own claim to them; the thesis ("resolve per-binding, never per-cwd") was right and simply had one more member than either the PR or my review enumerated. This PR extends the same principle to the read surface rather than contradicting it — consolidation, not a patch on a patch. The asterisk belongs on the *enumeration* ("we found all the surfaces"), not on the *principle* or on #19's write-scoped claim.

Verdict below.

jangtrinh:
Both addenda pushed (f010c20):

1. Strengthened the non-broadcast test: added an unrelated plugin bound to a different project (`Project Other File` → `otherProjectDir`), collects its frames from after its own bind settles, and asserts it receives no `SYNC_CONFIG` when Project C's bind fires the post-bind push — proves targeting positively.
2. Dropped the now-dead `projectDir()` in figma-sync-config.ts (and its test) — updated project-bind.ts's historical comment to reference `changeLogDir()` instead so it doesn't dangle on a removed name.

Gates: typecheck pass, build pass, `npm test` 88 files / 1144 tests pass.

PR #23, opened 2026-07-31
Polish follow-ups from PR #10 review

### LINKED ISSUE #11: Polish follow-ups from PR #10 review
Four non-blocking items from the stage-4 re-review of PR #10 (all cited in the follow-up review comment there):

1. `cleanup()` in exec-stdlib-component-build.ts is unguarded — a throwing cleanup would mask the ORIGINAL error; wrap it so the first error always propagates. Its 'safe to call multiple times' claim currently holds against the permissive mock `remove()`, not proven against real Figma.
2. `knowledge/component-sets.md` (~:35-42) reads as full restoration, but a verify-mismatch leaves the combined set standing with the base inside it — say exactly what is and isn't restored.
3. `buildModeB` renames as it validates — a mid-loop rejection leaves earlier renames with no closure yet; validate fully before the first rename.
4. Two new file headers cite a '200-line cap' that no repo doc states, while exec-stdlib-variables.ts is 205 lines — either add the convention to CLAUDE.md or drop the citations.

jangtrinh:
## Summary

Four non-blocking polish items from PR #10's stage-4 re-review (issue #11):

1. **cleanup() masking the original error** — `componentSet()`'s catch block called
   `build.cleanup()` unguarded; if cleanup itself threw (a locked node, a clone already
   removed elsewhere), that new error replaced the original failure the caller was
   trying to see. Now wrapped: a cleanup failure is attached (`cleanupError`), never
   substituted — the original error's code and message always propagate.
2. **`knowledge/component-sets.md` over-stated restoration** — reworded to say exactly
   what cleanup restores (build's own renames/clones) and what it doesn't (a
   verify-mismatch after a successful `combineAsVariants` leaves the combined
   COMPONENT_SET standing on the canvas, not un-combined).
3. **`buildModeB` renamed as it validated** — a mid-loop rejection could leave earlier
   components already renamed with no cleanup, since the failure happened before a
   `BuildResult` (and its `cleanup()`) even existed. Split into two passes: validate
   every input fully, then rename.
4. **"200-line cap" citation with no repo doc backing it** — dropped from the two
   PR #10 headers (`exec-stdlib-component-build.ts`, `exec-stdlib-component-matrix.ts`).
   No doc in this repo states the convention, and it isn't enforced (a sibling stdlib
   file already runs over it, uncited).

Closes #11.

## Test plan

- Both code fixes (items 1 and 3) are test-first: added tests reproduce the bug and
  fail against pre-fix code, then pass after the fix (verified locally).
- `npm run typecheck` — clean
- `npm run build` — clean (regenerated `plugin/code.js` / `plugin/ui.html`, committed
  build outputs per this repo's convention)
- `npm test` — 88 test files, 1146 tests, all green (includes the panel gate against
  the pinned `kernel/design-os` submodule)

### DIFF
--- a/knowledge/component-sets.md
+++ b/knowledge/component-sets.md
@@ -32,14 +32,28 @@ return await ui.componentSet({
 });
 ```
 
-- **Run this behind `--undo-group`.** `ui.componentSet` cleans up its OWN mutations
-  (restores a renamed base, removes the clones it created) if one of ITS OWN later
-  steps throws — that is a narrow, self-scoped safety net, not a substitute for the
-  real undo bracket. If a DIFFERENT step in the same script fails after
-  `ui.componentSet` already returned successfully, `--undo-group` is what rolls the
-  whole script back; without it, a successfully-built set stays on the canvas even
-  if a later line in the same script throws. Always wrap a script that calls this
-  in `exec-js --undo-group` unless you specifically want partial results to survive.
+- **Run this behind `--undo-group`.** `ui.componentSet` cleans up its OWN mutations —
+  and ONLY those — if one of ITS OWN later steps throws (a size mismatch, a name that
+  didn't parse back to the intended axes, `combineAsVariants` itself rejecting the
+  call). Say precisely what that means, because "cleans up" understates it:
+  - **IS restored**: the base's original name (mode 1), and each combined component's
+    original name (mode 2) — whatever `buildModeA`/`buildModeB` renamed gets renamed
+    back. A clone `buildModeA` created for the matrix is removed.
+  - **IS NOT restored**: if `combineAsVariants` already ran before the failure (i.e. the
+    throw came from the post-combine verification, not from `combineAsVariants` itself),
+    the COMPONENT_SET it created stays on the canvas — cleanup renames/removes the
+    build's own nodes, it does not un-combine Figma's structure or delete the set.
+    Concretely: a verify-mismatch after a successful `combineAsVariants` leaves the
+    combined set standing. In mode 1, the base (renamed back to its original name)
+    remains as the set's only child once the clone(s) built for the matrix are removed.
+    In mode 2, every original component stays parented inside the set — each renamed
+    back to its own original name, but the set itself and its full membership persist.
+  - This is a narrow, self-scoped safety net for build's OWN pre-combine mutations, not
+    a substitute for the real undo bracket. If a DIFFERENT step in the same script fails
+    (including this leftover COMPONENT_SET case above), `--undo-group` is what rolls the
+    whole script back; without it, a successfully-built set — or the orphaned one left
+    by a verify-mismatch — stays on the canvas. Always wrap a script that calls this in
+    `exec-js --undo-group` unless you specifically want partial results to survive.
 - **Exactly one of `base`+`axes` or `components`** — both or neither throws, naming the
   two valid shapes.
 - **Axis and value names must not contain `=` or `,`** — Figma parses variant properties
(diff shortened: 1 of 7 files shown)


jangtrinh (commented):
**Review verdict: APPROVE** — all four follow-ups done; the cleanup-masking bug (item 1, flagged 3x across #12/#13/#19) is genuinely closed and its test discriminates (revert → cleanup boom masks E_EVAL). Item 3 is fail-before-mutate for real (revert → CompA renamed). Doc reword is accurate to what stays on canvas; typecheck clean. One non-blocking: the cleanupError attach itself isn't guarded against a frozen err (not reachable here, but wrapping it makes the un-masking airtight). (Posted as a review comment: GitHub refuses a formal approval state on a self-authored PR.)

Reviewed `9a7ca82` on a fresh clone; 25/25 on the componentSet suite, `tsc -p plugin` clean, and both code fixes discriminate under revert. All four follow-ups are genuinely done. The one that matters — the cleanup-masking bug flagged three times now across #12/#13/#19 — is closed. Approving.

## Verified

**Item 1 (the real fix) is closed, and its test is not a tautology.** The catch now wraps `build.cleanup()` in its own `try`; a throwing cleanup attaches as `err.cleanupError` and the ORIGINAL error is rethrown as the same object — `throw err`, never a reconstruction, so its `code` and `message` are byte-preserved. The test forces a genuine E_EVAL verify-mismatch (combineAsVariants succeeds but drops a child) AND makes `cleanup()` throw (`FakeNode.prototype.remove` → "cleanup boom"), then asserts the caller sees `E_EVAL … "componentSet combined"`. I reverted the try/catch to the old `build.cleanup(); throw err;` and that exact test goes red — the "cleanup boom" error masks the E_EVAL, precisely the class bug — and restoring makes it green.

**Item 3 is fail-before-mutate for real.** `buildModeB` is now two passes: pass 1 resolves every id, type-checks it, rejects a COMPONENT_SET member, and runs `assertCleanToken` on all keys/values — with zero `node.name =` in it; pass 2 does the renames only after every input passed. I reverted it to rename during the validation pass and the test goes red (`CompA` becomes `State=default` when `component[1]`'s banned token is rejected); restored, green. Pass 2 itself contains no throwing path, so nothing can rename-then-fail.

**Item 2 doc is accurate, not aspirational.** The reword states exactly what the code leaves: after a post-combine verify-mismatch the COMPONENT_SET stays on canvas, the base/components are renamed back, and mode-1 clones are removed — which matches `buildModeA`/`buildModeB`'s `cleanup()` (restore names, remove clones; never un-combine or delete the set). The "IS / IS NOT restored" split is correct against the actual behavior.

**Item 4 is done for its stated scope.** The "200-line cap" citation is gone from both componentSet headers (build + matrix), reworded to "one shape". No new TS holes; the `as ComponentNode` casts sit after the type-narrowing checks.

## Non-blocking

1. **The masking guard has one residual gap — apt to close given the theme.** The `cleanupError` attachment `(err as { cleanupError?: unknown }).cleanupError = cleanupErr` is not itself protected: if `err` were ever a frozen / non-extensible object, that assignment throws a TypeError in strict mode (which plugin ESM is), and THAT would propagate instead of `err` — re-masking the original, the exact failure this fix exists to prevent. It is not reachable with this codebase's error shapes (every thrown value here is a plain extensible `Error`/`withCode` result), so this is defensive only. But since the whole point is "cleanup handling must never mask the original," wrapping the attach in its own `try {}` (or an `Object.isExtensible(err)` check) makes the guarantee airtight for any exotic thrown value — a fitting final nail for a class flagged three times.

2. *Informational:* the "200-line cap" citation still appears in five unrelated files (`exec-js-normalize`, `exec-stdlib-instance`, `exec-stdlib-slot`, `exec-stdlib.ts`). The PR correctly removed only the two in its scope; a repo-wide sweep of that recommended-not-enforced convention is a separate cleanup, not this PR's job.

Verdict below.

PR #26, opened 2026-07-31
feat: absorption phase-03 — FigJam

### LINKED ISSUE #3: Absorption phase-03 — FigJam (sticky/connector/table/code-block/board read)
Spec: plans/260731-1054-capability-absorption/phase-03. STEP 1 = manifest editorType + boot-path change ('find out what breaks'), mandatory design-file regression check. BLOCKED-not-workaroundable if dynamic-page documentAccess is incompatible with figjam (spec concern #1).

jangtrinh:
Closes #3.

## Summary
The manifest change IS the phase — `plugin/manifest.json`'s `editorType` widens to `["figma", "figjam"]`. Step 1 (before any helper code) was a boot-path trace to determine whether the existing design-editor path breaks in FigJam, reported to team-lead as a checkpoint before proceeding.

**Boot-path trace finding (nothing needs a skip):** `figma.showUI`/`figma.loadAllPagesAsync` are editor-agnostic per the typings — no editorType note on either, nor on `documentAccess: "dynamic-page"` anywhere the typings mention it (~40 places, all scoped to PageNode-loading/deprecated-getter behavior). The live-sync capture path (`main.ts`'s `onDocumentChange` + gap-fill) is **already honest-by-construction for FigJam nodes**: the component-scoped branch (`resolveComponentIdentity`) walks up for `COMPONENT`/`COMPONENT_SET` types that never exist in FigJam, so it's naturally inert there (never posts, never fabricates); the widened `EDIT_FEED` branch and gap-fill's snapshot/diff both stamp `nodeType` **verbatim** off the real node — a sticky reports `'STICKY'`, never a guessed design-file type. `STATUS.bootSkipped` (additive, present-only-when-non-empty, same contract as `senderMismatchCount`/`legacyMigrationDeferred` — issue #15/#19) reports empty today; the field exists for a future editor surface.

**Reverse-guard audit** (which EXISTING design-only helpers need a FigJam refusal — the bar: raw-throw or fabricate, not "already refuses cleanly"):
- `componentSet`/`slot`/`setProps` — **already safe**. `figma.createComponent()`/`combineAsVariants()` ARE documented Figma-Design-only, but every call site checks `node.type !== 'COMPONENT'` (or `'INSTANCE'`) before reaching them — since no FigJam node is ever that type, the check always fires first with an already-clear message. No new guard.
- `swapInstance` — **the one real gap, fixed generally**. Had NO type check on `inst` at all — any wrong-type node hit a raw, uncoded `in swapComponent: node is not an instance`. Pre-dates FigJam (reachable today in a design file with e.g. a FRAME); fixed with the same type check `setProps` already had, not an editor guard (a guard would leave the same hole open in Figma itself).
- `annotate`/`vars` — already safe or unconfirmed either way; documented as `[re-verify]` in `knowledge/figjam.md`, never asserted.

**New:** `shared/editor-surface.ts`'s `requireEditor` wrapper (`exec-stdlib-editor.ts` — built in phase-02, first wired here), and `ui.figjam.*`: `sticky`/`stickies` (batch, one font load reused for the whole batch — a real 200x saving)/`connector`/`shape`/`section`/`table`/`codeBlock`/`arrange`/`board`/`connections`. Every helper's first line is `requireEditor(...)` — no hand-written refusal string anywhere.

**Honest-reporting deviations from the fork** (all four, per the phase's own Definition of Done): `arrange` reports `skipped` ids instead of throwing only when ALL fail; `board`'s `truncated` is computed from whether the source list was exhausted, not `results.length >= maxNodes` (wrong on a board with exactly `maxNodes` matching nodes); `connections` never drops an unresolved endpoint (labels it `unresolved: true` instead).

## Test plan
- [x] Test-first for the swapInstance fix: `tests/exec-stdlib-instance-swap.test.ts`'s 3 cases confirmed to fail (raw uncoded error, no `code`) against the pre-fix code before the type check landed.
- [x] 68 new tests across `exec-stdlib-editor`, `exec-stdlib-figjam-{content,table,arrange,read}`, `exec-stdlib-instance-swap`, `executor-ops-status` — editor refusal on every helper, the font-fallback path (connector/shape/codeBlock), the batch partial-failure shape (`stickies`), `arrange`'s grid math for n=1,2,5,9 and each layout, the corrected `truncated` computation (exact-maxNodes case), and unresolved-edge labelling.
- [x] `mock-figma.ts` extended (shared fixture, not a parallel mock): FigJam node factories, a predicate-based `findAll`, a settable `loadFontAsync` failure, and `x`/`y` as real default-0 fields (closing a gap `arrange`'s own resolution check surfaced — `width`/`height` already defaulted, `x`/`y` didn't).
- [x] `npm run build` — clean (cli 238.5kb, plugin 176.1kb)
- [x] `npm test` — 95 files / 1198 tests green (was 94/1187 before this branch — the delta includes 3 tests brought in by a main-branch merge, plus this PR's own 68, minus/plus normal churn — no unexplained drift)
- [x] No `npm run lint` script exists in this repo — consistent with prior gates, skipped.

## NOT done — flagging, not faking
Per this track's standing pattern (every prior absorption-phase PR), the following remain **owner-manual live-canvas verification**, consolidated here since none of my sessions have live Figma Desktop access:

**FigJam board (this phase):**
1. Plugin opens without console errors; `figma-agent status` reports `editorType: "figjam"` and any `bootSkipped` entries (expected empty).
2. `sticky`, then `stickies` with one deliberately invalid spec → 4/1 split, 4 stickies visible on the board.
3. Two shapes + `connector` with a label → visible connector.
4. `table(3,3,{data})` populated; `table(2,2,{data: 3×3})` → `dataRowsIgnored` reported.
5. `codeBlock` (TypeScript); `section`; `arrange` over created ids, each layout.
6. `board()` shows the created nodes; `connections()` shows the step-3 edge with its label.
7. A design-only helper (e.g. `ui.slot.create`) → refusal naming FigJam vs Figma.
8. `exec-js --undo-group` creating 3 stickies then throwing → board clean, `rolledBack: true`.

**Figma design file (regression, this phase):**
9. `status`, `scan-design-system`, `html-to-figma`, one `exec-js`, `ui.slot.list` — all behave exactly as before the manifest change.
10. A FigJam helper (e.g. `ui.figjam.sticky`) → refusal naming Figma vs FigJam.

**Carried forward from earlier phases (still pending, same blocker):**
11. Phase-01 (`ui.vars.*`, `ui.componentSet`) live build + `--undo-group` rollback.
12. Phase-02 (`ui.slot.*`, `ui.annotate.*`) live create/list/append/reset cycle + editor-guard smoke.

`documentAccess: "dynamic-page"` + FigJam compatibility itself is **strongly evidenced, not live-confirmed** — the fork's own shipped manifest runs this exact combination in production, and the typings never document a conflict — but step 1 above is what actually proves it on a real board.

Holding for review.


jangtrinh (commented):
**Review verdict: APPROVE** — the two headline risks are clean: the manifest widening is additive with the design-file boot statically untouched (main.ts has zero editorType branches), and the guard direction is correct both ways (every ui.figjam.* refuses outside FigJam; NO design helper calls requireEditor, so none wrongly refuses in a design file). All three fork-deviations report honestly, and the fork's own truncated boundary bug is genuinely fixed — verified adversarially (injecting >= fails the exactly-maxNodes test). swapInstance's general type-check fires in a design file. Full suite 95/1198 matches gates. One non-blocking perf note (connections is uncapped and double-fetches each endpoint); the design-file/FigJam live smoke remains owner-manual, as flagged. (Posted as a review comment: GitHub refuses a formal approval state on a self-authored PR.)

Reviewed `0aea3be` on a fresh clone: 95 files / 1198 tests green (matching the reported gates), all seven risk areas check out, and the one fix I could attack adversarially — the fork's own `truncated` boundary bug — genuinely discriminates. The manifest widening and the guard direction are both correct. Approving; the two notes below are non-blocking.

## The two headline risks are clean

**Manifest widening is additive and the design-file boot is statically untouched.** `editorType: ["figma","figjam"]` adds FigJam without removing Figma. The only module-load change in `main.ts` is `const bootSkipped: string[] = []` — a plain empty array; I grepped `main.ts` for `editorType` and there is **zero** reference, so nothing at boot branches on the editor type and the design-file sequence (showUI, documentchange subscription, gap-fill) runs byte-identically. The design-file regression is genuinely owner-manual (no live canvas in CI), but there is no static path by which this PR changes it.

**Guard direction is correct on both sides.** All 11 `requireEditor` call sites are `ui.figjam.*` helpers requiring `['figjam']`, so each refuses actionably in a design file (`editorRefusal` returns a four-part message; `found=null` never satisfies). Critically, **no design-only helper calls `requireEditor` at all** — I grepped every caller — so nothing was pointed the wrong way to wrongly refuse in a design file. `editorRefusal`'s null-safety (`found !== null && required.includes(found)`) means an unknown host refuses rather than slips through.

## Focus verification

- **arrange (partial-failure):** unresolved/non-positionable ids go to `skipped[]` and `arranged` is the real resolved count — the fork throws only when ALL fail, losing the partial truth; this reports it.
- **board (the fork's own boundary bug):** `truncated = matching.length > maxNodes` computed from the SOURCE list, not `results.length >= maxNodes`. I injected the fork's `>=` form and the "EXACTLY maxNodes → truncated false" test fails; restored, it passes. The bug is genuinely fixed, not copied.
- **connections (unresolved endpoints):** an endpoint whose node can't resolve is kept with `unresolved: true`, never dropped — a dropped edge would silently alter the graph. Position-only endpoints report `position`, not a fabricated node.
- **swapInstance (general fix):** `if (inst.type !== 'INSTANCE')` is a runtime type-check, NOT editor-gated — the test passes a FRAME under the plain (design-file) mock and gets the clean `E_EVAL`, so a wrong-type node is refused in Figma itself, not just FigJam. Correct scoping of the "general correctness gap".
- **mock x/y fix:** `x`/`y` are now real own-fields from construction (beside the existing `width`/`height`), so `'x' in node` is true on a fresh node — which `arrange`'s resolution check depends on. No prior x/y accessor was shadowed (they're new fields), and the full suite stays green, so no existing test's meaning changed. The new FigJam factories encode a real refusal path (`setMockFontFailure` makes `loadFontAsync` throw, exercising the Inter-Medium fallback), not only happy paths.
- **bootSkipped envelope:** `...(bootSkipped.length > 0 && { bootSkipped: [...bootSkipped] })` — present-only-when-non-empty, empty today, spread of a copy, caller-supplied (never guessed). Same contract as `senderMismatchCount`/`legacyMigrationDeferred`; correct mechanism for phase-04 with no fabrication.
- **[re-verify] honesty:** all three items (dynamic-page/FigJam compat, annotate/vars availability, sticky-colour API) are worded unverified under "Open questions — `[re-verify]` on a live canvas" in `knowledge/figjam.md`, never asserted as confirmed.

## Non-blocking

1. **`connections` has no cap and re-fetches each endpoint.** Unlike `board` (`maxNodes`) and `arrange` (`MAX_ARRANGE_NODES`), `connections` walks every connector with no bound, and `resolveEndpoint` already calls `getNodeByIdAsync(ep.endpointNodeId)` to set `unresolved`, after which the `connectedNodes` loop calls `getNodeByIdAsync(ep.nodeId)` **again** for the same id — so each connected endpoint is fetched twice. On a large board that is 2–3 async lookups per connector with no ceiling. Threading the already-fetched node out of `resolveEndpoint` (and adding a cap mirroring `board`'s) would remove both the redundant round-trips and the unbounded walk. Correctness is fine; this is throughput on a big board.

2. *Owner-manual, restated not blocking:* the design-file and FigJam live-canvas runs remain the open verification — the manifest widening's real effect (FigJam actually loading the plugin without the design-only boot path erroring) is the one thing CI structurally cannot prove. The static read is clean; the live smoke is still owed, and the PR flags it.

Verdict below.

PR #28, opened 2026-07-31
feat: absorption phase-04 — Figma Slides

### LINKED ISSUE #4: Absorption phase-04 — Figma Slides
Spec: plans/260731-1054-capability-absorption/phase-04. Check whether their 1920x1080 background rectangle hack is superseded by a real API before porting (spec concern #5).

jangtrinh:
Closes #4. Last phase of the capability-absorption track (baskets A+B complete after this).

## Re-anchor findings (reported to team-lead before helper code)

- **dynamic-page + slides compatibility**: same evidence class as FigJam — the fork's own shipped manifest declares `editorType: ["figma","figjam","slides","dev"]` together with `documentAccess: "dynamic-page"`. Strong evidence, not live-confirmed; NOT BLOCKED.
- **Background API (the phase's own highest-value check)**: confirmed directly against developers.figma.com's `SlideNode` docs — `SlideNode` carries a real `fills`/`setFillsAsync` via `GeometryMixin`. The fork's 1920×1080 background-rectangle workaround is obsolete; `ui.slides.background` sets the fill directly, no rectangle at all (`method: 'slide-fill'`).
- **Boot-path trace**: nothing needs a Slides-specific skip — gap-fill/live capture already honest-by-construction for Slides' real node hierarchy (`SLIDE_GRID`→`SLIDE_ROW`→`SLIDE`). One genuinely new case (structural containers with no x/y) is already handled by the existing `'x' in node` guard.
- **Reverse-guard audit**: generalizes cleanly from FigJam — componentSet/slot/setProps/swapInstance already safe (no SLIDE node is ever COMPONENT/INSTANCE type); annotate/vars stay `[re-verify]`, checked fresh for Slides (not assumed identical to FigJam).

## What shipped

- `plugin/manifest.json`: `editorType` widened to `["figma","figjam","slides"]`.
- `plugin/src/main/exec-stdlib-slides-{types,resolve,crud,view,content}.ts` + `exec-stdlib-slides.ts` factory (all under the 200-line cap: 63/15/131/84/134/35 lines). 16 methods: `list`, `grid`, `content`, `create`, `remove`, `duplicate`, `reorder`, `setTransition`, `transition`, `viewMode`, `focused`, `focus`, `skip`, `background`, `addText`, `addShape`.
- Every helper guards via `requireEditor` (phase-02's editor-surface guard, now wired for its third editor).
- `reorder` refuses a grid that drops or duplicates a slide (the fork's own `setSlideGrid` silently accepts a grid missing slides — this repo's own addition).
- `addText`/`addShape` explicitly re-parent to the target slide and verify the result — first-class fix for the single-slide-view default-parenting trap (Figma's own create methods append to the FOCUSED slide by default in single-slide view, not wherever the caller intends).
- `knowledge/slides.md`: full writeup, including the "navigation is not undoable" caveat (`viewMode`/`focus` are queued as mutating but their view-change side effect isn't covered by `--undo-group`), the reverse-guard-audit table, and `[re-verify]` items worded unverified.
- Mock fix (found via my own test): `FakeNode.appendChild` never detached a node from its previous parent — a real gap, now fixed generally (a node can only have one parent in real Figma), not special-cased for Slides.
- `THIRD-PARTY.md`: new derivation row citing the fork's Slides handlers + the deliberate rewrites/fixes (background, reorder, content, duplicate).

## Gates

- `npm run typecheck` — pass
- `npm run build` — pass
- `npm test` — 96 files / 1223 tests, all green (was 95/1198 before this branch)

## Owner-manual checklist (live canvas)

- [ ] Plugin opens without console errors in a real Slides deck; `status` reports `editorType: "slides"`.
- [ ] `list()`/`grid()` on a deck with ≥3 slides across ≥2 rows — cross-check row/col against the Figma UI.
- [ ] `create()`/`duplicate()`/`remove()` verified in the UI.
- [ ] `reorder()` swapping two slides → read-back matches; a grid with one id omitted → refusal, deck unchanged.
- [ ] `setTransition` with a valid style → `transition()` read-back matches; an invalid style → refusal naming the valid set.
- [ ] `viewMode('grid')`, `focus(id)` → confirm the view actually changed and the reply reports `viewMode: 'single-slide'`.
- [ ] `background(id, '#hex')` on a deck whose slide size is NOT 1920×1080, confirming the real `setFillsAsync` path (no `Background` rectangle appears in the layers panel).
- [ ] `addText` with a non-Regular weight and `addShape`; confirm both land on the intended slide even after focusing a different one first (single-slide-view trap).
- [ ] `content(id)` → our serializer's shape, not a 7-field stub.
- [ ] `exec-js --undo-group` on a script that adds text, navigates, then throws → text is gone, navigation is NOT undone (the phase's honesty test).
- [ ] Regression: a Slides helper refuses in a FigJam board and in a design file; existing FigJam/design-file capabilities (`status`, `scan-design-system`, `ui.figjam.*`, `ui.slot.list`) behave unchanged.

## Test plan

- [x] Unit tests cover editor refusal (all 16 helpers), grid-integrity refusals (missing/duplicated/dropped ids), SLIDE type assertion, transition-enum validation, focus's view-mode-first ordering, background's real-API behavior, and the default-parenting trap.
- [ ] Live canvas checklist above (owner-manual).


jangtrinh (commented):
**Review verdict: APPROVE** — completes the absorption track (baskets A+B). Manifest widening additive with boot statically untouched (main.ts has zero editorType branches); guard direction correct every way (all ui.slides.* require ['slides'], no design/figjam helper wrongly refuses). The load-bearing addText/addShape parent-verify is real: the mock reproduces the single-slide focused-slide default-parenting trap and the test discriminates (stripping appendAndVerify drops the node on the wrong slide). reorder names missing/duplicated/dropped ids; createSlide positional is tsc-locked; enums match the API+fork cross-check and exclude the EASE_IN_AND_OUT_BACK artifact; the mock's appendChild-detaches fix is general and left the suite green (96/1223). Two non-blocking doc notes (background's ambiguous  field; INTERACTIVE_SLIDE_ELEMENT fact undocumented) + the owner-manual Slides live smoke. (Posted as a review comment: GitHub refuses a formal approval state on a self-authored PR.)

Reviewed `a6edbd1` on a fresh clone: 96 files / 1223 tests green (matching the gates), all ten focus points check out, and the load-bearing one — addText/addShape parent-verify — genuinely discriminates. Same clean shape as #26. Approving; this completes the absorption track's baskets A+B. Two minor doc notes, non-blocking.

## The correctness heart holds (point 2)

`appendAndVerify(slide, node, capability)` does `slide.appendChild(node)` then throws `E_EVAL` unless `node.parent === slide`; both `addText` and `addShape` call it. What makes this trustworthy is that the MOCK now reproduces the real trap: `createDefaultParented` appends a fresh `createText`/`createRectangle`/`createEllipse` to `focusedSlide` — but ONLY under `editorType:'slides'` + `slidesView:'single-slide'` + a set `focusedSlide`, exactly Figma's documented behavior. The test builds `target` and `focusedElsewhere`, focuses the latter, calls `addText(target.id, …)`, and asserts the node lands in `target.children` AND `focusedElsewhere.children` is empty. I stripped the `appendAndVerify` call so the helper relies on default parenting, and that exact test fails — the node lands on the focused slide, not the target. This is a real, discriminating guard against the silent-misattribution class, not a happy-path assertion.

## The two headline risks are clean (points 4, 5)

**Manifest `["figma","figjam","slides"]` is additive; design-file + FigJam boot statically untouched.** The `main.ts` diff is a comment-only change to the `bootSkipped` rationale — I grepped and there is still zero `editorType` branch at boot, so both prior editors' sequences run byte-identically.

**Guard direction is correct in every direction.** All six `ui.slides.*` view helpers and six crud helpers require `['slides']`; the figjam helpers all require `['figjam']`; no design-only helper calls `requireEditor` at all. A copy-paste of the wrong editor token would show as a `['figjam']` on a slides helper — none exists. So a slides helper refuses in both Figma and FigJam, a figjam helper refuses in Slides, and design helpers never wrongly refuse.

## Focus verification (points 1, 3, 6, 7, 8, 9, 10)

- **background:** `slide.setFillsAsync([...])` directly on the SlideNode — no rectangle, `method:'slide-fill'`; awaiting the async setter is the success signal (it rejects on failure). See note 1 on the `updated` field name.
- **reorder:** computes `missing` (unknown ids), `duplicated`, AND `dropped` (current ids absent from input), throws `E_INVALID_ARGS` naming all three, and reads the result back off `getSlideGrid()` — the repo's addition over the fork's silent-accept, honoring the mutating-reorder law.
- **createSlide positional:** `figma.createSlide(opts.row, opts.col)`; the mock and test both type it `createSlide(row?, col?)` and call it positionally, so `tsc` locks against the fork's `{row,col}` shape.
- **enums:** 23 transition styles / 8 curves; the code excludes the `EASE_IN_AND_OUT_BACK` summary artifact from both lists, matching the documented API + fork-zod cross-check. (I can't re-fetch the Figma docs offline, but the code matches the stated cross-check and omits the artifact — which is exactly what was asked.)
- **mock appendChild detach:** the detach (`if (child.parent) …filter`) lives in the base `FakeNode.appendChild`, general, not Slides-special-cased; `createDefaultParented` is scoped so every non-Slides `createText`/`createRectangle` is unaffected, and the full suite stays green — no existing test's meaning changed.
- **INTERACTIVE_SLIDE_ELEMENT:** correctly never created or mutated — `addShape` is restricted to RECTANGLE/ELLIPSE, `addText` to TEXT, and `content()` runs it through the shared serializer (verbatim `node.type`, read-only). See note 2.
- **[re-verify] honesty:** the row/col-collision, distinct-`duplicate()`, annotate/vars-availability, and cap items are all worded unverified under "Open questions — `[re-verify]` on a live canvas" in `knowledge/slides.md`.

## Non-blocking

1. **`background`'s `updated` field is ambiguous.** It returns `updated: hadFill` — i.e. whether a fill existed BEFORE (replace) vs not (add). A caller could read `updated:false` as "nothing happened" when the fill was in fact applied to a previously-fill-less slide. The apply itself is honest (the awaited `setFillsAsync` is the success signal), but a clearer name (`replacedExistingFill`) or a one-line doc would remove the ambiguity.

2. **The "INTERACTIVE_SLIDE_ELEMENT is uncreatable / position-only" fact isn't documented.** The code correctly never touches it, but `knowledge/slides.md` never mentions it — so an agent doesn't learn why `addShape`'s type set is narrow or that this node type is read-only. A one-line entry (the fork carried this fact) would close the recipe's completeness, matching the "absorbed fact → recipe prose" discipline the rest of the track follows.

3. *Owner-manual, restated:* the Slides live-canvas smoke (the plugin actually loading in a Slides deck without the design-only boot erroring) is the one thing CI structurally can't prove — the static read is clean and the boot provably doesn't branch on editorType, but the live run is still owed, as the PR flags.

Verdict below.

jangtrinh:
Both addenda pushed (8425372):
1. background()'s updated field renamed to hadPriorFill -- reads correctly now (prior-state fact, not a change-happened claim). Test names updated to match.
2. knowledge/slides.md documents INTERACTIVE_SLIDE_ELEMENT -- real, Plugin-API-uncreatable, position-only, why addShape stays RECTANGLE/ELLIPSE-only.

Gates re-run: typecheck, build, and the full vitest suite all green (96 files/1223 tests). Ready to merge per your ruling.

PR #31, opened 2026-08-01
feat: shared safeCleanup() helper, close cleanup-masking class

### LINKED ISSUE #24: Shared safeCleanup() helper — close the cleanup-masking class airtight, once
The "a throwing cleanup() masks the original error" bug has now been found and fixed FOUR times independently (PR #12 broker cleanup, PR #13 addProperty, PR #19-adjacent, PR #23 componentSet). Each fix is correct but local — the next cleanup-on-error site will reintroduce it. Close the CLASS: add one shared `safeCleanup(originalError, cleanupFn)` helper that runs cleanup in its own try, attaches any cleanup failure as `.cleanupError` WITHOUT substituting, and is airtight against the edge PR #23's reviewer flagged (a frozen/non-extensible `err` — guard the attachment with Object.isExtensible or its own try so the un-masking guarantee never itself throws). Then route the 4 existing sites through it. One helper, one test, the class cannot resurface on a fifth site.

jangtrinh:
## Summary
- Closes #24: the "a throwing cleanup() masks the original error" class has been found and independently fixed four times (PR #12, #13, #19-adjacent, #23). This adds one shared `safeCleanup(originalError, cleanupFn)` helper in `shared/safe-cleanup.ts` so the guarantee lives in one place.
- Contract: run `cleanupFn()` in its own try; if it throws, attach the failure to `originalError` as `.cleanupError` without substituting, then rethrow the original error unchanged (code + message byte-preserved).
- Airtight against the frozen/non-extensible `originalError` edge flagged by PR #23's reviewer: the `.cleanupError` attachment is guarded by `Object.isExtensible` plus its own try, so the un-masking guarantee can never itself throw and re-mask. When attachment isn't possible, the cleanup failure is logged instead of dropped.

## Sites routed through the helper
- `plugin/src/main/exec-stdlib-component-set.ts` — `componentSet`'s cleanup (was PR #23's local fix; contract matched exactly).
- `cli/src/transport/broker-discovery.ts` — `writeFileAtomic`'s temp-file cleanup (PR #12's site; same try-cleanup-rethrow shape).

## Site NOT routed (with reason)
- `plugin/src/main/exec-stdlib-slot-property.ts`'s `addProperty` rollback (PR #13) is intentionally left alone. Its locked behavior — asserted by `tests/exec-stdlib-slot.test.ts` ("names the leftover property in the error when the rollback deletion itself fails") — throws a **new** `Error` whose message concatenates the original failure and the rollback failure (`"...not linked to...additionally...cleanup...deletion refused"`). That is a different contract from `safeCleanup`'s byte-preserved rethrow-with-attached-property, and forcing it through the helper would either break that test or require bolting extra message-building logic onto the call site, defeating the point of centralizing. Flagged rather than forced.
- Could not find an independent "PR #19-adjacent" cleanup-masking site in the current tree — the nearby fix from that window (`1bca6ca`, legacy-migration-defers-on-failure in `broker-daemon.ts`) is a different pattern (swallows and never rethrows, doesn't mask an original error), so it isn't a `safeCleanup` candidate.

## Tests
- New `tests/safe-cleanup.test.ts`: throwing cleanup surfaces the original error unchanged with `.cleanupError` attached; non-throwing cleanup is fully transparent; a custom error subclass's `code`/`message` survive byte-for-byte; the airtight frozen-`originalError` case (verified to fail against a naive assignment-without-guard implementation, restored after confirming); a frozen `originalError` with a non-throwing cleanup stays untouched.
- All existing tests for the two routed sites remain green, unmodified.

## Test plan
- [x] `npm run typecheck`
- [x] `npm run build`
- [x] `npm test` (97 files / 1228 tests, includes the panel gate against the pinned `kernel/design-os` submodule)

### DIFF
--- a/cli/src/transport/broker-discovery.ts
+++ b/cli/src/transport/broker-discovery.ts
@@ -16,6 +16,7 @@ import {
   type BrokerAdvertisement,
 } from '../../../shared/protocol.ts';
 import { CliError, envMs } from './protocol-helpers.ts';
+import { safeCleanup } from '../../../shared/safe-cleanup.ts';
 
 // Mirrors broker-daemon.ts's own `HEARTBEAT_MS` override (same env var, same
 // fallback) — see envMs's doc for why this must be the ONE shared reader, not a
@@ -105,8 +106,10 @@ function writeFileAtomic(path: string, contents: string): void {
     writeFileSync(tmpPath, contents);
     renameSync(tmpPath, path);
   } catch (err) {
-    try { unlinkSync(tmpPath); } catch { /* best-effort */ }
-    throw err;
+    // Best-effort temp-file removal on failure (issue #24: routed through the
+    // shared `safeCleanup` so the original write/rename error always propagates,
+    // never masked by an unlink failure).
+    safeCleanup(err, () => unlinkSync(tmpPath));
   }
 }
 
--- a/plugin/code.js
+++ b/plugin/code.js
@@ -2813,6 +2813,28 @@
     };
   }
 
+  // shared/safe-cleanup.ts
+  function safeCleanup(originalError, cleanupFn) {
+    try {
+      cleanupFn();
+    } catch (cleanupError) {
+      attachCleanupError(originalError, cleanupError);
+    }
+    throw originalError;
+  }
+  function attachCleanupError(originalError, cleanupError) {
+    const canAttach = originalError !== null && typeof originalError === "object" && Object.isExtensible(originalError);
+    if (!canAttach) {
+      console.error("safeCleanup: cleanup failed and originalError is not extensible, logging instead of attaching:", cleanupError);
+      return;
+    }
+    try {
+      originalError.cleanupError = cleanupError;
+    } catch (attachError) {
+      console.error("safeCleanup: failed to attach cleanupError despite isExtensible check:", cleanupError, attachError);
+    }
+  }
+
   // plugin/src/main/exec-stdlib-component-set.ts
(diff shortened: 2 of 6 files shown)


jangtrinh:
**APPROVE** — stage-4 review (orchestrator re-ran gates in the branch worktree: typecheck ✅, build ✅ (code.js 188.6kb), 30 focused tests ✅ safe-cleanup + component-set; broker-daemon integration suite deferred — a live plugin session is connected and the daemon harness would pollute broker discovery per issue #32, and this change touches no broker runtime path).

The helper is correct against all three criteria the class demands: composes (attaches `.cleanupError`) rather than substitutes, rethrows the original error object identity with code+message byte-preserved, and guards the non-extensible/frozen-error edge with `Object.isExtensible` + an inner try that logs (never re-masks). Tests cover the frozen-error airtight case that motivated the guard, plus the custom-subclass code preservation (E_EVAL survives). Both bare-masking call sites (`writeFileAtomic`, `componentSet`) are routed through it.

One non-blocking note (filed as a follow-up, not a merge blocker): the helper's doc lists `exec-stdlib-slot-property.ts` among the confirmed sites, but that site is intentionally NOT routed and should not be silently grouped with the masking sites — it already preserves `code` via `withCode(..., original.code)` and deliberately combines the cleanup detail into the thrown message ("fallback honesty"), so it is not a member of the masking class. Routing it through `safeCleanup` would be a behavior change (cleanup detail moves from `.message` to `.cleanupError`). Decide explicitly in the follow-up; for now the doc slightly overclaims 'one place'.

Issue #35, opened 2026-08-01
Multi-editor targeting: --instance <id> flag + panel "Target this plugin" button (fileName is not a unique key)


jangtrinh:
## Problem
With the capability-absorption track, a user routinely has 3 plugins connected at once (a Figma **Design** file, a **FigJam** board, a **Slides** deck). Routing today keys on **file name** (`--file`, exact/case-insensitive; precedence `--file` > `FIGMA_AGENT_FILE` > active). Two **unsaved** files both report `figma.root.name === "Untitled"`, so the name is not a unique key and the agent cannot target one over the other.

## Verified behavior (live, this session)
- **Agent CAN self-target a distinctly-named file, no focus/button needed:** `exec-js - --file "VSF - PCP"` hit the Design file (`editor: figma`) while the Slides deck was the *active* plugin. Targeting works agent-driven today — when names are distinct.
- **The collision is handled honestly, not silently:** `--file "Untitled"` returns a clean refusal that already surfaces the unique key:
  > `--file "Untitled" matches 2 connected files [Untitled#inst_…a209756n, Untitled#inst_…gkc2n3h0] — close one panel, or rename the files apart`
  Nothing vanishes silently; the instanceId is right there. What's missing is a way to *act* on it.

## Decision (owner, this session)
Add a stable per-instance target: **`--instance <id>`** so the agent targets the exact instance the collision error already prints — **without forcing a rename** — **plus** a panel **"Target this plugin"** button for one-click pin.

## Scope
1. **CLI `--instance <instanceId>`** (global flag, same layer as `--file`). Precedence: `--instance` (exact instanceId) > `--file` (exact name) > `FIGMA_AGENT_FILE` > active. Optionally accept the `Name#instanceId` composite the error already emits as a `--file` value, routed to the instanceId.
2. **Panel "Target this plugin" button** — pins the daemon route to THIS instanceId (an instanceId analogue of the `FIGMA_AGENT_FILE` pin), so subsequent no-flag commands hit it. Show the current pin in the panel; make it clearable. Pin is per-daemon, cleared on that instance's disconnect (never silently re-points to another instance — surface "target disconnected" instead).
3. Update the `E_NO_PLUGIN` / collision message to name `--instance <id>` as the precise fix, alongside the existing rename/close nudge.

## Acceptance
- With two "Untitled" files open, `exec-js - --instance <figjam-id>` runs on exactly that board; the other is untouched.
[...]

Issue #32, opened 2026-08-01
Test isolation: broker-daemon-harness writes to the REAL /tmp broker log + a fake plugin registers, interfering with a live plugin during dev


jangtrinh:
Observed 2026-08-01: while the owner was doing a live-canvas smoke, an executor running the full `npm test` in a worktree caused the plugin's broker discovery to fail ("Looking for the broker") — the harness test broker wrote `__TEST_BROKER_EXIT__` + a fake plugin registration (`instanceId: plugin-sync-live-1`, `fileName: 'Project C File'`) to the REAL `/tmp/figma-agent-broker.log`, and a real plugin's scan can stumble onto a test broker. The issue-5 fix DI'd the advertise PATH and ports for the harness, but the LOG path (and possibly a real-range port in some test) is still the shared `/tmp` one. Fix: thread a DI'd log path (and assert no test binds 9410-9419) so a full test run on a dev machine can never disturb a live plugin. Until then: process rule — don't run the plugin repo's full suite on the same machine while a live-canvas smoke is in progress.

jangtrinh:
Closed by #46 (merged → main). The last shared real resource (LOG_FILE) is now injectable via BrokerDaemonOptions.logFile; the harness passes a scratch path, so a test broker never appends to the live /tmp log (broker.json + port range were already isolated). The 4 previously-deferred harness describe blocks (--instance/force-release/contention/read-only) now RUN — 24/24 pass, independently re-verified.

PR #36, opened 2026-08-01
Add --instance <id> exact-target routing flag

jangtrinh:
## Why

With FigJam + Slides + Design all connectable at once, routing keys on **file name** (`--file`, exact). Two unsaved files both report `figma.root.name === "Untitled"`, so the name is not a unique key: `--file "Untitled"` refuses with a collision naming both instanceIds. `instanceId` IS unique (minted per iframe load, survives reconnect) and is already carried on every `PLUGIN_HELLO` and printed in the collision error — but nothing could route by it until now.

This is Phase 1 of a two-phase plan (issue #35): the CLI flag only. A panel "target this plugin" button is a separate follow-up.

## Precedence

| Rank | Filter | Scope | Match |
|---|---|---|---|
| 1 | `--instance <id>` | per request | exact, unique |
| 2 | `--file <name>` | per request | exact |
| 3 | `FIGMA_AGENT_FILE` | per daemon (env) | substring |
| 4 | active | ambient | most-recent |

`--instance` and `--file` on the same request are mutually exclusive — passing both refuses with `E_INVALID_ARGS` naming both flags, it never silently picks one.

## What changed

- `RouteFilter` gains a `kind: 'name' | 'instance'` discriminator so the registry knows whether `value` is a file name or an opaque instanceId.
- `PluginRegistry.matching`/`.selectTarget` branch on `kind:'instance'`, reusing the existing `getByInstanceId()` lookup — an instanceId match is always 0 or 1 entries, never a name-fuzzy list, even when two live entries share a fileName.
- The daemon's `E_NO_PLUGIN` message names the instanceId (not "open the file's panel") when an `--instance` target isn't connected, and the `--file` ambiguity message now also suggests `--instance <id>` as the exact fix.
- CLI: `--instance <id>` documented alongside `--file`, with the same empty-value guard and a boundary-level mutual-exclusion check.

Scope note: the granted file list didn't include `shared/protocol.ts` (the `RequestMsg.expectedInstance` wire field) or `cli/src/transport/broker-client.ts` (the CLI-side `setExpectedInstance` state, mirroring `setExpectedFile`) — both are required plumbing for the flag's value to cross the CLI→broker process boundary at all, so they're included here. `cli/src/transport/broker-status.ts`'s `noPluginMessage` also needed the instance-aware message branch. All four are additive/backward-compatible (new optional field, new optional param appended last).

## Test-first

Added failing-then-passing coverage (each confirmed red on the pre-change code, verified via the diff sequence in the PR history):
- `tests/route-filter.test.ts` — `--instance` beats `--file` beats env beats none; resolved filter carries `kind:'instance'`, `exact:true`.
- `tests/plugin-registry.test.ts` — `matching`/`selectTarget` with `kind:'instance'` return the single live entry for a known id, `[]` for an unknown/dead id, and never a name-fuzzy list even when two live entries share a fileName.
- `tests/broker-status.test.ts` — `noPluginMessage` names the instanceId and never lists connected files for an `--instance` miss.
- `tests/broker-daemon-harness.test.ts` — two new cases (instanceId-not-connected → `E_NO_PLUGIN` naming the id; `--instance` routes to the exact instance among two same-named live plugins). **Written but deliberately not run** — this harness spawns a real in-process broker, and a live plugin session connected on the executing machine made running it unsafe (broker-discovery pollution, tracked separately). Left for a follow-up run once no live session is active.

## Gate

- `npm run typecheck` — pass
- `npm run build` — pass
- `npx vitest run tests/route-filter.test.ts tests/plugin-registry.test.ts` — 49/49 pass
- Also re-ran the other pure suites touched by the `RouteFilter` shape change (`broker-status.test.ts`, `file-guard-plumbing.test.ts`, `activity-labels.test.ts`, `job-table-sender-verification.test.ts`) — 34/34 pass
- Full `npm test` and the broker-daemon-harness suite were intentionally not run (live-broker-safety constraint above)

### DIFF
--- a/cli/src/figma-agent.ts
+++ b/cli/src/figma-agent.ts
@@ -5,7 +5,7 @@ import { resolve } from 'node:path';
 import { runBrokerDaemon } from './transport/broker-daemon.ts';
 import { parseArgs, type CommandArgs } from './arg-parse.ts';
 import { CliError } from './transport/protocol-helpers.ts';
-import { getLastFileContext, setExpectedFile, setProjectDir, setReadOnly } from './transport/broker-client.ts';
+import { getLastFileContext, setExpectedFile, setExpectedInstance, setProjectDir, setReadOnly } from './transport/broker-client.ts';
 import { printErrorJson, printJson, withFileContext } from './util/json-out.ts';
 import * as batch from './commands/batch.ts';
 import * as changes from './commands/changes.ts';
@@ -120,6 +120,12 @@ Commands:
 Global: --file "<exact file name>"   route to that file's plugin AND refuse to run anywhere else
                                      (exact, case-insensitive; beats FIGMA_AGENT_FILE; payloads
                                       >512KB route by the env pin but are still guarded)
+        --instance <instanceId>     route to that EXACT plugin instance (from \`status\`'s
+                                     instanceId column) — beats --file and FIGMA_AGENT_FILE; the
+                                     one way to target a specific instance when --file's name is
+                                     ambiguous (e.g. two unsaved files both named "Untitled").
+                                     Mutually exclusive with --file: passing both refuses with
+                                     E_INVALID_ARGS naming both, it never silently picks one.
         --dir <projectDir>          this invocation's project root (default: cwd); stamped on
                                      every request so panel/idle sync can apply into the right
                                      project once bound (\`figma-agent bind\`) — never a guess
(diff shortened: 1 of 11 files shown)


jangtrinh:
**APPROVE** — stage-4 review (orchestrator re-ran gates on the branch: typecheck ✅, build ✅, 64 unit tests ✅ across route-filter + plugin-registry + broker-status).

Correctness verified against the load-bearing invariants:
- **Precedence holds**: `resolveRouteFilter` returns instance→file→env→none in that order; instance is always exact/unique.
- **Instance match is 0-or-1**: `matching({kind:'instance'})` routes through the existing `getByInstanceId` + a liveness check — never a name-fuzzy list, even with two same-named files. No new index.
- **No silent re-route** (the one that mattered): an instanceId that matches nothing PARKS on its own filter (kind:'instance') and re-admits ONLY when that exact id connects — it never falls back to another instance or to 'active'; on timeout it refuses with an instance-aware `E_NO_PLUGIN` that points at `status`, not at a file panel.
- **Mutual exclusion**: `--instance`+`--file` together refuse at the CLI boundary before either setter runs.
- **Judgment call is sound**: skipping the `E_PLUGIN_STALE` (`pluginSupportsFileGuard`) check for kind:'instance' is correct — that guard exists because *name* routing needs the plugin to re-check `figma.root.name`; instance routing is socket-exact (getByInstanceId → that WS), so there is nothing for the plugin to re-guard.
- **Backward-compat**: `expectedInstance?` is optional and omitted-when-unset, `makeRequestFrame` appends it last — a pre-flag frame serializes byte-identically. The scope expansion (protocol.ts / broker-client.ts / broker-status.ts) was necessary (the setter lives in broker-client, the wire field in protocol) and is additive.

One MINOR fixed in review (commit 583c771): 7 `spec 260801-1050` plan-id citations had leaked into code comments + test `describe` names — replaced with the invariant stated directly (repo rule: no plan ids in stable code artifacts). No behavior change; gates re-green after.

The 2 new `broker-daemon-harness.test.ts` cases (instance-miss → E_NO_PLUGIN names the id; `--instance` isolates one of two same-named live plugins) remain unrun here — that harness spawns a real broker and would pollute the live plugin session's discovery (issue #32). They will be validated first by the live `--instance` smoke that this PR unblocks, then run in full post-smoke.

Issue #38, opened 2026-08-01
--read-only is trusted, not enforced: a mis-declared read that mutates bypasses the per-file queue and interleaves (silent, scales with agent count)


jangtrinh:
Architecture audit finding (strongest-model review of the multi-agent concurrency model, HEAD 85550a6). **This is the single silent breach of the same-file-write invariant.**

## The hole
Same-file write safety rests on the broker's per-file mutation queue PLUS honest `--read-only` declarations — NOT on the plugin being single-threaded (the JS thread prevents *simultaneous* execution but not *interleaved* execution: two async dispatches overlap at `await` points; `plugin/src/main/main.ts:420,460-485` explicitly handles overlapping in-flight commands).

Reads skip the queue (`broker-daemon.ts:807-811`, gated by `readOnly===true` / `IMPLICIT_READ_ONLY`). `--read-only` is TRUSTED, not enforced — the help text admits it verbatim (`cli/src/figma-agent.ts:133-135,180`: *a mis-declared mutation can interleave with another agent's work*). With many agents authoring exec-js, a sloppy 'read' that actually touches the scene bypasses the queue and interleaves with a running mutation in the same instance → read-modify-write corruption. Silent; probability scales with agent count.

## Fix (infra already exists)
Enforce read-only plugin-side: carry the `readOnly` flag to the plugin; the mutation-tracking already there (`beginAgentMutation`/`declaredIds`/documentchange in `main.ts`) can detect if a declared-read-only dispatch recorded ANY agent mutation → fail it with a distinct error code, and count violations in `status`. Converts the silent breach into a loud refusal. Est ~half-day.

## Acceptance
- An exec-js run with `--read-only` that mutates the scene is REFUSED with a distinct code (not silently applied).
- The violation count is visible in `status`.
- Genuinely read-only scripts under `--read-only` still bypass the queue and run concurrently (no regression).

jangtrinh:
Closed by #40 (merged → main `2329634`). A `--read-only` EXEC_JS that mutates is now refused with `E_READONLY_VIOLATION` + counted in `status.readOnlyViolations`; attribution is per-dispatch via the `activeCount===1` signal (never false-positives a legitimate read). Auto-rollback deferred (leak sealed in its own undo step). 2 broker-harness tests written, deferred to post-live-smoke.

PR #40, opened 2026-08-01
Refuse a --read-only EXEC_JS that mutates the scene

jangtrinh:
## The silent breach

`--read-only` skips the per-file mutation queue so a read can run while a mutation for
the same file is in flight. Every typed mutating command already refuses `--read-only`
(`refusesReadOnlyAssertion`, `cli/src/transport/broker-client.ts`) — except `EXEC_JS`,
whose script the broker can't statically inspect. So a mis-declared `--read-only`
`EXEC_JS` was the one path where a "read" could still write: it skipped the queue and
could interleave with a running mutation in the same plugin instance, with nothing to
show for it afterward.

## Scope

EXEC_JS only — the only command this hole could ever apply to. Typed commands are
untouched.

## The attribution problem (and why a naive check is wrong)

The plugin's `onmessage` handler can have two dispatches in flight at once (interleaved
at `await` points). A check like "any `documentchange` event during this read →
violation" sounds right but isn't: it would blame a read-only read for a *separate*,
legitimate mutation that happens to be running concurrently on other nodes — turning
every overlapping read into a false refusal. That's worse than the silent gap it
replaces.

The six typed mutating commands avoid this by declaring their target node ids *before*
they run (`beginAgentMutation`/`declaredIds`), so a change on a declared id during that
window is unambiguously theirs. `EXEC_JS` can't do this — it's an opaque script that
might touch anything, so there's no id to pre-declare.

What every dispatch DOES have is `activeCount` — how many requests this plugin instance
currently has in flight. The fix: a `documentchange` batch is only ever counted as
attributable to *a* dispatch when `activeCount === 1` at the moment it lands. If exactly
one dispatch is active, that dispatch is provably the only thing that could have caused
it. A concurrent, separate mutation (`activeCount` at 2+) is excluded from consideration
entirely — its own documentchange never gets attributed to anyone else's read.

This is intentionally coarser than per-node attribution, and it accepts a real, narrow
residual gap: two undeclared `EXEC_JS` calls racing each other (one `--read-only`, one
genuinely mutating) can't be told apart by this signal — `activeCount` never drops to 1
during their overlap, so neither gets flagged for that window. That's a false *negative*,
never a false *positive*, and is documented in `plugin/src/main/readonly-guard.ts`.

## What changed

- `plugin/src/main/readonly-guard.ts` (new) — the pure attribution state machine above,
  no `figma` dependency (mirrors `edit-actor.ts`'s own reasoning for why: `main.ts` can't
  be imported outside a live plugin sandbox, so the part worth trusting has to be
  testable without one).
- `plugin/src/main/main.ts` — wires the guard into `onDocumentChange` and the request
  dispatch loop; a violating `--read-only` `EXEC_JS` now replies `E_READONLY_VIOLATION`
  instead of applying silently; `readOnlyViolations` counter added, surfaced via STATUS.
- `plugin/src/ui/ui-relay.ts` — **anchor-drift fix**: `RequestMsg.readOnly` was already on
  the wire protocol, but the UI relay's `handleRequest` never forwarded it into the
  `postMessage` sent to main — so main would never have seen the flag at all regardless
  of anything done in `main.ts`. Fixed alongside the rest of this change; noting it here
  since the original scope only named `main.ts`.
- `shared/protocol.ts` — new `E_READONLY_VIOLATION` error code.
- `plugin/src/main/executor-ops.ts` — `opStatus` takes an optional `readOnlyViolations`
  count, omitted from the payload when zero (same "present only once meaningful"
  contract as `bootSkipped`/the broker's `senderMismatchCount`).

## Test-first

Wrote `tests/readonly-guard.test.ts` and `tests/readonly-exec-js-guard.test.ts` before
implementing `readonly-guard.ts`'s real logic, then verified the load-bearing
concurrency test actually goes red against the naive, spec-forbidden design ("any
documentchange while busy → violation") before implementing the `activeCount === 1`
attribution — confirmed 2 failures on that naive version, both passing once the real
per-dispatch check replaced it.

- `tests/readonly-guard.test.ts` — pure attribution logic, including the load-bearing
  case: a `--read-only` read overlapping a separate declared mutation is never flagged;
  only a dispatch that mutates while it is the sole active one is.
- `tests/readonly-exec-js-guard.test.ts` — the same wiring driven against real
  `opExecJs` runs (a script that actually calls `figma.createFrame()` vs. one that only
  reads), rather than synthetic counters.
- `tests/executor-ops-status.test.ts` — `readOnlyViolations` presence/omission in STATUS.

Both `npm run typecheck` and `npm run build` pass. The full suite was intentionally not
run in this environment (a live plugin session was connected and a broker-spawning test
harness would have polluted its discovery) — the touched/new test files above were run
directly and pass.

## Rollback (open question, resolved as: defer)

A detected violation means the script already mutated. This PR is v1: detect, refuse,
count — never a silent apply. The leaked mutation is still sealed into its own undo step
via the existing commit-on-failure path (`EXEC_JS` is in `MUTATING_COMMANDS`), so a
designer can revert it with one manual undo. Automatic rollback beyond that didn't fall
out cleanly from the existing `--undo-group` sentinel machinery without adding real risk
(the sentinel bracket is opt-in per call, not universal), so it's deferred rather than
built here.

## Deferred

A broker-spawning integration test (plugin ↔ real broker, `--read-only` end to end) was
explicitly out of scope for this environment per the task constraints — left for the
orchestrator to run separately, not included in this PR.

### DIFF
--- a/plugin/code.js
+++ b/plugin/code.js
@@ -2341,7 +2341,7 @@
     if (typeof params.x === "number") node.x = params.x;
     if (typeof params.y === "number") node.y = params.y;
   }
-  function opStatus(bootSkipped2 = []) {
+  function opStatus(bootSkipped2 = [], readOnlyViolations2 = 0) {
     return {
       fileName: figma.root.name,
       page: figma.currentPage.name,
@@ -2365,7 +2365,13 @@
       // knowledge/figjam.md) — the field exists so a FUTURE editor (phase-04 Slides, or
       // a later FigJam finding) has somewhere to report one, without a payload shape
       // change.
-      ...bootSkipped2.length > 0 && { bootSkipped: [...bootSkipped2] }
+      ...bootSkipped2.length > 0 && { bootSkipped: [...bootSkipped2] },
+      // Concurrency & jobs (issue #38) — same "present only once meaningful" contract as
+      // bootSkipped just above and the broker's own senderMismatchCount: a fleet that has
+      // never seen a mis-declared `--read-only` EXEC_JS keeps this payload byte-identical
+      // to before the field existed; a non-zero count makes a real pattern visible instead
+      // of vanishing into a silently-applied mutation.
+      ...readOnlyViolations2 > 0 && { readOnlyViolations: readOnlyViolations2 }
     };
   }
   function opGetSelection(params) {
@@ -4596,6 +4602,23 @@
   var PANEL_WIDTH = 300;
   var PANEL_HEIGHT = 420;
 
+  // plugin/src/main/readonly-guard.ts
+  function createReadOnlyGuardState() {
+    return { soleActorChangeEvents: 0 };
+  }
+  function recordDocumentChangeBatch(state, activeCount2) {
+    if (activeCount2 === 1) state.soleActorChangeEvents += 1;
+  }
+  function snapshotChangeEvents(state) {
+    return state.soleActorChangeEvents;
+  }
+  function violatedSinceSnapshot(state, snapshot) {
+    return state.soleActorChangeEvents > snapshot;
+  }
+  function isReadOnlyExecJs(cmd, readOnly) {
+    return cmd === "EXEC_JS" && readOnly === true;
+  }
+
   // plugin/src/main/main.ts
   figma.showUI(__html__, {
     visible: true,
@@ -4687,6 +4710,8 @@
   function actorState() {
     return { activeCount, lastDrainAt, declared: declaredIds, lastAgentAt };
   }
+  var readOnlyGuard = createReadOnlyGuardState();
+  var readOnlyViolations = 0;
   var idleMs = DEFAULT_IDLE_MS;
   var idleTimer = null;
   var changesSinceCommit = 0;
@@ -4707,6 +4732,7 @@
   }
   function onDocumentChange(event) {
     pruneDeclaredIds(declaredIds, Date.now());
+    recordDocumentChangeBatch(readOnlyGuard, activeCount);
     const raw = [];
     const edits = [];
     for (const dc of event.documentChanges) {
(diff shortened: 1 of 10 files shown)


jangtrinh:
**APPROVE** — stage-4 review (orchestrator re-ran gates on the branch: typecheck ✅, build ✅, 19 tests ✅ across readonly-guard + readonly-exec-js-guard + executor-ops-status).

The attribution design is sound and conservative — the one property that had to hold: **never false-positive a legitimate read.** `recordDocumentChangeBatch` bumps the counter ONLY when `activeCount === 1` (a change reliably attributable to the sole in-flight dispatch); the guard snapshots before `dispatch()` and checks `violatedSinceSnapshot` after. Two dispatches overlapping → the change isn't counted → a false NEGATIVE, never a false positive. The residual gap (two undeclared exec-js racing) is documented in code and fails on the safe side. The load-bearing concurrency test was verified red against the naive 'any documentchange while busy' design before the real check went in.

Two things I verified beyond the happy path:
- The `ui-relay.ts` drift the executor caught is real and load-bearing — without carrying `req.readOnly` through the relay, the whole feature is dead on arrival; good that a test would have caught it.
- The violation fires AFTER `dispatch()` (the script already ran), so v1 is detect + refuse (`E_READONLY_VIOLATION`) + count (`status.readOnlyViolations`), with the leaked mutation sealed in its own undo step. Auto-rollback correctly deferred.

One MINOR fixed in review (commit 1166f48): 10 `(issue #38)` citations had leaked into code comments + a test `describe` — same repo-rule as the last PR's fix, stripped to state the invariant directly. Gates re-green after. (Filing a follow-up for a CI guard so this class stops recurring.)

The 2 broker-harness tests remain unrun here (live plugin session, issue #32) — flagged for post-smoke, same as PR #36.

Issue #41, opened 2026-08-01
CI guard: fail on plan-id / issue-number citations in code comments and test names


jangtrinh:
Recurring class — caught and fixed by hand in TWO consecutive PRs now (#36 `spec 260801-1050`, #40 `issue #38`). Repo rule (CLAUDE.md / stable-code-artifacts): code comments, test `describe`/`it` names, commit messages, and migration names must state the invariant directly, never cite a plan id, issue number, phase label, or finding code. A prose rule drifts; it needs a check that fails without it.

## Ask
Add a lint/CI grep (own script, wired into the gate) that FAILS when a tracked source/test file contains a citation like `issue #<n>`, `spec <6-digit-date>`, `phase <n>` in a comment or a test name. Scope to `plugin/src`, `cli/src`, `shared`, `tests`. Allow the citation in `plans/` (gitignored anyway) and in PR/issue bodies.

## Acceptance
- A file with `// ... (issue #42)` in a comment fails the check with a clear message pointing at the line.
- A test named `describe('x (issue #42)')` fails it.
- The current tree passes (both prior leaks are already stripped).
- Wired into `npm run` gate + CI so it can't merge with a leak.

Pattern to match at minimum: `issue #\d+`, `#\d+` adjacent to issue words, `spec \d{6}`, `\d{6}-\d{4}-` (plan-dir slugs), `phase \d` in comments. Tune to avoid false positives on legit content (e.g. a hex color, a real code identifier).

jangtrinh:
Closed by #44 (merged → main). scripts/check-comment-hygiene.mjs + npm run lint:comments now fail on NEW ephemeral citations (GH issue/PR numbers, dated plan-dir slugs, bare finding/(P#) labels) while allowing the repo's durable design-doc narrative (spec-NNN, wave, backlog, RI-P, phase). Content-keyed baseline grandfathers 109 pre-existing ephemeral refs; verified a new ephemeral ref fails and durable forms pass. Wired into the gate.

PR #42, opened 2026-08-01
Refuse --force-release on a healthy running job unless --force

jangtrinh:
## The hole

`figma-agent job <id> --force-release` freed ANY slot-holding job as long as it was the
current slot holder — it never checked whether the job was still healthily running. An
impatient caller could free another job's slot while its script was still mid-mutation,
letting the next queued mutation dispatch into the same file while the first was still
executing: genuinely concurrent writes to one file.

## The health signal (already in the code)

The watchdog marks a job `state !== 'running'` once it declares the job wedged (no reply
within the watchdog window) — the slot stays held on purpose so the caller can still
force-release it. That's the whole distinction: a job is "healthy" only while
`state === 'running'` AND its elapsed time is still inside the watchdog window.

## The fix

- `cli/src/transport/job-table.ts` — a pure `isHealthyRunningJob(job, now, watchdogTimeoutMs)`
  predicate, unit-tested directly.
- `cli/src/transport/broker-daemon.ts` — the force-release handler now refuses a healthy
  job unless the request carries `override: true`; a wedged job is unaffected. Every
  force-release (wedged or `--force` override) now logs the requester's `activity` label
  via a new pure `buildForceReleaseAuditLine` helper — no force-release is silent anymore.
- `cli/src/commands/job.ts` — `--force-release` now threads `override: args.bool('force')`
  and labels the request `Force-release · <jobId>` for the audit log.
- `cli/src/figma-agent.ts` — `job` help block documents `--force`.

## Test-first

- `tests/job-table.test.ts` — `isHealthyRunningJob` unit tests (red before the predicate
  existed; the daemon had no health check at all).
- `tests/force-release-guard.test.ts` — the audit-line builder always names the requester.
- `tests/job-command.test.ts` — `run()` threads `override`/`activity` into `runCommand`
  correctly for both a bare `--force-release` and `--force-release --force`.
- `tests/broker-daemon-harness.test.ts` — three end-to-end cases against the real
  `handleJobCommand`/`advanceQueue` closures: (1) a healthy running job is refused, slot
  stays held, nothing advances; (2) the same job with `--force` proceeds, the queued job
  advances; (3) regression — a watchdog-wedged job still force-releases with a bare
  `--force-release`, unchanged. **Not run in this session** — this machine has a live
  Figma plugin connected and this harness's own daemon log write isn't scratch-redirected
  (issue #32), matching the file's own existing precedent for its `--instance` tests a few
  lines above. `vitest list` confirms all three collect cleanly with no syntax/import
  errors; left for a machine without a live session (or CI) to execute.

## Gate tails

- `npm run typecheck` — pass, clean.
- `npm run build` — pass (both bundles built).
- `npx vitest run tests/job-table.test.ts tests/force-release-guard.test.ts tests/job-command.test.ts` — 53/53 pass.

## Deferred

The 3 harness-based force-release tests are written but not executed here — see above.

### DIFF
--- a/cli/src/commands/job.ts
+++ b/cli/src/commands/job.ts
@@ -8,6 +8,8 @@
 //   figma-agent job --list [--file <name>]
 //   figma-agent job <jobId> --cancel            (queued only)
 //   figma-agent job <jobId> --force-release     (unblock a file a wedged script still holds)
+//   figma-agent job <jobId> --force-release --force  (override a HEALTHY still-running job;
+//                                                      discards its result, unverified)
 import type { CommandArgs } from '../figma-agent.ts';
 import type { JobInfo } from '../../../shared/protocol.ts';
 import { CliError, ChunkAssembler, isChunkMsg, parseWireMsg } from '../transport/protocol-helpers.ts';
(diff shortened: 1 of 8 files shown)


jangtrinh:
**APPROVE** — stage-4 review (orchestrator re-ran gates on the branch: typecheck ✅, build ✅, 53 unit tests ✅ across job-table + force-release-guard + job-command).

The guard is correct and the invariant it protects is exactly right: `isHealthyRunningJob` = `state === 'running' && startedAt !== undefined && now - startedAt < WATCHDOG_TIMEOUT_MS`. A healthy job is refused without `--force` (its script may be mid-mutation — freeing the slot would recreate the concurrent-write the whole wave prevents); a watchdog-wedged job (`state !== 'running'`, because the watchdog already called `finish()`) is NOT healthy and keeps unwedging with a bare `--force-release` — the legitimate path does not regress. Refusal message states elapsed + watchdog threshold + the discard consequence. Every force-release (allowed or overridden) logs the requester `activity` — never silent.

Two executor calls I confirm as good:
- Putting `isHealthyRunningJob` in `job-table.ts` next to `isFinishedState` (outside the spec's literal file list) — correct home for a pure predicate, and it's what makes the guard unit-testable without the broker harness.
- The refusal/audit wording is the executor's own construction — reviewed, it's clear, actionable, and honest ('result will be discarded, unverified').

The 3 end-to-end harness tests are written + collect-verified but not run (live plugin session, issue #32) — deferred to the orchestrator, same precedent as #36/#40, and the executor confirmed zero pollution of the live broker log. Comment hygiene is clean — no plan-id/issue-# for this work leaked (the CI guard #41 will sweep the pre-existing external-constraint refs repo-wide).

PR #43, opened 2026-08-01
Refuse design-file-only ui.* commands cleanly outside Figma

jangtrinh:
## The gap

Design-only `ui.*` commands (components, variants, instances, slots) had no reverse
editor guard. In FigJam or Slides, `ui.componentSet({ base: 'X', ... })` would not
refuse — it proceeded into component logic and died with a confusing
`base must be a COMPONENT node id, got not found: X`, which reads like a bad id
rather than "wrong editor". The forward guards (`ui.figjam.*`/`ui.slides.*` calling
`requireEditor`) already refuse cleanly in the wrong editor; this closes the reverse
direction.

## Fix

One shared helper, `requireDesignFile(capability)` beside `requireEditor` in
`plugin/src/main/exec-stdlib-editor.ts`:

```ts
export function requireDesignFile(capability: string): void {
  requireEditor(capability, ['figma']);
}
```

Called as the FIRST line of each gated entrypoint — before any arg validation or
node lookup — so the refusal names the capability, the wrong editor, and the fix,
matching the forward guards' message shape.

## Classification (verified per command, not assumed)

**Gated** — each depends on a Figma design-file concept with a documented or
type-level Figma-Design-only dependency:
- `ui.componentSet` — `figma.combineAsVariants` is typed `"Note: This API is only
  available in Figma Design"`.
- `ui.slot.create` / `ui.slot.list` / `ui.slot.append` / `ui.slot.reset` /
  `ui.slot.addProperty` — a SLOT only exists on a COMPONENT/COMPONENT_SET/INSTANCE,
  none of which any FigJam board or Slides deck can ever contain.
- `ui.setProps` / `ui.swapInstance` — `componentProperties`/`swapComponent` are
  INSTANCE-only; no FigJam/Slides node type is ever `INSTANCE`.

These already had their own type checks (`node.type !== 'COMPONENT'` /
`inst.type !== 'INSTANCE'`), so they were not crash-prone — just confusing. Those
checks are unchanged and still guard the orthogonal bug they always did (a
wrong-type node passed IN Figma itself, where the new gate is a no-op).

**Deliberately left ungated** (checked against `@figma/plugin-typings`, not
assumed) — gating these would be an over-gating regression, not a fix:
- `ui.vars.*` (rename/remove/describe/addMode/renameMode/removeMode/setModeValue) —
  `figma.variables.*` carries no "Figma Design only" note anywhere in the typings.
- `ui.boundFill` — same family (`node.setBoundVariable` /
  `setBoundVariableForPaint`), same absence of a documented restriction.
- `ui.annotate.*` (get/set/categories) — `AnnotationsMixin` is implemented by base
  shape/text node types (`RECTANGLE`, `TEXT`, `VECTOR`, …) that also exist in
  FigJam and Slides, and `AnnotationsAPI` carries no editor restriction either.

This matches and extends this repo's own prior reverse-guard audits
(`knowledge/figjam.md`, `knowledge/slides.md`), which had already flagged
`vars`/`annotate` as unverified rather than confirmed-restricted — both docs are
updated in this PR to record the resolved decision.

## Test-first

New suite (`tests/exec-stdlib-design-file-guard.test.ts`) proves, per gated command:
refuses in FigJam with the editor-named message before any arg validation (fails
red pre-fix, passes post-fix); does not refuse in a Figma design file. A regression
guard proves `byPath`/`q` (generic traversal) and the left-ungated `vars.*`/
`boundFill`/`annotate.*` still execute in FigJam — over-gating those would be a
regression, not a fix.

Existing tests for the now-gated commands (`exec-stdlib-component-set.test.ts`,
`exec-stdlib-slot.test.ts`, `exec-stdlib-props.test.ts`,
`exec-stdlib-instance-swap.test.ts`) needed `setMockEditorType('figma')` added
alongside their existing `installMockFigma()` calls, since the mock defaults to no
reported editor type (matching the real "unknown host" refusal contract) and these
tests exercise the design-file happy path.

## Gates

- `npm run typecheck` — pass
- `npm run build` — pass (plugin/code.js content-hash build id updated accordingly)
- `npx vitest run --config vitest.config.ts` — 1173 passed (only unrelated failure:
  the `kernel/design-os` submodule not checked out in this worktree, a pre-existing
  environment condition, not a regression from this change)

### DIFF
--- a/knowledge/figjam.md
+++ b/knowledge/figjam.md
@@ -41,30 +41,49 @@ Audited, not assumed: the bar is "would raw-throw a confusing platform error, or
 fabricate/mutate wrongly" — a helper that already refuses cleanly via its own
 existing type check does NOT need a new guard.
 
-- **`ui.componentSet` / `ui.slot.*` — already safe.** `figma.createComponent()` /
-  `figma.combineAsVariants()` ARE documented `"Note: This API is only available in
-  Figma Design"` in the typings, but every call site (`exec-stdlib-component-build.ts`,
-  `exec-stdlib-slot.ts`) checks `node.type !== 'COMPONENT'` on every input BEFORE
-  reaching those APIs — since no FigJam node can ever be type `COMPONENT`, that check
-  always fires first, with an already-clear message. The raw Figma-Design-only API is
-  never reached with a bad type.
-- **`ui.setProps` — already safe.** Explicit `inst.type !== 'INSTANCE'` check before
-  any INSTANCE-only call.
-- **`ui.swapInstance` — the one real gap, fixed generally, not with an editor guard.**
-  Had NO type check on `inst` at all — a caller passing any non-INSTANCE node hit a
-  raw, uncoded platform error (`in swapComponent: node is not an instance`). This
-  pre-dates FigJam (already reachable in a design file with e.g. a FRAME); FigJam just
-  makes it far more likely. Fixed with the same `inst.type !== 'INSTANCE'` check
-  `setProps` already had — a general fix, not an editor guard, since an editor guard
-  would leave the same hole open for a wrong-type node in Figma itself.
-- **`ui.annotate.*` — already safe.** `get`/`set` gate via `requireAnnotatable`'s
-  `'annotations' in node` check first; `categories()`'s only call to the file-level
-  `figma.annotations.getAnnotationCategoriesAsync()` is already wrapped in try/catch →
-  clean `E_EVAL` (issue #2's own BLOCKER 1 fix). Whether annotations are actually
-  available in FigJam at all is `[re-verify]` — no editor-type note found in the
-  typings either way.
-- **`ui.vars.*` — likely already safe, `[re-verify]`.** Variable CRUD operates
-  entirely through `figma.variables.*` (document-level, not node-type-dependent) — no
+**Update — a second pass front-loaded a clean refusal ahead of the type checks
+below.** "Already safe" (below) meant non-crashing, not clean: a caller in FigJam
+still saw a raw `not found: X` once a bogus id failed the type check, which reads
+like a bad id rather than "wrong editor". `ui.componentSet`, `ui.slot.*` (create/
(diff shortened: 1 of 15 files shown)


jangtrinh:
**APPROVE** — stage-4 review (orchestrator re-ran gates in the branch worktree: typecheck ✅, build ✅, 82 tests ✅ across the new design-file-guard test + the 4 updated command tests). The one vitest failure the executor reported is `kernel/design-os` submodule-not-checked-out — an env condition (fresh worktree, no submodule init), unrelated to this change and present on main too.

The classification is the crux and the executor got it right — and conservatively, which is the correct bias here:
- **Gated (7):** `ui.componentSet`, `ui.slot.{create,list,append,reset,addProperty}`, `ui.setProps`, `ui.swapInstance` — each now calls `requireDesignFile(capability)` as its FIRST line, so a FigJam/Slides caller gets a clean editor refusal instead of a downstream `not found: X` that reads like a bad id.
- **Left ungated (`ui.vars.*`, `ui.boundFill`, `ui.annotate.*`):** deliberately, backed by `@figma/plugin-typings` (no documented editor restriction — variables are document-level, annotations cover node types that exist in FigJam/Slides) AND this repo's own prior audit (`knowledge/figjam.md`/`slides.md` already marked them `[re-verify]`, never confirmed-restricted). This matches the spec's explicit rule: ambiguous → leave ungated, since a false refusal is a regression while a missing guard is status quo. The knowledge docs are updated to record the resolved decision honestly (distinguishing 'non-crashing' from 'clean refusal').
- `byPath`/`q` (generic traversal) untouched — the regression the classification had to avoid.

Fix is at the shared layer (one `requireDesignFile` helper, routed to each site — no per-site `['figma']` re-derivation). Comment hygiene clean — no plan-id/issue-# for this work leaked. Live FigJam/Slides canvas re-verification wasn't available in the executor env; the confirmed bug (componentSet in FigJam) is exactly what the gated path now refuses, and I verified `--instance`-targeted componentSet-in-FigJam produced that raw error live earlier — this closes it.

PR #45, opened 2026-08-01
Durable per-file contention counter for queued mutation time

jangtrinh:
## Why

`queuedMs` (per-file time a mutation waited in the FIFO before dispatch, `job-table.ts`'s `markRunning`/`cancelQueued`) only lives on FINISHED job records, which are capped at `JOB_FINISHED_CAP = 200` and evicted after `JOB_TTL_MS = 10 min`. The locked decision to consider finer-than-per-file locking is gated behind "revisit only when `queuedMs` measures > 5 min/day for real" — but a per-file daily total built only from the job table can never accumulate that long. The trigger was permanently un-fireable. This is instrumentation only: no locking or queue behavior changes.

## Durable store design

- `cli/src/transport/contention-log.ts` — a small JSON store keyed `{ [fileSlug]: { [YYYY-MM-DD]: { totalQueuedMs, jobCount } } }`, a deliberate near-copy of `error-log.ts`'s path handling (broker-cwd-independent default + explicit-project routing, same `file-identity.ts` slug every other durable feed uses) but a different shape: a running total, not an append-only event log. `addQueued(path, fileSlug, queuedMs, now)` write-throughs with UTC-day bucketing; retention prunes any day older than 30 days on the next write to that file's own bucket (deterministic by date, stated in a comment — never a silent drop).

## Daemon-owns-IO seam

`job-table.ts` stays socket/IO-free per its own top-of-file invariant. `broker-daemon.ts` adds one `recordContention(rec)` closure, called right after every one of the 7 places a job's terminal state is set (`st.jobs.finish(...)` × 6 call sites, `st.jobs.cancelQueued(...)` × 1) — it resolves the same bound-vs-broker-cwd path DOC_CHANGE/EDIT_FEED/the error log already use, then write-throughs `rec.queuedMs` if defined. A job that fails straight out of the queue without ever running (`markRunning` never called) carries no `queuedMs`, and the write-through is a documented no-op for that case — an existing gap in what `job-table.ts` measures, not something this change alters.

## Reader

`figma-agent contention [--file <name>] [--since <days>]` (`cli/src/commands/contention.ts`, registered in `figma-agent.ts`) — pure fs read, works with the plugin closed, mirrors `errors`' shape. Its output is the actual measurement the locked decision needs.

## Test-first

New `contention-log.test.ts` (accumulation, per-day/per-file buckets, retention pruning) and `contention-command.test.ts` (filter/flatten, `run()` over a fixture) — all new code, so these necessarily failed before this change existed. Also appended 3 scenarios to `tests/broker-daemon-harness.test.ts` proving the daemon wiring seam end to end (unbound default path, bound project path, the cancel-while-queued path) — **not run as part of this PR's own gate**, per the live-plugin-session harness constraint; left for the orchestrator to run.

## Gate

- `npm run typecheck` — pass
- `npm run build` — pass
- New unit tests (`contention-log.test.ts` + `contention-command.test.ts`, 27 tests) — pass
- `npm test` / the broker-daemon-harness suite — **not run**, per the live-plugin-session constraint

## Test plan
- [ ] Orchestrator runs the full suite (harness tests included) in an environment without a live plugin session attached

### DIFF
--- a/cli/src/commands/contention.ts
+++ b/cli/src/commands/contention.ts
@@ -1,0 +1,88 @@
+// `figma-agent contention` — the agent's read path over the durable per-file contention
+// counter (`design/figma-contention.json`, contention-log.ts). This is the evidence the
+// locked decision to revisit finer-than-per-file locking depends on ("only when queuedMs
+// measures > 5 min/day for real" — job-table.ts's own decision-6 note): the job table
+// alone cannot answer that question, because its finished records are capped and TTL'd
+// (job-table.ts) long before a day's total could accumulate. This command's output IS
+// the measurement that trigger needs.
+//
+//   figma-agent contention [--file <name>] [--since <days>]
+//
+// Pure fs read, no broker round-trip — works even with the plugin closed, same contract
+// as `changes`/`errors`. Unlike `changes` (one feed file per project file), the store is
+// one shared JSON object keyed by every file's own slug, so `--file` FILTERS entries
+// (case-insensitive substring against the raw fileSlug key) rather than picking a file on
+// disk — the same shape as `errors`' own `--file`.
+import type { CommandArgs } from '../figma-agent.ts';
+import { CliError } from '../transport/protocol-helpers.ts';
+import { contentionLogPath, readContentionStore, type ContentionStore } from '../transport/contention-log.ts';
+
+export interface ContentionRow {
+  fileSlug: string;
+  day: string;
+  totalQueuedMs: number;
+  jobCount: number;
+}
+
+export interface ContentionFilter {
+  file?: string;
+  /** Only days within the last N UTC calendar days, inclusive of today. */
+  sinceDays?: number;
+}
+
+/** `--since <days>` must be a non-negative finite number — `args.num` already rejects a
+ *  non-numeric string, but (same gap `changes.ts`'s `validateLimit` closed for `--limit`)
+ *  lets `Infinity`/negative values through as clean numbers that would otherwise silently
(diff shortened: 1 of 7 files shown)


jangtrinh:
**APPROVE** — stage-4 review (orchestrator re-ran gates in the branch worktree: typecheck ✅, build ✅, 27 unit tests ✅ across contention-log + contention-command).

Design is right: the write-through lives in the daemon (`recordContention` at broker-daemon.ts:705, called after `finish()`/`cancelQueued()`), so `job-table.ts` stays socket/IO-free — its own invariant. The append is best-effort with a `contentionLogAppendFailures` counter + log (Law 1: a dropped write leaves a count, same contract as `appendErrorLog`/`appendEditFeed`). Per-file durable path routes to the bound project or the broker-cwd default, and the store survives the job TTL/cap — which is the whole point: the decision-6 'queuedMs > 5min/day' trigger can finally accumulate and be read via `figma-agent contention`. Instrumentation only — no queue/locking change.

Concerns reviewed, all acceptable:
- 3 harness wiring tests deferred (live plugin session, issue #32) — same precedent as #36/#40/#42/#43, correct.
- A job that fails straight out of queue (never ran → no `queuedMs`) → `recordContention` no-ops, documented in code, not a regression (it never waited-then-ran, so it has no queued-time to record).
- No unbound→bound migration for contention data (unlike error-log) — a deliberate KISS scope cut per the spec's non-goals; acceptable for instrumentation, and flagged honestly for a future revisit if the counters-at-default-path matter.

Comment hygiene: no ephemeral this-work refs leaked.

PR #46, opened 2026-08-01
fix: make the broker daemon's log path injectable

jangtrinh:
## Summary

`BrokerDaemonOptions` already injects `advertisePath`, `ports` (harness uses `[0]`, OS-ephemeral), and `exit`, so the in-process daemon harness (`tests/broker-daemon-harness.test.ts`) never touches the real `/tmp/figma-agent-broker.json` advertisement or the real 9410-9419 port range. One shared resource was still hardcoded: `LOG_FILE = '/tmp/figma-agent-broker.log'` (`broker-daemon.ts`), written by `log()` via `appendFileSync`. A harness-spawned broker therefore appended to the real log a live dev session shares.

- Added `logFile?: string` to `BrokerDaemonOptions`; `runBrokerDaemon` resolves `activeLogFile = options?.logFile ?? LOG_FILE` once at startup, before any `log()` call. Production behavior (no options passed) is unchanged.
- The harness now passes a scratch `logFile` (same tmpdir as its other overrides) in its one `runBrokerDaemon` call site.
- Added `tests/broker-daemon-log-file-default.test.ts`: mocks `node:fs`'s `appendFileSync` (wrap, not replace — same convention as `broker-advertisement-atomic.test.ts`) to prove (a) omitting `logFile` still resolves to the real default path, and (b) an injected `logFile` receives every write while the default path is never appended to. The mock never lets a write actually reach the real default path, so this test cannot interfere with a live session even in a failure case.

With the harness now fully isolated, its previously-deferred `--instance` / force-release / contention tests run for the first time. That surfaced two pre-existing bugs in those already-written fixtures (unrelated to the log-path fix):
- A bound-project fixture never created the target's `design/` dir, which `resolveProjectDir`'s `isUsable()` requires — it silently fell back to the broker's cwd default instead of the bound project.
- A `--instance` targeting a non-connected instance correctly parks (same as any unmatched filter) for the plugin-wait window, which defaults to 12s — longer than vitest's 5s test timeout. Shrunk via the already-existing `FIGMA_AGENT_PLUGIN_WAIT_MS` override, the same pattern the watchdog tests already use.

## Test plan
- [x] `npm run build` — pass
- [x] `npx vitest run --config vitest.config.ts tests/broker-daemon-harness.test.ts` — **22/22 pass** (previously 0 of these deferred cases had ever run)
- [x] `npx vitest run --config vitest.config.ts tests/broker-daemon-log-file-default.test.ts` — 2/2 pass
- [x] `npm test` (full suite, submodule initialized) — **104 files / 1319 tests pass**
- [x] Verified only these two test files call `runBrokerDaemon`; neither binds the real 9410-9419 range nor writes the real `/tmp/figma-agent-broker.log`

### DIFF
--- a/cli/src/transport/broker-daemon.ts
+++ b/cli/src/transport/broker-daemon.ts
@@ -52,6 +52,13 @@ import type { ComponentChange } from '../../../shared/figma-changes.ts';
 import type { EditInput, EditSource } from '../../../shared/edit-feed.ts';
 
 const LOG_FILE = '/tmp/figma-agent-broker.log';
+// Resolved once at `runBrokerDaemon` startup (default: the constant above, unchanged
+// production behavior) — the log path is injectable so a test broker never appends to
+// the shared /tmp log a live dev session reads. `log()` reads this mutable binding
+// rather than the constant directly; safe as module state because each daemon run
+// resolves it exactly once, before any request handling (and therefore any `log()`
+// call) can occur.
+let activeLogFile: string = LOG_FILE;
 
 // envMs (protocol-helpers.ts) lets manual acceptance shrink the idle-shutdown /
 // heartbeat / plugin-wait knobs to seconds — shared with broker-discovery.ts so
@@ -162,7 +169,7 @@ interface BrokerState {
 
 function log(line: string): void {
   try {
-    appendFileSync(LOG_FILE, `${new Date().toISOString()} [${process.pid}] ${line}\n`);
+    appendFileSync(activeLogFile, `${new Date().toISOString()} [${process.pid}] ${line}\n`);
   } catch { /* logging is best-effort */ }
 }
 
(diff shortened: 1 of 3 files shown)


jangtrinh:
**APPROVE** — stage-4 review + independent verification. I re-ran the gates in the branch worktree: typecheck ✅, build ✅, and ran the harness suite MYSELF — **24/24 pass** (22 broker-daemon-harness + 2 log-file-default), including the 4 describe blocks (`--instance`, force-release, contention, read-only) that were written-but-deferred across #36/#40/#42/#43/#45. Confirmed the isolation holds: the real `/tmp/figma-agent-broker.log` mtime did NOT change across my run (its last write is the live session's own broker, not the suite) — the scratch `logFile` redirect works.

The production change is minimal and default-preserving: `activeLogFile = options?.logFile ?? LOG_FILE`, resolved once at `runBrokerDaemon` startup, `log()` writes to it. No `logFile` → the exact `/tmp/figma-agent-broker.log` as before. Only 3 files touched (the DI + 2 test files); zero broker-routing logic changed. The 2 incidental fixture bugs fixed (missing `design/` dir, missing wait-ms shrink) are test-only and were only reachable once the deferred blocks could run.

This closes the accumulated 'deferred to orchestrator' debt: those harness tests are no longer theoretical — they run and pass. Comment hygiene clean (the stale 'issue #32 / not run by executor' blocks are correctly de-staled now that isolation is real).

One benign note (not blocking): the suite emits a MaxListenersExceededWarning (11 brokers each add a SIGTERM/SIGINT listener) — a test-harness artifact, not a leak in production; worth a `setMaxListeners` in the harness someday.

PR #48, opened 2026-08-01
perf: dedupe node fetches and bound the walk in ui.figjam.connections

jangtrinh:
## What

`ui.figjam.connections` was fetching each endpoint's node twice per
connector: `resolveEndpoint` calls `getNodeByIdAsync(ep.endpointNodeId)`
to build the serializable `EndpointResult`, then the loop right below it
calls `getNodeByIdAsync(ep.nodeId)` again for the same id to build the
`connectedNodes` entry — 4 async fetches per connector where 2 suffice.

`resolveEndpoint` now returns the already-fetched node alongside its
result (an internal-only `ResolvedEndpoint`, stripped back down to the
public `EndpointResult` shape before it goes into `edges`), and the loop
reuses that reference instead of re-fetching.

`figma.currentPage.findAll(n => n.type === 'CONNECTOR')` still walks the
whole page — connectors can nest, unlike `board()`'s top-level-only
read, so `findAll` stays. What's now bounded is the *result*: capped at
`MAX_CONNECTORS_READ`, mirroring `board()`'s existing
`MAX_BOARD_READ_NODES` / `truncated` pattern. `totalConnectors` still
reports the true page count; `truncated` is set when the cap bites, so
a caller can tell some connectors weren't processed rather than the
walk silently stopping. The existing `unresolved: true` handling for an
endpoint whose node can't be resolved is unchanged — an edge is never
dropped either way.

## Test-first

Added to `tests/exec-stdlib-figjam-read.test.ts`:
- a `getNodeByIdAsync` call-count assertion (spy on the mock) for a
  single connector between two nodes — this failed against the
  pre-change code (`4 times`, not `2`) and passes after the dedupe.
- a bounded-walk test: more connectors than `MAX_CONNECTORS_READ` →
  exactly the cap in `edges`, `truncated: true`, no throw.
- an at/under-cap test: `truncated` falsy, everything present.
- the existing `unresolved: true` edge-never-dropped test still passes
  unchanged.

## Gate (worktree)

- `npm run typecheck` — pass
- `npm run build` — pass
- `npx vitest run` on the figjam test family (`exec-stdlib-figjam-read`,
  `-content`, `-table`, `-arrange`) — 44/44 pass

Full `npm test` / broker-daemon-harness intentionally not run here
(live-plugin-session territory, tracked separately) — this change is
plugin-unit level with a mocked `figma`, which is the right gate for it.

### DIFF
--- a/plugin/code.js
+++ b/plugin/code.js
@@ -3368,6 +3368,7 @@
   var MAX_CODE_BLOCK_CHARS = 5e4;
   var MAX_ARRANGE_NODES = 500;
   var MAX_BOARD_READ_NODES = 1e3;
+  var MAX_CONNECTORS_READ = 1e3;
   var STICKY_COLORS = {
     YELLOW: { r: 1, g: 0.85, b: 0.4 },
     BLUE: { r: 0.53, g: 0.78, b: 1 },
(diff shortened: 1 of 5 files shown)


jangtrinh:
**APPROVE** — stage-4 review (re-ran in branch worktree: typecheck ✅, figjam-read suite 11/11 ✅, no ephemeral comment-hygiene leak). The dedupe is the real win and the output contract is preserved: `resolveEndpoint` returns an internal `ResolvedEndpoint` (node ref) the loop reuses, and `toEndpointResult()` strips the node before it reaches the public shape — so callers see exactly the same fields, at 2 fetches/endpoint instead of 4. Test-first proven (stashing the fix made the call-count assertion fail 4≠2). The bounded walk keeps `findAll` (connectors nest, unlike board's top-level read), caps at `MAX_CONNECTORS_READ = 1000` mirroring `MAX_BOARD_READ_NODES`, sets `truncated: true`, and still reports the true `totalConnectors` — nothing vanishes silently. `unresolved: true` handling untouched and verified. Measured, minimal, contract-preserving.

PR #50, opened 2026-08-01
Pin daemon routing to a plugin via the panel (#35 P2)

jangtrinh:
## Summary

P1 (#36) gave the agent exact targeting via `--instance <id>`. This is P2: a
human one-click way to pin the daemon's routing to the plugin they're
looking at, so subsequent no-flag commands route there without juggling
instanceIds.

**Precedence chain** (extends P1's):
`--instance` (per-request) > `--file` (per-request) > **`targetInstancePin`
(daemon runtime, new)** > `FIGMA_AGENT_FILE` (env) > active (recency).

**Daemon** (`broker-daemon.ts`):
- `targetInstancePin: string | null` — set/cleared by the panel, consulted in
  `admitRequest` only when neither per-request flag is present. A live pin
  routes through P1's own `kind:'instance'` path (no new lookup). A pin
  pointing at a gone instance refuses the command outright
  (`E_TARGET_DISCONNECTED`, naming the instance) rather than silently
  falling through to the env pin or recency — a standing pin must never
  quietly re-point at another plugin.
- The pin clears automatically the moment its instance disconnects, so a
  dead pin can never wedge future no-flag commands.
- New panel→daemon messages `SET_TARGET`/`CLEAR_TARGET`, carrying the
  panel's own instanceId — validated against the sender's own registered
  socket, so a panel can only ever pin *itself*.
- `PEERS` gains a `pinned` boolean alongside `isActiveTarget`, so a panel can
  tell "target because pinned" from "target because most-recent".

**Routing** (`route-filter.ts`): `resolveRouteFilter` takes the (already
liveness-checked) pin as a new precedence tier; `pinDisconnected` is the pure
predicate for the refuse-vs-fall-through decision — both unit-tested
directly, independent of any registry/socket mocking.

**Panel**: a "Target this plugin" button (`panel.html`) that reads
"Targeted ✓" when this panel is the pin (click clears it), wired through
`ui-relay.ts`'s existing DOM-event plumbing (the same shape `SYNC_REQUEST`
already uses) and `panel-ui.ts`/`panel-model.ts` for the toggle label + the
pinned/active context note.

**Non-goals** (per spec): no change to `--instance`/`--file` semantics, no
auto-pinning, no persistence across broker restarts (runtime-only, like the
rest of the connection state).

## Test plan

- [x] `npm run build` — pass
- [x] `npm run lint:comments` — clean (0 new citations)
- [x] `tests/route-filter.test.ts` — precedence + `pinDisconnected`, unit,
      pure (new + existing pass)
- [x] `tests/panel-model.test.ts` — `fileNote` pinned wording +
      `targetButtonLabel` (new + existing pass)
- [x] `tests/broker-daemon-harness.test.ts` — pin routing precedence
      (beats recency, per-request flags still override), `SET_TARGET`/
      `CLEAR_TARGET` lifecycle, disconnect-clears-pin, `PEERS`
      `pinned`/`isActiveTarget` (6 new tests + all 22 pre-existing pass, 28
      total) — uses the harness's own scratch `logFile`/ephemeral-port
      isolation, never touches a live broker session
- [ ] Manual: open two Figma files, click "Target this plugin" on one,
      confirm a no-flag `figma-agent` command lands there regardless of
      which file was most recently active

## Notes

- The E_TARGET_DISCONNECTED refusal path is unit-tested at the pure
  `pinDisconnected` seam (deterministic) rather than raced end-to-end in the
  harness — the daemon's own disconnect handler clears the pin synchronously
  with the registry removal, so there's no naturally reachable window in a
  harness test where the pin is live-set-but-dead; the pure predicate is the
  clearer, non-flaky place to prove that branch.

### DIFF
--- a/cli/src/transport/broker-daemon.ts
+++ b/cli/src/transport/broker-daemon.ts
@@ -25,7 +25,7 @@ import {
 } from './protocol-helpers.ts';
 import { PluginRegistry, type PluginEntry } from './plugin-registry.ts';
 import { buildBrokerHelloData, noPluginMessage } from './broker-status.ts';
-import { resolveRouteFilter, type RouteFilter } from './route-filter.ts';
+import { pinDisconnected, resolveRouteFilter, type RouteFilter } from './route-filter.ts';
 import {
   appendChangeFrames, changeLogPathFor, migrateLegacyUnboundChanges, migrateStagedChanges,
   unboundStagingPath, unboundStagingRoot,
@@ -483,6 +483,13 @@ export async function runBrokerDaemon(options?: BrokerDaemonOptions): Promise<vo
   const idleMs = readIdleMs();
   let syncInFlight = false;
 
+  // Target pin (#35 P2): the panel's "Target this plugin" button routes here — a
+  // deliberate human act, ranked above the env pin and recency but below a per-request
+  // `--instance`/`--file` (see route-filter.ts). RUNTIME-ONLY (never persisted, never
+  // survives a broker restart, like every other piece of `st`) and cleared the moment its
+  // instance disconnects (`handleClose`) — a pin must never keep pointing at a dead plugin.
+  let targetInstancePin: string | null = null;
+
   /** Send one unsolicited EventMsg to a single socket (best-effort). */
   const sendEvent = (ws: WebSocket, type: EventMsg['type'], data: Record<string, unknown>): void => {
     try { if (ws.readyState === WebSocket.OPEN) ws.send(JSON.stringify({ type, data } satisfies EventMsg)); }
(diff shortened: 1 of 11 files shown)


jangtrinh:
**APPROVE** — stage-4 review (re-ran in worktree: typecheck ✅, lint:comments clean ✅, 60 route-filter+panel-model tests ✅, executor's full run 88 incl. 6 new harness tests). Precedence is exactly right and Law-1-safe: the runtime `targetInstancePin` is consulted BEFORE `resolveRouteFilter` (so per-request `--instance`/`--file` still win), ranks above the env pin + recency, and a no-flag command with the pin set but its instance GONE refuses with `E_TARGET_DISCONNECTED` — never silently falls through to env/active. The pin clears on the pinned instance's disconnect (`handleClose`), and PEERS `pinned` = live-pin outranks recency so the panel honestly shows 'commands go here because you pinned it'. `pinDisconnected` is a pure predicate composing the daemon's own liveness check — the right, non-flaky seam for the disconnect case (the daemon clears synchronously, so there's no real race to force end-to-end).

The AGENT already had `--instance` (P1/#36); this gives the HUMAN the one-click 'quick active' the owner asked for. Two-Figma-file manual verification is correctly flagged owner-manual, not claimed. Comment hygiene passes (the narrowed option-2 matcher allows the bare `#35` phase refs by design — the accepted trade-off to avoid hex false-positives).

PR #51, opened 2026-08-01
Add zombie-watchdog observability (transport vs app liveness)

jangtrinh:
## Why

The broker conflates two liveness layers today: **transport liveness** (a WS
`pong`, which Chromium auto-answers from its network process even while the
iframe's JS is frozen) and **app liveness** (a JS-originated frame — HELLO,
FILE_INFO, a command reply/chunk — which dies the instant the freeze hits).
Because the pong keeps `lastSeenAt` fresh, a fully-frozen plugin instance
reads as perfectly healthy today. That is the invisible failure mode this PR
makes visible.

A secondary, less severe mode also exists: a throttled-but-not-frozen
background tab whose socket drops and reconnects repeatedly with an
unchanged scene.

## What this PR adds (observability only)

- `PluginEntry` gains `lastAppFrameAt`, `reHelloCount`, `sameSceneStreak`,
  `lastReHelloAt`.
- The daemon's app-frame frame-receipt site now calls a new
  `registry.touchApp(ws)` (bumps `lastAppFrameAt` + `lastSeenAt`). The raw WS
  `'pong'` handler is unchanged — it still calls the existing `touch(ws)`
  (transport-only, never touches `lastAppFrameAt`). **This one split is the
  whole design.**
- `register()` tracks a same-scene reconnect streak, windowed to 10 minutes
  so a daily sleep/wake reconnect (same scene, hours apart) never
  accumulates toward the flapper threshold. `touchActive` and a
  scene-*changing* `updateScene` both reset the streak — any real activity
  proves the instance isn't flapping right now.
- A pure predicate, computed at read time (no new timer):
  `suspectedZombie = appSilenceMs > FROZEN_AFTER_MS || sameSceneStreak >= 3`.
  `FROZEN_AFTER_MS` defaults to 75s, overridable via `FIGMA_AGENT_ZOMBIE_MS`
  (same `envMs` pattern as the daemon's other cadence knobs).
- `PluginStatusEntry` gains `appSilenceMs` (always present — the raw sensor,
  so an agent can judge staleness itself) and `reHelloCount` /
  `suspectedZombie` / `zombieReason` (present only when they carry
  information, same contract as the existing `senderMismatchCount` field —
  the healthy zero-case payload stays byte-identical).
- A log line fires only on a flag **flip**, piggybacked onto the existing
  WS-ping heartbeat interval — no new timer, no per-tick spam.

## Not building (deliberate, filed separately)

- **No remediation** — a zombie's socket is never terminated here. A frozen
  iframe can't run its own reconnect loop, so killing the socket would
  convert a visible zombie into an invisible absence.
- **No routing change** — a flapping zombie can still steal the routing
  target via `register()`'s recency bump. Left as a separate follow-up
  (deprioritizing a suspected zombie in `selectTarget`).
- No plugin-side/relay changes, no persistence (broker-restart blindness is
  a separate, optional follow-up).

## Tests

`tests/plugin-registry.test.ts` (all fake sockets + an injected clock, no
live broker):
- **The load-bearing assertion**: a transport `touch` (pong) does NOT clear
  the frozen flag; only `touchApp` does.
- Predicate boundary: exactly at `FROZEN_AFTER_MS` is not flagged, one ms
  over is.
- 3 same-scene reconnects within the window flag as a flapper; a
  scene-changing `updateScene` or `touchActive` resets the streak; an
  identical-scene `updateScene` does NOT reset it (no change = no proof);
  daily-cadence (>10 min apart) reconnects never accumulate past streak 1.
- Present-only serialization: `appSilenceMs` always present, zero-case
  payload has no `suspectedZombie`/`zombieReason`/`reHelloCount` keys;
  `reHelloCount` appears once > 0 even before any flag trips;
  `suspectedZombie` + `zombieReason` appear together once frozen.

## Gate

- `npm run typecheck` — pass
- `npm run build` — pass
- `npx vitest run tests/plugin-registry.test.ts tests/broker-status.test.ts tests/executor-ops-status.test.ts` — 67 passed
- `npm run lint:comments` — clean, 0 new citations

Did not run the full `npm test` suite per scope (broker-daemon-harness spawns
a live broker process; out of scope for this pure-registry/status change).

## Owner-manual (not done here, flagging)

Live smoke test: background a FigJam window for ~5 minutes, then confirm
`figma-agent status` shows `suspectedZombie: true` with a LOW
`lastHeartbeatAge` (transport still ponging) but a high `appSilenceMs`. This
also verifies the Chromium-auto-pong assumption this whole design rests on —
if the frozen iframe instead gets culled as WS-dead before this fires, the
failure mode is different from what's assumed here and the broker-restart
persistence follow-up becomes the primary fix instead.

### DIFF
--- a/cli/src/transport/broker-daemon.ts
+++ b/cli/src/transport/broker-daemon.ts
@@ -23,7 +23,7 @@ import {
   ChunkAssembler, deleteConnectionChunk, envMs, getConnectionChunks, isChunkMsg, isEventMsg, isReplyMsg, isRequestMsg,
   parseWireMsg, rawToString, sendWireMsg, sweepAbandonedChunks, type ChunkBuffers,
 } from './protocol-helpers.ts';
-import { PluginRegistry, type PluginEntry } from './plugin-registry.ts';
+import { PluginRegistry, suspectedZombie, type PluginEntry } from './plugin-registry.ts';
 import { buildBrokerHelloData, noPluginMessage } from './broker-status.ts';
 import { pinDisconnected, resolveRouteFilter, type RouteFilter } from './route-filter.ts';
 import {
@@ -1289,7 +1289,10 @@ export async function runBrokerDaemon(options?: BrokerDaemonOptions): Promise<vo
     if (!msg) return;
     // Hidden control frame from a newer CLI build replacing this broker.
     if ((msg as { type?: string }).type === 'BROKER_SHUTDOWN_REQUEST') shutdown(0, 'BROKER_SHUTDOWN_REQUEST');
-    const isPlugin = st.registry.touch(ws); // any plugin frame = LIVENESS (heartbeat cull)
+    // Any parsed frame here only exists because real JS constructed and sent it —
+    // this is the APP-liveness sensor (zombie-watchdog observability), distinct from
+    // the transport-only `touch` the raw WS 'pong' handler below uses.
+    const isPlugin = st.registry.touchApp(ws);
     if (isChunkMsg(msg)) {
       // Reply-side chunks (plugin → broker) pass straight to routeFromPlugin, which now
       // accumulates every frame onto the job record itself (concurrency & jobs, backlog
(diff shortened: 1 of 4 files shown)


jangtrinh:
**APPROVE** — stage-4 review (re-ran in worktree: typecheck ✅, lint:comments clean ✅, 59+ registry/status tests ✅; executor's full focused run 67/67). The design's one load-bearing idea is implemented correctly: app frames go through the new `touchApp` (bumps `lastAppFrameAt`), while the WS pong handler stays on plain `touch` — so a frozen iframe whose Chromium network stack keeps auto-ponging still goes stale on the sensor that matters. `suspectedZombie = now - lastAppFrameAt > FROZEN_AFTER_MS (75s, env-overridable via FIGMA_AGENT_ZOMBIE_MS) || sameSceneStreak >= threshold` — primary app-silence detector + secondary flapper streak, exactly the advisor's design. Status fields are present-only (appSilenceMs always; suspectedZombie/zombieReason/reHelloCount only when meaningful) so the zero-case payload stays byte-identical and existing snapshot tests pass. The `zombieReason` string is the human signal ('app-silence Nms, same-scene streak N'). Observability only — no remediation, no routing change, no persistence, matching the NOT-building list. Live smoke (background a FigJam window 5 min → status shows suspectedZombie with low lastHeartbeatAge, which also falsifies the Chromium-auto-pong assumption) is correctly flagged owner-manual.

PR #52, opened 2026-08-01
fix: deprioritize suspected-zombie plugins in default routing selection

jangtrinh:
## The trap

`register()` bumps `lastActiveAt` on every (re-)HELLO, and `matching()`/`selectTarget()` sort live entries by `lastActiveAt` descending. A throttled-background plugin that flaps (reconnects every ~1-5 min with an unchanged scene) keeps re-stealing the freshest `lastActiveAt`, so a no-flag command could silently route to a frozen/backgrounded editor instead of the one the user is actually working in — even though the zombie-watchdog detector (`suspectedZombie`) already exists and correctly flags it.

## Fix

`matching()`'s unfiltered (no `--file`/no `--instance`) branch now deprioritizes any live entry where `suspectedZombie(entry, now)` is true: non-suspected entries sort ahead of suspected ones, recency-ordered within each group. When every live entry is suspected, the group collapses to one and the most-recent among them is returned — a fallback, never a refusal, since a suspected zombie is still a connected, usable plugin.

Explicit targeting is untouched: `--instance <id>` resolves via `getByInstanceId` (bypasses this sort entirely) and a filtered `matching()`/`selectTarget()` call (`--file <name>`, exact or substring) keeps pure recency, so it still reaches a suspected zombie by name on purpose.

## Test-first

Added to `tests/plugin-registry.test.ts`:
1. One healthy + one flapping-zombie entry (zombie made the *more*-recent via `lastActiveAt`) — a no-flag `selectTarget()` returns the healthy one. Verified this fails against the pre-fix code (returns the zombie by recency).
2. `--instance <zombie-id>` and exact `--file <zombie-name>` still resolve to the zombie.
3. All live entries suspected → returns the most-recent among them, not null.
4. No suspected zombies → identical to prior pure-recency behavior.

## Gate tails

- `npm run typecheck` — clean
- `npm run build` — clean
- `npm run lint:comments` — clean (0 new citations)
- `npx vitest run tests/plugin-registry.test.ts` — 48/48 passed

### DIFF
--- a/cli/src/transport/plugin-registry.ts
+++ b/cli/src/transport/plugin-registry.ts
@@ -290,19 +290,34 @@ export class PluginRegistry<S extends RegistrySocket = RegistrySocket> {
     }
     const live = this.liveEntries();
     const hits = f ? live.filter((e) => fileMatches(e.scene.fileName as string | undefined, f, opts?.exact === true)) : live;
+    // A named/filtered ask is a deliberate targeting decision — it always resolves by
+    // pure recency, reaching a suspected zombie if that's what the name (or instance,
+    // handled above) names. Only the UNFILTERED default selection deprioritizes a
+    // suspected zombie: it must never win the no-flag routing slot over a healthy
+    // plugin purely by having re-HELLO'd more recently. When every live candidate is
+    // suspected, the group collapses to one and recency alone decides — a fallback,
+    // never a refusal, since a suspected zombie is still a connected, usable plugin.
+    if (!f) {
+      const now = this.now();
+      return [...hits].sort((a, b) => {
+        const zombieRank = Number(suspectedZombie(a, now)) - Number(suspectedZombie(b, now));
+        return zombieRank !== 0 ? zombieRank : b.lastActiveAt - a.lastActiveAt;
+      });
+    }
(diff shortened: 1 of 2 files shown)


jangtrinh:
**APPROVE** — stage-4 review (gated the change: typecheck ✅, lint:comments clean ✅, plugin-registry suite 48/48 ✅; test-first confirmed red pre-fix). The deprioritization is surgical and correct: only `matching()`'s UNFILTERED default path partitions — `Number(suspectedZombie(a,now)) - Number(suspectedZombie(b,now))` sorts healthy (0) ahead of suspected (1), then recency within each group. The `--instance` (kind:'instance') and exact `--file` paths are untouched, so explicit targeting still reaches a suspected zombie on purpose (verified by dedicated tests). All-suspected falls back to most-recent — never a refusal, since a suspected zombie is still a connected, usable plugin. Good self-catch on the test where 3×60s ticks tripped the FROZEN branch on the healthy fixture by elapsed time alone — switching to 2s ticks isolates the intended same-scene-streak branch, so the test passes for the RIGHT reason.

Note for the record: the #52 commit briefly landed on the executor's local `main` by mistake — it was never pushed to origin/main (origin stayed at #51's `44f230b`); I reset my local main to origin to drop the stray before merging this PR cleanly.

PR #53, opened 2026-08-01
Surface awaitingReconnect visibility for plugins not yet reconnected after a broker restart

jangtrinh:
## Summary

- A freshly-spawned broker's registry starts empty, so a backgrounded editor (FigJam/Slides, throttled by Figma so its reconnect loop stalls) is invisible after a broker restart until it happens to reconnect — the focus-dance of clicking every window to revive it.
- The broker now persists the live plugin set (`{instanceId, fileName, lastSeenAt}` only), debounced (~1-2s, flushed on every change so the on-disk snapshot is at most one window stale), to `last-plugins.json` beside its own advertisement file — never relying on a shutdown hook, since the broker can be SIGKILL'd.
- On startup, entries last seen within 10 minutes of this broker's start are held as a top-level `status.awaitingReconnect: [{ fileName, lastSeenBeforeShutdown }]` hint. An entry clears the moment that instanceId — or fileName as a fallback, since a fresh iframe load mints a new instanceId — re-HELLOs. The whole list expires once this broker has been up 30 minutes, even if nothing reconnected.
- `awaitingReconnect` is a hint from last-known state, never a claim of a live connection: it is never mixed into the live `plugins[]` array, and its own shape carries no `instanceId` so it can't be mistaken for an addressable target.

Non-goals: no auto-reconnect (that needs the owner's manual GUI click), no change to the existing zombie-watchdog detection or the live `plugins[]` truth.

## Test plan

- [x] `npm run build` — pass
- [x] `npx vitest run tests/last-plugins-log.test.ts` — 18/18 pass (pure fs-layer + filtering functions: round-trip persistence, threshold filtering, clear-on-reconnect by instanceId/fileName-fallback, 30-minute expiry, status shape never carries instanceId)
- [x] `npx vitest run tests/broker-daemon-harness.test.ts` — 30/30 pass, including two new end-to-end scenarios across two real in-process broker instances sharing a scratch advertisement path: a restart surfaces `awaitingReconnect` while `plugins[]` stays genuinely empty, then a fileName-matched re-HELLO clears it; and the list expires after a shrunk uptime window even with nothing reconnected
- [ ] **Owner-manual, not claimed here**: restart the real broker with a backgrounded FigJam window connected — confirm `figma-agent status` reports it under `awaitingReconnect` until the window is focused and reconnects

## Notes

- Only ran the specific test files above per the task's constraint — never the full `npm test` / a live broker — and everything uses scratch advertisement dirs/log files, never the real `/tmp/figma-agent-broker.json`.

### DIFF
--- a/cli/src/commands/status.ts
+++ b/cli/src/commands/status.ts
@@ -88,5 +88,11 @@ export async function run(args: CommandArgs): Promise<unknown> {
   return {
     broker, plugins, activePlugin, plugin, protocolVersion: broker.protocolVersion,
     ...(wanted ? { pluginsAll: all } : {}),
+    // Broker-restart reconnect visibility — a HINT from last-known state, never a live
+    // plugin (it must never appear in `plugins`/`pluginsAll` above); present only when
+    // the broker actually has one to report, same mirror-only-when-non-empty contract
+    // as `senderMismatchCount`/`legacyMigrationDeferred` on `broker` above.
+    ...(Array.isArray(hello.awaitingReconnect) && hello.awaitingReconnect.length > 0
+      && { awaitingReconnect: hello.awaitingReconnect }),
   };
 }
--- a/cli/src/transport/broker-daemon.ts
+++ b/cli/src/transport/broker-daemon.ts
@@ -25,6 +25,10 @@ import {
 } from './protocol-helpers.ts';
 import { PluginRegistry, suspectedZombie, type PluginEntry } from './plugin-registry.ts';
 import { buildBrokerHelloData, noPluginMessage } from './broker-status.ts';
+import {
+  clearReconnected, filterAwaitingReconnect, lastPluginsPathFor, readLastPlugins,
+  toAwaitingReconnectStatus, writeLastPluginsAtomic, type AwaitingReconnectEntry, type LastPluginRecord,
+} from './last-plugins-log.ts';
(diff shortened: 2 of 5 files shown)


jangtrinh:
**APPROVE** — stage-4 review (gated the change: typecheck ✅, lint:comments clean ✅, last-plugins-log 18/18 ✅; the executor also ran the broker-daemon-harness 30/30 incl. 2 new e2e scenarios — now safe post-#46 via scratch logFile). The design is honest and Law-1-clean: `last-plugins.json` is written debounced (1.5s, env-overridable) beside the advertisement file, read on startup, entries held only if last-seen < 10 min before start, cleared per-entry on re-HELLO (by instanceId OR fileName, since a fresh iframe mints a new id), and the whole list expires at 30 min uptime. `awaitingReconnect` is present-only on `status`/HELLO and — the key invariant — NEVER appears in the live `plugins[]` array (that stays = actually-connected truth); it's explicitly a hint from last-known state. This closes the broker-restart blind spot that caused the focus-dance: a fresh broker now names which backgrounded editors haven't come back.

For the record: the #53 commit again briefly landed on the executor's local `main` — never pushed (origin stayed at `27c818b`); I reset local main to origin before merging. The live confirmation (restart the broker with a backgrounded FigJam window → status lists it in awaitingReconnect) is correctly flagged owner-manual.

Issue #56, opened 2026-08-12
auto-connect: figma-agent introduces itself to every agent (skill + hook + peek + presence + cowork)


jangtrinh:
# auto-connect: figma-agent introduces itself to every agent (skill + hook + peek + presence + cowork)

> Draft GitHub issue body. `plans/` is gitignored — this carries what must survive machine loss.
> Full local plan: `plans/260812-2130-auto-connect/plan.md` (+ `phase-00..04`).
> Approved product specs: `plans/specs/auto-connect/brainstorm.md`, `.../usecases.md`.
> Planned against HEAD `2537b3f`.

## Outcome

An agent starting a fresh session already knows whether a broker is alive, which Figma files
are connected, and whether the plugin build matches the CLI — without being told. It knows
every command from a skill the CLI itself emits, so the reference can never drift from the
binary. The human's one remaining step (Figma gives no API to open a plugin) shrinks to a
click on a `figma://` link. When the designer edits the agent's work, the agent can wait
exactly one designer change-cycle and read back what changed.

Root problem, stated once: the transport was never the gap. The broker auto-spawns, the plugin
auto-reconnects. **The gap is that an agent has no knowledge of the connect path at the moment
a session begins.**

## Invariants (each already paid for in this repo)

1. **One shared reader.** `--help` and the skill's command reference render from ONE catalog; a
   drift test fails when the committed `SKILL.md` differs from the emitter's output. A
   prose-only standard drifts — ship the emitter and the linter together.
2. **A session-start hook never spawns a daemon.** `status --peek` reads the `/tmp`
   advertisement plus at most one short WS query. It must not reach `ensureBroker()`.
   Exit 0 even when idle; target p95 < 200 ms.
3. **No config write without consent.** Confirm, back up, abort on invalid JSON, `--dry-run`.
4. **Additive envelope only.** New fields are omitted when unset, so an unguarded frame
   serializes byte-identically to what a pre-flag CLI sent.
5. **No second memory store.** `cowork` reads the existing edit feed and the existing
   supervised-memory ledger (see #54 for what a `designer` event actually means).
6. **Nothing vanishes silently.** Every skip, refusal, and timeout carries a count or a named
   reason. `corrections: null` + a reason, never `[]` standing in for "unknown".

## Slices (each independently shippable)

- **Slice 0 — green baseline.** `main` has 2 failing tests, both in
[...]

PR #57, opened 2026-08-12
fix(panel): drop banned unicode glyphs from comments and the target-pin label

jangtrinh:
## Summary

Restores a green test suite by removing unicode mark glyphs (checkmark, cross, ellipsis, multiplication sign) that had crept into `panel.html` and `panel-model.ts`, tripping the panel's owner-locked typography contract (every visual mark must be an SVG, never text).

Two kinds of violation:
- Comment prose (measurement notes like `1168×992`, `13 × 1.5`, an inline ellipsis) — rewritten in plain ASCII, no behavior change.
- The pinned "Target this plugin" button's own label returned `"Targeted ✓"` as a literal rendered string instead of letting the icon carry the mark, unlike every other status label in this module (see `syncResultLabel`'s existing `"Synced — ..."` precedent). The label now reads `"Targeted"`; the pin icon still shows visually. The one dependent unit test asserting the old string was updated to match.

## Known flake (not a regression)

`tests/broker-daemon-harness.test.ts`, describe block `daemon harness — target pin (#35 P2)`, intermittently fails with a `waitFor: condition never became true within the deadline` timeout on different sub-tests across runs. Reproduced identically before and after this change; one full-suite run passed 109/109 files clean. Left untouched — not caused by, or fixed by, this PR.

## Test plan

- [x] `npm test` — one clean run: `Test Files 109 passed (109)`, `Tests 1445 passed (1445)`
- [x] `npm run build` — clean
- [ ] Not verified: whether the flaky `target pin (#35 P2)` suite is CI-environment-specific vs. local-machine timing; left as pre-existing per the plan's own note.

Related: #56

### DIFF
--- a/plugin/src/ui/panel-model.ts
+++ b/plugin/src/ui/panel-model.ts
@@ -92,10 +92,10 @@ export function fileNote(count: number, isActiveTarget: boolean, pinned = false)
 }
 
 // ─── Target pin (#35 P2) — the "Target this plugin" button's own label ──────
-// A toggle: pinned → "Targeted" (click clears it, no check-mark glyph — the icon
-// carries the mark, matching syncResultLabel's precedent below); unpinned →
-// "Target this plugin" (click sets it). Kept as a pure mapping so the DOM glue
-// (panel-ui.ts) never composes this string itself.
+// A toggle: pinned → "Targeted" (click clears it, no check-mark glyph — this
+// button has no icon, so the label TEXT alone carries the pinned/unpinned
+// state); unpinned → "Target this plugin" (click sets it). Kept as a pure
+// mapping so the DOM glue (panel-ui.ts) never composes this string itself.
 export function targetButtonLabel(pinned: boolean): string {
   return pinned ? 'Targeted' : 'Target this plugin';
 }
--- a/plugin/src/ui/panel-model.ts
+++ b/plugin/src/ui/panel-model.ts
@@ -92,11 +92,12 @@ export function fileNote(count: number, isActiveTarget: boolean, pinned = false)
 }
 
 // ─── Target pin (#35 P2) — the "Target this plugin" button's own label ──────
-// A toggle: pinned → "Targeted ✓" (click clears it); unpinned → "Target this plugin"
-// (click sets it). Kept as a pure mapping so the DOM glue (panel-ui.ts) never composes
-// this string itself.
+// A toggle: pinned → "Targeted" (click clears it, no check-mark glyph — the icon
+// carries the mark, matching syncResultLabel's precedent below); unpinned →
+// "Target this plugin" (click sets it). Kept as a pure mapping so the DOM glue
+// (panel-ui.ts) never composes this string itself.
 export function targetButtonLabel(pinned: boolean): string {
-  return pinned ? 'Targeted ✓' : 'Target this plugin';
+  return pinned ? 'Targeted' : 'Target this plugin';
 }
 
 // ─── Idle-commit sync prompt (spec 004 P4) ────────────────────────────────────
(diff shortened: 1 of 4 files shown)


jangtrinh (commented):
_Posted as a review comment: GitHub does not allow an approve/request-changes state on a pull request opened by the same account. The verdict below is the review state._

**Verdict: approve.** No blockers. Gates re-run independently in a detached worktree; the deviation flagged in the description checks out against the module's own precedent.

## Verified, not taken on trust

**The two failures this PR claims to fix are real and are at `main`.** Re-run at `2537b3f`:

```
$ npx vitest run --config vitest.config.ts tests/figma-plugin-panel.test.ts
 FAIL  tests/figma-plugin-panel.test.ts > the plugin panel — typography contract (owner-locked)
       > draws every mark as an SVG — no text glyphs in markup, CSS, or the panel scripts
AssertionError: glyph in panel.html: expected true to be false
 ❯ tests/figma-plugin-panel.test.ts:327:52
      Tests  2 failed | 41 passed (43)
```

Green on this branch. The gate was fixed, not weakened — `GLYPHS` at `tests/figma-plugin-panel.test.ts:325` is unchanged.

**No rebuild drift.** The usual failure mode for a change that touches both `plugin/src/ui/panel.html` and the generated `plugin/ui.html` is a hand-edited bundle. Ran `npm run build` at this branch's tip and `git status --porcelain` came back empty — `plugin/ui.html` and `plugin/code.js` are exactly what the source compiles to, including the `"Targeted ✓"` → `"Targeted"` change at `plugin/ui.html:1022`.

**The label deviation matches the module's stated precedent.** `plugin/src/ui/panel-model.ts:122` already documents it in so many words: *"No glyph: the icon beside this line (a Phosphor check-circle) carries the mark instead."* Dropping the literal check-mark from `targetButtonLabel` puts it in line with `syncResultLabel`, and the one dependent assertion was updated rather than deleted.

`npm run typecheck`, `npm run build`, `npm run lint:comments` all exit 0 on this branch. Full suite: 114 files / 1471 tests passed. The `target pin (#35 P2)` flake called out in the description did not reproduce in my run — noted, not counted against this PR.

## Non-blocking

**`plugin/src/ui/panel-model.ts:93-95` — the new comment claims an icon that this button does not have.**

```
// A toggle: pinned → "Targeted" (click clears it, no check-mark glyph — the icon
// carries the mark, matching syncResultLabel's precedent below); unpinned → ...
```

The precedent is sound for `syncResultLabel`, which does render beside a check-circle. This button has no icon:

- `plugin/src/ui/panel.html:924` — `<button class="sync-btn" id="fga-target-btn" type="button">Target this plugin</button>`, no icon child
- `plugin/src/ui/panel.html:669` — `#fga-target-btn { margin-top: var(--space-1); }`, no `::before`/`::after` content
- `plugin/src/ui/panel-ui.ts:101` — `targetBtn.textContent = targetButtonLabel(peersPinned);` — `textContent` replaces all children, so an icon could not survive here even if one were added to the markup

Behavior is fine: the label text itself still distinguishes the two states ("Target this plugin" vs "Targeted"), so nothing is lost visually. The problem is only that the comment (and the PR description's "the pin icon still shows visually") states a constraint that is not true of this element, which will mislead whoever next edits the button. Either drop the "the icon carries the mark" clause and say the label text carries the state, or render an icon into the button and stop using `textContent` to write it.

jangtrinh:
Fixed the non-blocking finding: `plugin-model.ts:93-95`'s comment claimed the pinned-target button had an icon carrying the check mark. It doesn't (`textContent` replaces all children, no icon in the markup) — reworded to say the label text itself carries the state. Comment-only, pushed as `1fad52b`. Rebuild confirmed no drift (`plugin/ui.html`/`plugin/code.js` unchanged beyond the source edit); `npm run typecheck`/`lint:comments`/`build` all exit 0; panel + panel-model tests green.

PR #58, opened 2026-08-12
feat(cli): status --peek, emitted skill, and skill/hook installers

jangtrinh:
## Summary

First slice of auto-connect (#56): a session-start-safe way for an agent to know the connect state before doing anything else, plus a one-command way to install the CLI's own skill and hook.

- **`status --peek [--json]`** — reads only the `/tmp` broker advertisement plus one short broker query with its own deadline. Never calls the broker-spawning path `status` normally uses, so it's safe from a Claude Code `SessionStart` hook every session, plugin open or not. Reports `versionMatch`/`protocolMatch` against the connected plugin build — `null` (not `false`) when the plugin never reported a version, so "unknown" is never confused with "mismatched". `--json` is accepted for compatibility but stays a documented no-op — this CLI always prints one JSON object.
- **The broker registry now retains `pluginVersion`/`protocolV`** from every `PLUGIN_HELLO` instead of discarding them (they were already being stripped out of the scene payload only) — surfaced additively on `status`'s `plugins[]` rows.
- **`--help` and a new emitted skill (`skills/figma-agent/SKILL.md`) render from the same command catalog** — a drift test fails the moment the committed skill file stops matching a fresh `npm run emit:skill`, or a command is registered without a catalog entry (or vice versa).
- **`install-skill`** — writes the emitted skill into a Claude Code skills folder. Version-compares against anything already there and never silently overwrites; a non-interactive run needs an explicit `--with-craft`/`--no-craft` to decide whether to also bundle `skills/es-figma-craft`.
- **`install-hook`** — adds the `SessionStart` hook to Claude Code settings. Confirms before writing, backs up the existing file first, and aborts untouched (no backup, no write) if the file isn't valid JSON. The hook command itself no-ops silently (`|| true`) on a machine without this CLI.

## Known flake (not a regression)

Same as the prior PR (#57): `tests/broker-daemon-harness.test.ts`'s `target pin (#35 P2)` describe block intermittently times out on `waitFor` — reproduced identically on this branch, with and without this change.

## Test plan

- [x] Red-first: all 5 new test files fail with "Cannot find module" against pre-phase-01 source (verified via `git stash` of the source-only changes, tests kept) before implementing.
- [x] `npm test` — same pre-existing flake as PR #57, otherwise green; all new tests pass every run.
- [x] `npm run build` — clean, `plugin/` untouched (this slice is CLI-only)
- [x] UC-01 (no broker): `status --peek --json` exits 0, `connected:false`, `idle:true`, creates no advertisement file
- [x] UC-01 budget: idle peek well under 200ms
- [x] UC-03: hand-tampered `skills/figma-agent/SKILL.md` makes the drift test fail with a message naming the exact stale line; regenerating with `npm run emit:skill` restores green

## Not verified (owner live-test items, per the plan)

- `install-skill --folder ~/.claude/skills` against a real folder, then a fresh Claude Code session picking up the skill
- `install-hook --dry-run` then a real install, confirming a real session with the plugin open AND closed both inject sanely and never spawn a broker
- `versionMatch` reading `true` against the currently loaded plugin build in a live Figma session

Related: #56 — based on #57 (phase-00 green baseline); merge that first.

### DIFF
--- a/cli/src/command-catalog.ts
+++ b/cli/src/command-catalog.ts
@@ -1,0 +1,201 @@
+// The ONE structured source of every command's name/usage/description and every
+// global flag. `help-text.ts` renders `--help` from this; `skill-emitter.ts`
+// renders the skill's "Command reference" section from the SAME array — a
+// command can never be documented in one place and not the other (drift test:
+// tests/skill-emitter-drift.test.ts). Data only, no rendering, no fs.
+//
+// Content is a verbatim transcription of the former inline `HELP` template
+// literal (figma-agent.ts) — wording unchanged, only line-wrapping collapsed
+// into single `description` strings per entry.
+
+export interface CommandCatalogEntry {
+  /** The key COMMAND_MODULES (figma-agent.ts) dispatches on. */
+  name: string;
+  /** Usage + prose, exactly as it read in the old --help block (line-wraps collapsed). */
+  description: string;
+}
+
+export interface GlobalFlagEntry {
+  flag: string;
+  description: string;
+}
+
+export const HEADER_LINE = 'figma-agent — CLI bridge to the Figma plugin (via a local WS broker)';
+export const USAGE_LINE = 'Usage: figma-agent <command> [options]';
+export const FOOTER_LINE =
+  'All commands print one JSON object to stdout and exit 0, or {error:{code,message}} and exit 1.';
+
+export const COMMANDS: CommandCatalogEntry[] = [
+  {
+    name: 'status',
+    description:
+      'Broker + plugin connection info [--peek [--json]] — --peek reads only the /tmp broker '
+      + 'advertisement plus one short broker query; it NEVER spawns a broker. --json is accepted '
+      + 'for compatibility but is a no-op: output is always the same single JSON object.',
+  },
+  { name: 'seat', description: 'Probe seat → {seat, bridge, reason} [--seat free|paid skips the probe]' },
+  {
+    name: 'bind',
+    description:
+      '--file "<name>" --dir <projectDir>   bind a file to a project for panel/idle sync '
+      + '(refuses to guess otherwise) [--list] [--unbind]',
+  },
+  { name: 'get-selection', description: 'Serialize the current selection [--depth 1]' },
+  { name: 'inspect', description: '[nodeId|--node id] [--out file.png --scale 1 --timeout ms]' },
+  {
+    name: 'job',
+    description:
+      '<jobId> [--wait] [--wait-timeout 60000] | --list [--file name] | <jobId> --cancel (queued only) | '
+      + '<jobId> --force-release [--force]   poll/wait/cancel/list a job the CLI stopped waiting for '
+      + '(backlog 1.1+2.6+4.3) — --force-release refuses a HEALTHY still-running job unless --force is '
+      + 'also passed — --force overrides the guard and discards its result, unverified; a '
+      + 'watchdog-wedged job still unwedges without --force',
+  },
+  {
+    name: 'scan-design-system',
+    description: 'Components/variables/styles registry [--out file.json --timeout ms]',
+  },
+  {
+    name: 'scan-node',
+    description: '[SPIKE] Reverse-walk one node → FigmaExportNode spec <nodeId> [--timeout ms]',
+  },
+  {
+    name: 'mirror-verify',
+    description: 'Prove one node round-trips: scan → rebuild → scan → diff <nodeId> [--parent id --keep --timeout ms]',
+  },
+  {
+    name: 'scan-conventions',
+    description:
+      'Convention-DNA walk over sections → usage-dna.json '
+      + '[<sectionId...> --out file.json --budget 14000 --timeout ms]',
+  },
+  {
+    name: 'audit-ds',
+    description:
+      "DS-hygiene audit of the open file's component library "
+      + '[--out file.json --sections "01 A,02 B" --facts raw.json --from-facts raw.json --timeout ms]',
+  },
+  { name: 'create-frame', description: '--name n --w 400 --h 300 [--parent id --x 0 --y 0]' },
+  {
+    name: 'connect',
+    description: '--from id --to id [--label t --intent flow|annotation --flow n --transition id]',
+  },
+  { name: 'disconnect', description: '--id conn-id | --from id --to id' },
+  { name: 'list-connections', description: '' },
+  { name: 'reroute', description: '[--id conn-id | --flow name]' },
+  { name: 'draw-flow', description: '--flow path/to/flow.json [--page name]' },
+  { name: 'verify-connections', description: '[--flow path/to/flow.json]' },
+  { name: 'create-instance', description: '--component <key|id> [--parent id]' },
+  { name: 'set-variant', description: '--node id --props k=v,k2=v2' },
+  {
+    name: 'create-variable',
+    description: '--collection c --name n --type COLOR|FLOAT|STRING|BOOLEAN --value v [--mode m]',
+  },
+  {
+    name: 'bind-variable',
+    description: "--node id --field fills|cornerRadius|... --variable <id|name>",
+  },
+  {
+    name: 'set-autolayout',
+    description:
+      '--node id --mode H|V|GRID|NONE [--gap n --pad t,r,b,l --align-primary --align-counter --wrap '
+      + '--sizing-h --sizing-v --rows n --cols n --col-sizes ...]',
+  },
+  {
+    name: 'set-constraints',
+    description: '--node id --h MIN|MAX|CENTER|STRETCH|SCALE --v MIN|MAX|CENTER|STRETCH|SCALE',
+  },
+  { name: 'set-text', description: '--node id --chars "..." [--font f --size n --weight n]' },
+  {
+    name: 'clone-traits',
+    description: '--source id --target id --traits layout,fills-variables,typography,spacing,text',
+  },
+  { name: 'sync-corrections', description: '[--dir project] sync Figma edge memory with design/memory' },
+  { name: 'export-png', description: '--node <id|selection> --out file.png [--scale 2]' },
+  {
+    name: 'html-to-figma',
+    description: '--html <file|-> [--width 1280 --x --y --parent id --replace id]',
+  },
+  {
+    name: 'exec-js',
+    description:
+      '<file|-> [--timeout ms (cap 120000)] [--undo-group] — --undo-group brackets the script in ONE '
+      + 'undo step and reverts it on error; the script must not call figma.commitUndo/triggerUndo itself, '
+      + 'and a timeout cannot stop a running script (the plugin has no cancellation). While it runs, '
+      + 'figma.currentPage carries one extra invisible child (the undo sentinel) — a script that '
+      + "enumerates or counts the page's children will see it. `console` and `ui` are injected — a "
+      + 'script cannot declare its own.',
+  },
+  {
+    name: 'capture',
+    description:
+      '<url> [--out dir --headless --channel chrome --width 1440 --timeout ms --carousel-window ms]',
+  },
+  { name: 'batch', description: '<file.json> [--stop-on-error]' },
+  {
+    name: 'changes',
+    description:
+      '[--since ts|iso --owner-only --actor owner|agent|ambiguous --file name|slug --limit 50 --page name]  '
+      + 'read the owner-edit feed (wave 4.4) — pure fs, works even with the plugin closed; --owner-only '
+      + 'is sugar for --actor owner',
+  },
+  {
+    name: 'errors',
+    description:
+      "[--since ts|iso --file name --limit 50]  read the broker's error log (backlog 4.6) — full "
+      + 'untruncated message + cmd/activity/code/fileName, for an agent to read-and-fix; --file filters '
+      + "by the entry's own fileName",
+  },
+  {
+    name: 'contention',
+    description:
+      '[--file name --since days]  read the durable per-file/per-day queued-time counter — total ms a '
+      + 'mutation waited in the FIFO before dispatch, plus jobCount, per UTC day; --file filters by '
+      + 'fileSlug, --since limits to the last N days; pure fs, works with the plugin closed',
+  },
+  {
+    name: 'install-skill',
+    description:
+      '[--folder ~/.claude/skills] [--with-craft|--no-craft] [--yes]   write this emitted SKILL.md to a '
+      + 'folder Claude Code reads skills from; prompts before overwriting an older installed version and '
+      + 'before bundling skills/es-figma-craft — a non-interactive run needs --with-craft/--no-craft '
+      + 'or it skips craft and says so',
+  },
+  {
+    name: 'install-hook',
+    description:
+      '[--dry-run] [--yes] [--settings <path>]   add a SessionStart hook to Claude Code settings that '
+      + 'runs "figma-agent status --peek" at the start of every session; confirms before writing, backs '
+      + "up the file first, and aborts untouched if it can't parse the file as JSON",
+  },
+];
+
+export const GLOBAL_FLAGS: GlobalFlagEntry[] = [
+  {
+    flag: '--file "<exact file name>"',
+    description:
+      "route to that file's plugin AND refuse to run anywhere else (exact, case-insensitive; beats "
+      + 'FIGMA_AGENT_FILE; payloads >512KB route by the env pin but are still guarded)',
+  },
+  {
+    flag: '--instance <instanceId>',
+    description:
+      "route to that EXACT plugin instance (from `status`'s instanceId column) — beats --file and "
+      + "FIGMA_AGENT_FILE; the one way to target a specific instance when --file's name is ambiguous "
+      + '(e.g. two unsaved files both named "Untitled"). Mutually exclusive with --file: passing both '
+      + 'refuses with E_INVALID_ARGS naming both, it never silently picks one.',
+  },
+  {
+    flag: '--dir <projectDir>',
+    description:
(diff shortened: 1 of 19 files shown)


jangtrinh (commented):
_Posted as a review comment: GitHub does not allow an approve/request-changes state on a pull request opened by the same account. The verdict below is the review state._

**Verdict: request changes.** Two blockers, both reproduced below. The core of the slice is sound — `status --peek` really is spawn-free, the catalog really is one shared reader, and the envelope change really is additive — but one declared-green gate is red on this branch, and one advertised flag cannot work from the shipped binary.

Gates re-run in a detached worktree at `3e837d6` (`npm ci`, submodule initialised):

```
typecheck EXIT=0
build      EXIT=0
lint:comments EXIT=1      <-- BLOCKER 1
npm test:  Test Files 114 passed (114) / Tests 1471 passed (1471), exit 0
```

The `target pin (#35 P2)` flake did not reproduce in my run — noted, not counted against this PR.

---

## BLOCKER 1 — `npm run lint:comments` fails on this branch

The description checks this off as clean. It is not:

```
$ npm run lint:comments
comment-hygiene: 1 citation(s) not covered by the baseline.

State the invariant directly instead of citing an issue/PR number, a dated plan-dir
slug, or a bare finding/audit label:

  cli/src/commands/install-hook.ts:5: [dated plan-dir slug]
  // --yes) → backup → write. See plans/260812-2130-auto-connect/phase-01-*.md §1d.
EXIT=1
```

Not pre-existing — the same command at the base commit `a9473e2`:

```
comment-hygiene: clean (55 file(s) with baselined citations, 0 new).
EXIT_AT_57=0
```

`cli/src/commands/install-hook.ts:5` cites a plan directory that is not in this repository, so the pointer is dead for anyone reading the file from a clone. The order-of-operations sentence immediately above it already states the invariant completely; drop the trailing `See plans/...` clause.

## BLOCKER 2 — `install-skill --with-craft` cannot work from the shipped binary

`cli/src/commands/install-skill.ts:47-49`:

```ts
function repoRoot(): string {
  return resolve(dirname(fileURLToPath(import.meta.url)), '..', '..', '..');
}
```

Three levels up is correct from `cli/src/commands/`. The shipped entry point is `cli/dist/figma-agent.js` (`package.json` `bin`, `scripts/build.mjs:53`), which is two levels below the root — so the same expression lands one directory *above* the repository. Live run of the built bundle:

```
repo root actual      : .../wt-58
repoRoot() from source: .../wt-58                 (craft FOUND)
repoRoot() from bundle: .../scratchpad            (craft MISSING)

$ node cli/<outdir>/figma-agent.js install-skill --folder $TMP --with-craft --yes
{
  "installed": [".../figma-agent/SKILL.md"],
  "skipped": [{
    "path": ".../es-figma-craft",
    "reason": "source skill not found at .../scratchpad/skills/es-figma-craft
               — this install may not bundle it"
  }],
  ...
}
craft dir created at dest? false
EXIT=0
```

So `--with-craft` is a no-op for every real user, exits 0, and reports a path that does not correspond to any checkout. The `skipped` reason keeps this from being a silent loss, but the command still claims success for a flag the caller explicitly passed — an explicitly requested install that could not happen should not exit 0.

The reason this passed review is a seam gap: `tests/install-skill.test.ts:102` (`"--with-craft copies the repo's skills/es-figma-craft directory"`) imports `../cli/src/commands/install-skill.ts` directly, so it only ever exercises the source-tree path arithmetic. It certifies a code path that cannot exist in the artifact users run. `tests/status-peek.test.ts:143-165` already does this correctly — build in `beforeAll`, then run the bundle as a subprocess. Please give the craft copy the same treatment, and resolve the craft source from a build-time constant rather than by counting `..` from `import.meta.url`.

---

## Non-blocking

**1. `cli/src/transport/broker-peek.ts:41-46` — every failure is relabelled as a timeout.**

```ts
promise.catch(() => 'timeout' as const),
```

A connection refusal, a protocol error and an actual deadline expiry all become the same string, and the caller then prints `broker did not answer within ${queryMs}ms` — which is false for the first two. Measured on this branch:

```
CLOSED-PORT     elapsed(ms)= 4    degraded="broker did not answer within 200ms"
SILENT-LISTENER elapsed(ms)= 201  degraded="broker did not answer within 200ms"
```

Four milliseconds, reported as a 200 ms timeout. Carry the real reason through (`err.code`/`err.message` when the promise rejected, the deadline string only when the timer won).

This also means `tests/status-peek.test.ts:133` — *"a live pid that refuses to answer within the deadline reports degraded"* — does not test what its name says. Port `65530` with nothing listening refuses instantly; the 200 ms deadline is never reached. The path the session-start latency budget actually depends on (a broker that accepts the socket and then goes quiet) is untested. The second probe above is the shape that covers it: a plain `net.createServer` that accepts and never speaks.

**2. `cli/src/transport/broker-peek.ts:63-68` — `connected: false` is asserted where the answer is unknown.** This slice is careful to return `null` rather than `false` for `versionMatch` when the plugin never reported a version, and rightly so. The degraded branch does the opposite: a broker that was merely slow yields a flat `connected: false`, indistinguishable from "nothing is running" unless the consumer also reads `degraded`. With the 150 ms default at `:60`, a busy machine can inject "not connected" into a session where the plugin is open, and the agent will tell the user to open a panel that is already open. Consider `connected: null` on that branch, or documenting in `skill-template.ts` that `degraded` must be read before `connected` is believed.

**3. `cli/src/skill-template.ts:23` sends agents to a field the named command does not emit.**

> `status`'s `versionMatch`/`protocolMatch` fields tell you whether the connected plugin build matches this CLI.

`versionMatch` exists only in `broker-peek.ts` (grep across `cli/src` and `shared` returns `broker-peek.ts` and this template line, nothing else). Plain `status` returns `plugins[]` rows from `statusList()`, which carry `pluginVersion`/`protocolV` but no computed match. An agent following the skill reads `undefined` and cannot tell. One word — `status --peek` — fixes it.

Same inaccuracy in a comment: `cli/src/transport/plugin-registry.ts:73` says *"see `statusList()`'s versionMatch/protocolMatch contract"*. `statusList()` has no such contract.

Worth stating plainly, because it bears on invariant 1: the drift test proves `SKILL.md` matches the emitter byte for byte. It cannot prove the emitter's prose is true. Both of these were green through every gate.

**4. The one-catalog guard covers command names and artifact bytes, not flag text.** `tests/skill-emitter-drift.test.ts:32-36` compares `COMMAND_MODULES` keys against `COMMANDS` names, which catches "added a command, forgot the doc". Flags live inside free-text `description` strings with nothing tying them to the `args.bool(...)`/`args.str(...)` calls that read them, so a flag renamed in a command's runtime path leaves both `--help` and `SKILL.md` stale with every gate green. Nothing in `tests/` references `renderHelp` or `help-text` at all, so the `--help` surface itself is asserted nowhere. Every flag in this PR is documented correctly today; the gap is that nothing keeps it that way.

**5. `cli/src/commands/install-hook.ts:74` → `:116` — the file is read before the prompt and written after it.** `raw` is captured at `:74`; the interactive confirm at `:106` can sit for minutes; the write at `:116` serialises from the stale parse. A concurrent writer's edits are silently discarded. The backup at `:112-113` writes `raw` — the pre-prompt content — so it is not a copy of what was actually overwritten, which is the one job a backup has. Re-read and compare before writing (or `fstat` the file and abort if it moved).

**6. `cli/src/commands/install-skill.ts:100` — `cpSync` over an existing craft folder has none of the protection `SKILL.md` gets.** No version check, no prompt, no backup, and `cpSync` defaults to overwriting. A user who edited their installed `es-figma-craft` loses it to `--with-craft`. Given blocker 2, this path is currently unreachable from the shipped binary — worth fixing in the same pass.

**7. `tests/install-hook.test.ts` never covers a settings file that already has unrelated `SessionStart` groups.** That is the realistic case — a working config commonly has several. `addHook` at `:44-49` does spread the existing array correctly, but nothing proves it, and a regression there silently deletes a user's other session hooks. The existing test at `:58` only covers the identical-hook short-circuit.

**8. `cli/src/commands/install-hook.ts:52-54` — `detectIndent` matches spaces only.** `/\n( +)"/` misses a tab-indented file and returns 2, silently reformatting the whole document. Minor, and the backup covers it, but the function's own doc comment claims it matches the file's indentation.

---

## Checked and clear

- **`status --peek` never spawns.** `broker-peek.ts` imports only `readAdvertisement`/`isAdvertisementLive` from `broker-discovery.ts`, never `ensureBroker`; `spawn` appears in that module only inside `spawnBroker()` at `:242`, reachable only via `ensureBroker`. `readAdvertisement` is `readFileSync` + `JSON.parse`, `isAdvertisementLive` is `kill(pid, 0)` plus a clock comparison, `fetchBrokerHello(port)` takes an already-known port and opens one socket. `commands/status.ts:26` returns before `ensureBroker()` at `:28`. The error path returns rather than throws, and `figma-agent.ts:152` `process.exit(0)` means a socket left open by the deadline race cannot hold the process alive. `tests/status-peek.test.ts:161` asserting `readdirSync(noneDir)` is empty — not just that the advertisement is absent — is the right assertion.
- **The envelope change is additive.** `shared/protocol.ts` gains two optional interface fields only; no frame builder is touched. `plugin-registry.ts:353-360` spreads both conditionally (`...(e.pluginVersion !== undefined && {...})`), so a plugin that never sends them produces a byte-identical row.
- **The catalog is genuinely one reader.** `help-text.ts` declares no command name or flag text of its own, and `skill-emitter.ts:30` maps the same `COMMANDS` array. The byte-for-byte comparison at `tests/skill-emitter-drift.test.ts:19` does catch reordering and trailing-whitespace edits in the committed artifact.
- Module sizes are within the repo's limit (largest new file: `command-catalog.ts` at 201 lines, and it is data).

jangtrinh:
Fix round for the request-changes review, pushed as `0472242`. Every item below has a red-first test that failed against the pre-fix code before the fix landed.

**Blocker 1 (`lint:comments` EXIT=1):** `install-hook.ts:5`'s comment cited a `plans/260812-...` path not in this repo. Reworded to state the constraint directly. `npm run lint:comments` now exits 0.

**Blocker 2 (`--with-craft` no-op from the shipped bundle):** `repoRoot()` counted a fixed number of `..` from `import.meta.url`, correct from source but one level too many from the built entrypoint. Now walks up to the nearest `package.json`, which sits at the repo root in both shapes. Added a subprocess test that builds the CLI bundle and runs `install-skill --with-craft --yes` against the actual built artifact — the old source-import test only ever certified path arithmetic no real user runs. An explicitly requested `--with-craft` that still can't be satisfied (source directory genuinely missing) now exits non-zero with a named `E_INVALID_ARGS` reason instead of a silent skip.

**Non-blocking, all addressed:**
1. `broker-peek.ts` no longer relabels every rejected query as a timeout — a refusal and an actual deadline expiry carry distinct messages now. Added the real deadline-path test (a `net.createServer` that accepts and stays silent) and renamed the old closed-port test, which was actually testing an instant refusal.
2. The degraded branch reports `connected: null` (unknown) instead of a fabricated `false` — matches how the rest of this slice already treats an unreported version as unknown, not mismatched.
3. `skill-template.ts` and the `plugin-registry.ts:73` comment both pointed at plain `status` for `versionMatch`/`protocolMatch`; only `status --peek` computes them. Fixed both, re-emitted `SKILL.md`.
4. Extended the drift guard: `renderHelp()` output is now actually asserted against (nothing referenced it before), plus a check that this slice's own commands (`status`, `install-skill`, `install-hook`) don't have an `args.bool`/`args.str` flag that drifted out of their catalog description. Global flags are excluded from the per-command check (they're documented once, in `GLOBAL_FLAGS`).
5. `install-hook` re-reads `settings.json` right before writing and aborts (named reason) if it changed since the pre-prompt read — the confirm prompt can sit for minutes, and a write computed from stale content would silently discard a concurrent edit. Verified with a mocked-readline test that edits the file mid-prompt. The backup now reflects what's actually about to be overwritten, not the pre-prompt snapshot.
6. `install-skill`'s `es-figma-craft` copy gets the same never-overwrite-silently protection `SKILL.md` already had — compares the installed copy's `SKILL.md` content, prompts (or requires `--yes`) before replacing a user's local edits.
7. Added the missing coverage: appending the hook alongside an existing unrelated `SessionStart` group preserves it (this already worked; now it's proven).
8. `detectIndent` now recognizes tab-indented files instead of silently reformatting them to 2 spaces.

**Gates re-run on this branch:**
```
typecheck     EXIT=0
lint:comments EXIT=0
build         EXIT=0
npm test      Test Files 114 passed (114) / Tests 1481 passed (1481) — one clean run;
              the pre-existing target-pin (#35 P2) flake in
              tests/broker-daemon-harness.test.ts reproduced once more in an
              earlier run, isolated to that same describe block, unrelated to
              anything in this PR.
```

jangtrinh (commented):
**Follow-up review of `0472242` — verdict: approve-equivalent. Both blockers are dead; no remaining blockers.** (Posted as a comment because GitHub does not allow an approve state on a pull request opened by the same account.)

Delta only — re-verified in a fresh detached worktree at `0472242` (`npm ci`, submodule initialised). Gates:

```
typecheck      EXIT=0
lint:comments  EXIT=0      <-- was 1
build          EXIT=0
npm test       EXIT=1  — 1 failed | 1480 passed (1481), 113/114 files
```

The single failure is the known pre-existing flake, not this commit:

```
FAIL tests/broker-daemon-harness.test.ts > daemon harness — target pin (#35 P2)
     > SET_TARGET refuses a claimed instanceId that does not match the sender's OWN registered instance
Error: waitFor: condition never became true within the deadline
 ❯ waitFor tests/broker-daemon-harness.test.ts:133:39
```

Ran that file in isolation twice: `Tests 30 passed (30)` both times. Load-dependent under the full parallel suite. Not counted against this PR, but see the note at the end.

## Blocker 1 — closed

`lint:comments` reports `clean (55 file(s) with baselined citations, 0 new)`, exit 0. The comment now states the ordering constraint directly instead of pointing at a directory outside the repository.

## Blocker 2 — closed, and the new anchor holds up

`--with-craft` from the built bundle, which previously resolved above the repo root and silently skipped:

```
$ node cli/<outdir>/figma-agent.js install-skill --folder $TMP --with-craft --yes
{ "installed": [".../figma-agent/SKILL.md", ".../es-figma-craft"], "skipped": [], ... }
exit code: 0
craft dir landed? true (2 entries)
```

Probed the false-anchor concern the walk-up introduces:

- **cwd with its own `package.json`** — ran the same command with `cwd` set to a temp directory containing a decoy `package.json`. Identical result, craft installed. Correct by construction: the walk starts at `dirname(fileURLToPath(import.meta.url))`, so the process working directory never enters it.
- **A nested `package.json` between the module and the root** — there is none on the path. The only `package.json` files in the tree are the root one and four under `kernel/design-os`, all *below* the anchor, never between `cli/dist` and the root. `cli/` itself holds only `dist/`, `src/`, `tsconfig.json`. Walking `cli/dist` → `cli` → root finds the root's on the first hit.
- One residual shape worth keeping in mind, not a blocker: installed as a dependency under `node_modules/<pkg>/`, the walk stops at that package's own `package.json`, which is the right answer as long as `skills/` keeps shipping. There is no `files` field in `package.json`, so it does today.

Unsatisfiable explicit request now fails loudly instead of exiting 0 with a skip — parked the craft source and re-ran:

```
exit code: 1
{"error":{"code":"E_INVALID_ARGS","message":"--with-craft was requested but its source
 skill was not found at .../skills/es-figma-craft — this install does not bundle it"}}
```

## Spot-checks — the finding fixes are behavioral, not cosmetic

**Refusal vs. deadline (`broker-peek.ts:45-80`).** `withDeadline` now returns a discriminated `ok | timeout | refused`, and the refusal branch carries the underlying error through. Measured both paths:

```
REFUSED  elapsed(ms)=  3   connected=null  degraded=broker refused the connection:
                                            broker hello failed: connect ECONNREFUSED 127.0.0.1:65530
TIMEOUT  elapsed(ms)=202   connected=null  degraded=broker did not answer within 200ms
```

Two different facts, two different messages, and the 3 ms case no longer claims a 200 ms wait.

The two tests covering this bracket each other rather than restating the implementation: `tests/status-peek.test.ts:148` asserts `elapsed < 150` **and** `degraded` does *not* match `/did not answer within 200ms/` — a negative assertion aimed at the exact defect — while `:162` asserts `elapsed >= 190` **and** that the message does match. Collapsing the two branches again fails one or the other; neither can pass vacuously. The silent listener is a real `net.createServer` that accepts and never writes, which is the only way to reach the timer branch.

**`connected: null` (`broker-peek.ts:100-108`).** Typed `boolean | null` on the interface with the reason recorded at the field, and returned on both degraded branches (confirmed in the probe output above). The `idle` branch still returns `false`, which is right — there the absence is known.

**TOCTOU (`install-hook.ts:128-146`).** The re-read is after the prompt returns, byte-compares against the pre-prompt read, and aborts on any difference; the backup now writes `fresh.raw`, so it is a copy of what is actually being replaced. The test at `tests/install-hook.test.ts:165` injects a genuine mid-prompt change — it mocks `node:readline/promises` so `question()` itself writes different content to disk before returning `y`, with `isTTY` forced true so the real prompt branch runs — then asserts both that the run rejects and that the concurrent content survives byte-for-byte. That is the actual race window, not a stand-in for it. The mock is at the external boundary and is no more permissive than the real API.

Also confirmed the appending case I raised is now covered (`tests/install-hook.test.ts:150` asserts `SessionStart[1]`, so an existing unrelated group is preserved rather than replaced).

## One note, not blocking

The `target pin (#35 P2)` flake now fails on a full-suite run rather than only intermittently, and it is the same test the issue's known-flake note called out. Two clean isolated runs say it is contention, not logic — but it is currently the difference between `npm test` exiting 0 and 1, which makes the suite's own signal unreliable for the next reviewer. Worth its own issue with the `waitFor` deadline raised or the harness serialised, separate from this PR.

Issue #59, opened 2026-08-12
Flaky: broker-daemon-harness 'target pin (#35 P2)' block fails under machine load (waitFor deadline)


jangtrinh:
Under machine load, tests in the `target pin (#35 P2)` describe block of `tests/broker-daemon-harness.test.ts` fail nondeterministically with `waitFor: condition never became true within the deadline` (:133). Different tests in the block fail on different runs (SET_TARGET-refuses, PEERS-reflects-pin).

Evidence (2026-08-12, all at or based on 2537b3f):
- Untouched `main`, full suite: 1 failed (20:22 run) and 0 failed (a later run) — same HEAD.
- Untouched `main`, file isolated under load: 1 failed / 29 passed.
- `feat/auto-connect-slice-1` worktree, file isolated under load: 2 failed / 28 passed, twice.
- Reviewer machine-state, file isolated: 30/30 twice.

Conclusion: pre-existing timing flake, independent of any PR; contention-sensitive `waitFor` deadlines. It currently tips full `npm test` to exit 1 often enough to mask the suite's signal.

Fix candidates: raise/scale the `waitFor` deadline (env-scaled like batch-timeout-scaling), or serialize the harness describe blocks.

jangtrinh:
Closed by #62. Root cause was NOT test timing: `PluginRegistry.matching()` sorted by recency and fell back to Array.sort stability (registration order) on millisecond ties, so the earliest-registered plugin won ties instead of the most-recently-active one — a real routing bug that also hit production multi-plugin selection. Fixed with a monotonic `activitySeq` tie-break (consulted only on exact-ms ties). The harness now runs clean with zero retries (4/4 and 6/6 verbose runs on the machine that reproduced the failures); a pid-isolation assertion guards against cross-broker contamination. The interval-leak fix in `shutdown()` and the env-scaled waitFor deadline shipped in the same PR.

PR #61, opened 2026-08-12
feat(cli,plugin): agent presence label, status --wait, and a relaunch button

jangtrinh:
## Summary

Second slice of auto-connect (#56): a designer can see which harness is driving the canvas, an agent can block until the plugin actually shows up instead of failing once, and a human told to reopen the plugin gets a relaunch button waiting for them.

- **`--agent <id>` / `FIGMA_AGENT_ID`** — carried through the broker to the panel's activity feed unchanged (`claude · Created frame "Hero card"`, own span, never folded into the sentence text). Trimmed, control characters stripped, capped at 32 chars. Additive on the wire: an unlabelled request's frame stays byte-identical to a pre-flag CLI's — the panel's `cli` default is applied only at render time.
- **`status --wait [--timeout N]`** — blocks until a plugin registers (polling every 500ms), prints a `figma://` deep link for the bound file (or an honest reason it can't build one) plus a one-line hint to stderr the instant the wait starts, and exits non-zero naming the file on timeout instead of hanging or silently giving up. `--timeout` is in seconds (documented explicitly — every other `--timeout` in this CLI is milliseconds). Unlike `--peek`, `--wait` may start a broker on demand.
- **A "Reconnect figma-agent" relaunch button** — `figma.root.setRelaunchData` at plugin startup, guarded (a refusal is logged, never breaks startup).
- `--help` and the emitted skill stay in sync with all of the above (same one-catalog drift guard from the previous slice).

## Not verified (owner live-test items, per the plan)

- The exact `figma://file/<fileKey>` URL scheme against a real Figma desktop install — never fabricated if a binding is missing or has no fileKey (Figma Free), but the URL shape itself is unconfirmed.
- Whether `setRelaunchData` dirties the file, and whether "Reconnect figma-agent" actually appears in Figma's relaunch menu after one run — zero prior use of this API in this repo.
- Two harnesses driving one file with different `--agent` values, distinguished at a glance in a real panel.
- The literal UC-05 gherkin (`status --wait --timeout 5` as a real subprocess) was not run against this machine's shared real broker — a live broker predating this work (pid stable since well before this task started) is running on the real `/tmp/figma-agent-broker.json`/port range, and touching it risked disrupting other in-progress work. The identical code path (ensureBroker → deep-link print → poll → timeout → E_NO_PLUGIN) is fully exercised by `tests/status-wait.test.ts` against an isolated scratch broker instead.

## Test plan

- [x] Red-first: every new test file/case demonstrated failing against pre-change code (envelope additivity, activity-feed agent extraction, `figma-deep-link`, `plugin-wait`, `status --wait` integration) before implementing.
- [x] `npm run build` — clean
- [x] `npm test` — one fully clean run (118 files / 1508 tests); a second run reproduced the same pre-existing `target pin (#35 P2)` flake noted in the prior two PRs (issue #59), isolated to that describe block, unrelated to this change
- [x] Panel typography gate (`tests/figma-plugin-panel.test.ts`) re-run green after every panel edit
- [x] `npm run emit:skill` re-run; drift test green

Related: #56

### DIFF
--- a/cli/src/command-catalog.ts
+++ b/cli/src/command-catalog.ts
@@ -206,9 +206,10 @@ export const GLOBAL_FLAGS: GlobalFlagEntry[] = [
     flag: '--agent <id> | FIGMA_AGENT_ID',
     description:
       'which harness sent this (the flag wins over the env var) — shown in the panel activity feed '
-      + '("claude · Created frame ..."). Trimmed, control characters stripped, capped at 32 chars. '
-      + "Omitted entirely when unset — an unlabelled request's wire frame is byte-identical to a "
-      + "pre-flag CLI's; the panel's own \"cli\" default is applied at render time, never stamped "
-      + 'onto the wire.',
+      + '("claude · Created frame ..."). Trimmed, then validated against an ALLOWLIST: lowercase '
+      + 'letters, digits, ".", "_", "-" only, 1-32 characters — anything else refuses with '
+      + "E_INVALID_ARGS naming the allowed set, it is never silently sanitized. Omitted entirely "
+      + "when unset — an unlabelled request's wire frame is byte-identical to a pre-flag CLI's; the "
+      + 'panel\'s own "cli" default is applied at render time, never stamped onto the wire.',
   },
 ];
--- a/cli/src/command-catalog.ts
+++ b/cli/src/command-catalog.ts
@@ -29,9 +29,13 @@ export const COMMANDS: CommandCatalogEntry[] = [
   {
     name: 'status',
     description:
-      'Broker + plugin connection info [--peek [--json]] — --peek reads only the /tmp broker '
-      + 'advertisement plus one short broker query; it NEVER spawns a broker. --json is accepted '
-      + 'for compatibility but is a no-op: output is always the same single JSON object.',
+      'Broker + plugin connection info [--peek [--json]] [--wait [--timeout N]] — --peek reads '
+      + 'only the /tmp broker advertisement plus one short broker query; it NEVER spawns a broker. '
+      + '--json is accepted for compatibility but is a no-op: output is always the same single JSON '
+      + 'object. --wait blocks (MAY spawn a broker) until a matching plugin registers, printing a '
+      + 'figma:// deep link (when one can be built) to stderr immediately; --timeout N is in '
+      + 'SECONDS, default 60 — on timeout this exits non-zero with E_NO_PLUGIN instead of the '
+      + 'normal payload.',
   },
   { name: 'seat', description: 'Probe seat → {seat, bridge, reason} [--seat free|paid skips the probe]' },
   {
@@ -197,5 +201,14 @@ export const GLOBAL_FLAGS: GlobalFlagEntry[] = [
       'declare that this command only READS. Skips the per-file mutation queue (backlog 1.1+2.6+4.3). '
       + 'TRUSTED, NOT ENFORCED — the plugin sandbox cannot verify it, so a mis-declared mutation can '
       + "interleave with another agent's work.",
+  },
+  {
+    flag: '--agent <id> | FIGMA_AGENT_ID',
+    description:
+      'which harness sent this (the flag wins over the env var) — shown in the panel activity feed '
+      + '("claude · Created frame ..."). Trimmed, control characters stripped, capped at 32 chars. '
+      + "Omitted entirely when unset — an unlabelled request's wire frame is byte-identical to a "
+      + "pre-flag CLI's; the panel's own \"cli\" default is applied at render time, never stamped "
+      + 'onto the wire.',
   },
 ];
--- a/cli/src/commands/status.ts
+++ b/cli/src/commands/status.ts
@@ -10,7 +10,41 @@ import type { CommandArgs } from '../figma-agent.ts';
 import { fetchBrokerHello, runCommand } from '../transport/broker-client.ts';
 import { ensureBroker } from '../transport/broker-discovery.ts';
 import { peek } from '../transport/broker-peek.ts';
+import { waitForPlugin } from '../transport/plugin-wait.ts';
+import { buildDeepLink, type DeepLinkEntry } from '../transport/figma-deep-link.ts';
+import { fileIdentity, readBindCache, readBindMarker } from '../transport/project-bind.ts';
 import { CliError } from '../transport/protocol-helpers.ts';
+
+// auto-connect slice 2 — `--wait`'s own default, in SECONDS (not ms — every other
+// `--timeout` in this CLI is ms; this one follows the usecases.md gherkin's own
+// "--timeout 5" ⇒ "5s budget" wording, so `--help`/the skill say so explicitly).
+const DEFAULT_WAIT_SECONDS = 60;
+
+/**
+ * Find the bind-marker entry that `status --wait` should build a deep link from:
+ * an exact fileNameSlug match for `--file` when given, else the single most-
+ * recently-bound entry across every known project (an unambiguous best guess is
+ * more useful than no link at all when the caller didn't say which file). Reads
+ * the restart-survival cache + each project's own durable marker — the same
+ * source `figma-agent bind --list` reads.
+ */
+function findBoundEntry(wantedFile: string | undefined): DeepLinkEntry | null {
+  const cache = readBindCache();
+  const wantedSlug = wantedFile ? fileIdentity(null, wantedFile) : null;
+  let best: { fileKey: string | null; boundAt: number } | null = null;
+  for (const projectDir of cache.projectDirs) {
+    const marker = readBindMarker(projectDir);
+    if (!marker) continue;
+    for (const entry of marker.bindings) {
+      if (wantedSlug !== null) {
+        if (entry.fileNameSlug === wantedSlug) return { fileKey: entry.fileKey };
+        continue;
+      }
+      if (!best || entry.boundAt > best.boundAt) best = { fileKey: entry.fileKey, boundAt: entry.boundAt };
+    }
+  }
+  return wantedSlug !== null ? null : (best ? { fileKey: best.fileKey } : null);
+}
 // Concurrency & jobs (backlog 1.1+2.6+4.3), phase 02 §3 — each row now carries
 // `runningJob`/`queueDepth` (broker-status.ts's `buildBrokerHelloData`, given a
 // `jobStatusFor`). This CLI is JSON-only (no `--json` flag exists — figma-agent always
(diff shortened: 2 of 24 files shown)


jangtrinh (commented):
**Verdict: approve-equivalent — no blockers.** (Posted as a comment: GitHub does not allow an approve state on a pull request opened by the same account.) The riskiest thing in this slice, a 9th positional on the request-frame builder, is genuinely additive — I constructed the bytes both ways rather than reading the diff. Six non-blocking findings below, one of which produces a fabricated deep link from a corrupt marker.

Gates re-run in a fresh detached worktree at `07029f6` (`npm ci`, submodule initialised):

```
typecheck      EXIT=0
lint:comments  EXIT=0
build          EXIT=0      (git status --porcelain empty afterwards — no rebuild drift)
npm test       EXIT=1  — 1 failed | 1507 passed (1508)
```

The one failure is issue #59, unchanged and unrelated:

```
FAIL tests/broker-daemon-harness.test.ts > daemon harness — target pin (#35 P2)
     > SET_TARGET refuses a claimed instanceId that does not match the sender's OWN registered instance
```

Not counted. Panel typography gate re-run on its own: `Tests 43 passed (43)`.

---

## Envelope additivity — verified at the byte level

The 9th positional is the shape that worries me most, so I built both frames and compared bytes rather than trusting the test:

```
PRE : {"id":"c_1","cmd":"SET_TEXT","params":{...},"v":1,"activity":"Set text","expectedFile":"My File","expectedInstance":"p_1","projectDir":"/proj","readOnly":true}
POST: {"id":"c_1","cmd":"SET_TEXT","params":{...},"v":1,"activity":"Set text","expectedFile":"My File","expectedInstance":"p_1","projectDir":"/proj","readOnly":true}
BYTE-IDENTICAL: true
KEY ORDER pre : id,cmd,params,v,activity,expectedFile,expectedInstance,projectDir,readOnly
KEY ORDER post: id,cmd,params,v,activity,expectedFile,expectedInstance,projectDir,readOnly
hasOwnProperty("agent") when unset: false
MINIMAL      : {"id":"c_1","cmd":"STATUS","params":{},"v":1}
```

`PRE` is the 8-argument call, `POST` the 9-argument call with `agent` explicitly `undefined`. The undefined-slot hazard the positional shape suggests does not exist here because the wire form is a JSON **object** assembled by conditional assignment (`shared/protocol.ts:449`), not an array — an unset optional is an absent key, not a hole. Key order is untouched when unset; when set, `agent` lands between `projectDir` and `readOnly`, which nothing depends on.

Compatibility both directions holds by the same property: an old plugin receives one extra object key it never reads, and an old CLI's frame simply omits it, with the `cli` default applied at render time in `panel-ui.ts:170` rather than stamped on the wire.

Audited every `makeRequestFrame` call site for mis-addressing after the append. All existing callers pass 8 or fewer positionals, including the harness's `makeRequestFrame(reqId, 'SET_TEXT', {...}, undefined, undefined, undefined, undefined, 'inst-gone')` at `tests/broker-daemon-harness.test.ts:802` — still landing on `expectedInstance`. Appending rather than inserting was the right call, and the header comment saying the next field should be an options object instead of a 10th slot is the correct read.

## The NUL incident — clean

Byte-scanned every `.ts/.js/.html/.json/.md/.mjs` file in the tree (excluding `.git`, dependencies, the submodule) for any byte below 0x20 that is not tab/LF/CR, plus DEL: **0 files**. Nothing survived the incident.

`stripControlChars` (`cli/src/figma-agent.ts:110`) does what its comment claims. Ran the label through `resolveAgent`:

```
NUL                    -> len=6  cps=[63 6c 61 75 64 65]      ("claude", NUL gone)
ESC (C0 27)            -> len=8  (escape byte gone)
DEL (127)              -> len=2
newline (10)           -> len=2
```

See finding 1 for what it does *not* strip.

---

## Non-blocking

**1. `cli/src/figma-agent.ts:110-117` — "control characters stripped" is broader than what the filter does, on a surface where the gap matters.** The filter drops C0 and DEL. C1 controls and Unicode format/bidi characters pass through:

```
C1 NEL (U+0085)        -> len=3  cps=[61 85 62]
C1 CSI (U+009B)        -> len=3  cps=[61 9b 62]
ZWSP (U+200B)          -> len=3  cps=[61 200b 62]
RLO (U+202E)           -> len=3  cps=[61 202e 62]
LRI/PDI (U+2066/9)     -> len=5  cps=[61 2066 62 2069 63]
```

The panel writes this with `textContent`, so there is no injection here — but this label's whole job is attribution ("who touched my canvas"), and `U+202E` reverses the rendering direction of everything after it, so one agent's label can be made to read as another's in the activity feed. `U+200B` allows two visually identical labels that are different strings. Worth restricting to something closed rather than blocklisting: the label is documented as an id, so `code >= 32 && code !== 127 && !(code >= 0x80 && code <= 0x9f)` plus a strip of `Cf`-category characters, or simply a conservative allowlist.

This wording is now user-facing in `cli/src/command-catalog.ts:209` and in the emitted `SKILL.md`, so the doc and the filter should be reconciled in the same edit.

**2. `cli/src/figma-agent.ts:138` — the 32-char cap can split a surrogate pair.** `stripControlChars` iterates by code point (`for (const ch of value)`), which is careful; `clean.slice(0, AGENT_MAX_LEN)` then truncates by UTF-16 code unit, undoing it:

```
emoji at 32-char cut   -> len=32 loneSurrogate=true cps=[78 ... 78 d83d]
```

The label ends in a lone high surrogate — a malformed string that renders as a replacement character and survives JSON round-tripping as `\ud83d`. `[...clean].slice(0, AGENT_MAX_LEN).join('')` caps by the same unit the filter counts in.

**3. `cli/src/transport/figma-deep-link.ts:30` — a marker entry whose `fileKey` is absent or empty produces exactly the fabricated link this module promises never to produce.** The header says *"Never fabricates a link — a URL that opens nothing is worse than admitting there isn't one"*, and the type `DeepLinkEntry { fileKey: string | null }` does match `BindMarkerEntry` (`project-bind.ts:100`) exactly, so the drift resolution was right. But the value arrives from disk, and `readBindMarker` (`project-bind.ts:119-131`) validates only that `bindings` is an array — nothing checks the entries. Reproduced against the real function:

```
absent fileKey  -> {"url":"figma://file/undefined","reason":null}
empty fileKey   -> {"url":"figma://file/","reason":null}
null fileKey    -> {"url":null,"reason":"file has no fileKey (Free plan) — open the file manually"}
good fileKey    -> {"url":"figma://file/ABC123","reason":null}
```

An `=== null` check is a type-level guard on a value that crossed a trust boundary (a JSON file a user can hand-edit, or an older marker format). `typeof entry.fileKey !== 'string' || entry.fileKey === ''` restores the stated contract for both rows.

**4. `cli/src/commands/status.ts:36-46` — `--file` matches loosely for routing but exactly for the deep link.** `findBoundEntry` requires `entry.fileNameSlug === fileIdentity(null, wantedFile)`, while the plugin filter and `waitForPlugin` both use `fileMatches(..., true)` (substring, case-insensitive). So `status --wait --file "my"` can wait successfully on a plugin whose file is `My File` while reporting no deep link for it. The `null`-plus-reason output is honest, so nothing lies — but the two halves of one command answer the same flag differently, which will read as a bug from the outside.

**5. `plugin/src/main/main.ts:79` — `setRelaunchData` fires on every plugin open, not on bind.** The PR body describes it accurately ("at plugin startup"), so this is a scope question rather than a misstatement: the slice was framed as a relaunch button *on bound files*, and as written every file the plugin is ever opened on gets a document write, bound or not. Worth an explicit ruling before this lands, since the dirty-document and undo-history questions are exactly the unverified ones.

Related, and the reason I would not leave this unwatched: the `try/catch` makes a permanent silent no-op indistinguishable from success. If the call is refused (a viewer-permission file, a FigJam document, or a `documentAccess: "dynamic-page"` restriction — the repo has prior scars from sync getters behaving differently under that mode), the only trace is a `console.warn` in the plugin console. Nothing reaches `status`, the panel, or the CLI. Consider recording the refusal on the same present-only-when-relevant channel `bootSkipped` already uses a few lines below, so the live test has something to read.

**6. Coverage boundaries — accurate in the PR body, restating them so they're on the record.**

- The agent label's *render* path has no test: `grep -rn "log-agent\|activityRow" tests/` returns nothing. `tests/activity-feed.test.ts` covers `toActivityRecord` (including the deliberate no-fabricated-default), but the `r.agent ?? 'cli'` at `panel-ui.ts:170` — the actual two-agent acceptance criterion — is exercised nowhere. Correctly listed under "Not verified".
- `tests/status-wait.test.ts` calls `run()` from `cli/src/commands/status.ts` directly, which is the same function `COMMAND_MODULES` dispatches to, so the UC-05 substitution is a real path and not a reimplementation. It stops short of argv → `process.exit`, so "exits non-zero on timeout" is proven as a thrown `CliError('E_NO_PLUGIN')` rather than as an observed exit code; `tests/status-peek.test.ts:179` has the subprocess seam that would close that if the live-broker constraint ever lifts.
- The mock is narrow and honest: only `ensureBroker` is replaced, and it resolves to a **real** scratch broker daemon, so the WebSocket, `PLUGIN_HELLO`, the registry, `waitForPlugin`'s polling and the deep-link lookup all run for real. One small leniency: `fakeArgs.num` returns `Number(value)` while the real `arg-parse.ts:47` throws `E_INVALID_ARGS` on `NaN`. No test depends on that difference — I checked whether a non-numeric `--timeout` could reach `waitForPlugin` and produce a `NaN` deadline that never expires, and it cannot; the real parser rejects it first.

---

## Also checked and clear

- Deep link on stderr, JSON payload on stdout — consistent with the CLI's one-JSON-object-on-stdout convention, and printing it *before* the wait rather than only on failure is the right call.
- `waitForPlugin` always polls at least once before testing the deadline, so a zero or negative `--timeout` degrades to a single check rather than skipping the query entirely; a failed poll mid-wait is retried instead of being treated as a wait failure.
- The `--timeout`-in-seconds divergence from every other `--timeout` in this binary is called out explicitly in `command-catalog.ts:36-38`, so `--help` and the skill both carry it.
- `U+00B7` for the separator, not the `U+2022` the typography gate bans; `.log-agent` is present in the rebuilt `plugin/ui.html`, and the tree was clean after `npm run build`.

jangtrinh:
Fix round pushed as `1b92166`.

- **Bound-gated relaunch button.** `setRelaunchData` now fires only on a file the CLI has actually bound. The plugin's main thread can't read the bind marker itself (no filesystem access), so it waits for the broker's own answer — `SYNC_CONFIG` now also carries `bound`, computed from the same bind-index lookup already done for the idle window. A refusal or a deliberate skip lands on `bootSkipped` instead of a bare `console.warn`.
- **`--agent` is now an allowlist** (`^[a-z0-9._-]{1,32}$`) instead of a strip-and-truncate filter — anything outside it refuses with `E_INVALID_ARGS` naming the allowed set. Closes the C1/bidi-override passthrough and the surrogate-splitting truncation in one change. Documented in the catalog text `--help` reads.
- **`figma-deep-link.ts` no longer trusts a bare `=== null` check** on a value that crossed a trust boundary — a missing/empty `fileKey` (from a hand-edited or corrupted bind marker) now gets the same honest "no link" answer as the documented null case, never `figma://file/undefined` or `figma://file/`. `readBindMarker` validates each entry's own shape, not just that `bindings` is an array.
- **`status --wait`'s two halves (plugin match, deep-link lookup) now share one file-identity normalization**, so they can't disagree about which file `--file` meant.

Every behavioral change has a red-first test. Gates: `typecheck`/`lint:comments`/`build` all exit 0; slice-2 + touched test files green in isolation (143 tests); full `npm test` — one run, 117/118 files, 1515/1516 tests, the one failure is the pre-existing issue #59 flake in the same `target pin (#35 P2)` block, unrelated to this change.

Issue #54, opened 2026-08-07
Fresh nodes from typed creating commands log their own create-time writes as designer corrections (poisons the supervised-memory loop)


jangtrinh:
A node created by a typed command has its **own create-time property writes** recorded as a
`designer-correction` with `source: "designer"` and `unresolved: true`. No human touches the file.
`sync-corrections` then promotes those events into a project's canonical
`design/memory/figma-corrections.jsonl`, so the supervised-memory loop learns from the agent's own
writes as if a designer had corrected them.

Reproduced on a live canvas (`VSF - PCP`) on 2026-08-07 at `30a5b1f`.

## Repro

```sh
figma-agent create-frame --name "SPIKE" --w 120 --h 60
# then read the root's shared plugin data under the `ease_design` namespace
```

Two events land 102 ms apart for the same fresh node:

```jsonc
{ "eventId": "agent-…-10847-201384",      "source": "agent",
  "kind": "agent-operation",   "traits": { "command": "CREATE_FRAME" } }   // correct

{ "eventId": "correction-…-10847-201384", "source": "designer",
  "kind": "designer-correction", "unresolved": true,
  "causalParent": "agent-…-10847-201384",
  "traits": { "changeType": "PROPERTY_CHANGE",
              "properties": ["name", "width", "height"] } }               // false
```

`name`, `width` and `height` are exactly what the executor itself set.

## Mechanism

`main.ts` computes `mutationTargetIds(cmd, params)` and calls `beginAgentMutation()` **before**
dispatch. That arms the 2s suppression window (`correction-edge-store.ts`) from ids that must be
known in advance — and a node being created has no id yet, so the window is armed with an empty
set.

After dispatch, `resultMutationIds(cmd, result)` picks the new node's `result.id` up and
`recordAgentMutation()` writes the agent-operation event. That event is the causal parent
`recordDesignerCorrection()` requires, so the `documentchange` batch for the create — which is
delivered asynchronously, after the dispatch has returned — now passes every check and is written
as a designer correction.

The two halves are asymmetric: `resultMutationIds` feeds provenance but never feeds suppression.

Control: the same work through `EXEC_JS` produces **nothing**, because `EXEC_JS` is in neither
`mutationTargetIds` nor `resultMutationIds`, so no causal parent exists to hang a correction on.
That asymmetry is the tell.

## Blast radius

Every command in the `resultMutationIds` creating list: `CREATE_FRAME`, `CREATE_INSTANCE`,
[...]

PR #63, opened 2026-08-12
feat(cli): cowork — wait for one designer change-cycle

jangtrinh:
## Summary

Third and final slice of auto-connect (#56): an agent that has finished a task can wait for the designer to make one round of changes and stop, and gets back exactly what they touched plus any of the agent's own corrections still pending on those nodes — without polling and without a second memory store.

- **`figma-agent cowork [--wait S] [--timeout S]`** — waits for quiescence of `--wait` seconds (default 3, floor 1) after the designer edits something, or the overall `--timeout` (default 600s, in seconds like `status --wait`) with nothing recorded — the latter is a normal answer (`cycles: 0`), not an error.
- **Zero plugin change.** Detection is entirely broker-side: the broker already receives every edit, actor-labelled, on the existing `EDIT_FEED` stream (owner-edit change feed, wave 4.4) — `cowork` just watches it. Only `actor: owner` edits on a `source: live` batch arm the wait; the agent's own writes never do (the same class of bug issue #54 fixed for the correction ledger), a `gapfill` replay of edits made while the plugin was closed never does, and `ambiguous`-actor edits are counted (`ambiguousCount`) but don't arm either.
- **Broker-terminal**, the same shape as `PROJECT_BIND` and `JOB` — never forwarded to a plugin, so a long wait can't be culled by the per-file mutation queue or mistaken for a stuck job.
- **On a fired cycle**, prints the edited nodes (reusing the existing `editSentence` renderer — no second one) plus any pending corrections still open on those nodes, read with one `GET_CORRECTION_MEMORY` round trip against the existing supervised-memory ledger. Read-only: never writes the ledger, never a second store.
- **The watched plugin disconnecting mid-wait** replies immediately with `E_NO_PLUGIN` and a reconnect hint, instead of hanging to the full timeout. A plugin that drops between quiescence and the correction-ledger read reports `corrections: null` + a named reason — never an empty array standing in for "couldn't check."

## Decisions

- **§D2 resolved broker-side** (per the coordinator's ruling) — no plugin-side quiescence detector, no new wire event.
- `--wait`/`--timeout` are in seconds, matching `status --wait`'s own precedent (every other `--timeout` in this CLI is milliseconds) — documented explicitly in `--help`/the skill.
- No plugin matching the routing filter at registration time → immediate `E_NO_PLUGIN`, rather than parking and waiting for one to connect (a design choice: cowork has nothing to wait *on* without an already-open file, unlike an ordinary command which can reasonably wait for a just-spawned broker's plugin reconnect).

## Test plan

- [x] Red-first: `tests/cowork-waiter.test.ts` (pure, injected clock — no real 3s sleeps) and `tests/cowork-harness.test.ts` (real in-process broker + fake plugin WS) both fail against pre-change code (module not found / COWORK unhandled) before implementing.
- [x] `npm run typecheck` / `npm run lint:comments` / `npm run build` — all exit 0
- [x] `tests/cowork-waiter.test.ts` — 13/13 (arm/re-arm/ignore-agent/ignore-gapfill/fire/expire/partial-activity-still-fires)
- [x] `tests/cowork-harness.test.ts` — 7/7 (fires on a real cycle, coalesces multiple batches into one cycle, zero-edit timeout exits cleanly, agent-only traffic never fires, gapfill never fires, plugin-disconnect refuses fast, no-plugin-at-all refuses fast)
- [x] `npm run emit:skill` — drift test green (catalog gained the `cowork` command)
- [x] Full `npm test` — **2 consecutive clean runs**, 120/120 files, 1536/1536 tests, zero flake (confirms #59's fix holds under this stack)

## Not verified (owner live-test item)

A real designer editing a real canvas — does a wait actually return exactly the node(s) touched, and only those. Everything else (the four documented exits, the actor/source filtering, the correction-ledger read) is proven against a real in-process broker + real wire frames, just not a real Figma document.

Related: #56, #54 (correction-ledger actor semantics this reuses)


jangtrinh (commented):
**Verdict: request changes — one blocker.** (Posted as a comment: GitHub does not allow a request-changes state on a pull request opened by the same account.) The cowork design is the strongest work in this stack — actor filtering, waiter cleanup and the honest-unknown contract are all correct and verified below. One declared-green gate is red, and it is the same defect class already fixed once in the slice-1 round.

Gates in a fresh detached worktree at `202ffad` (`npm ci`, submodule initialised):

```
typecheck      EXIT=0
lint:comments  EXIT=1      <-- BLOCKER
build          EXIT=0      (git status --porcelain empty afterwards — no rebuild drift)
npm test  x3   EXIT=0, 0, 0 — 120 files / 1536 tests passed on all three runs
```

Three clean full-suite runs, no failures. Note the caveat in my review of the base PR: the target-pin block still needs retries to get there, and those retries are invisible in this reporter. That is a finding on #62, not here.

---

## BLOCKER — `npm run lint:comments` fails on this branch

The test plan checks this off as clean. It is not:

```
$ npm run lint:comments
comment-hygiene: 3 citation(s) not covered by the baseline.

State the invariant directly instead of citing an issue/PR number, a dated plan-dir
slug, or a bare finding/audit label:

  cli/src/transport/cowork-waiter.ts:31: [issue/PR number] * - `actor: 'agent'` edits never arm the timer — the exact class of bug issue #54 fixed
  tests/cowork-harness.test.ts:185: [issue/PR number] describe('cowork — agent-only traffic never fires a cycle (issue #54 class of bug)', () => {
  tests/cowork-waiter.test.ts:44: [issue/PR number] it('an agent-only batch never arms (issue #54 class of bug: the agent must never trigger its own cycle)', () => {
EXIT=1
```

Clean at the base commit `d8f0bbf`, so it is introduced here:

```
comment-hygiene: clean (55 file(s) with baselined citations, 0 new).
EXIT_AT_62=0
```

This is the same failure the slice-1 round already fixed once (a dead plan-path citation in `install-hook.ts`), now as issue numbers in one comment and two test names. The parenthetical adds nothing the surrounding sentence does not already say — `cowork-waiter.ts:31` already states the invariant completely ("an agent's own writes must never look like the designer stopping to think"). Two of the three are test *names*, where the rule matters most: a test name should describe the behavior, not cite where the bug was filed.

## Non-blocking

**1. `cli/src/commands/cowork.ts:90` — a bare `catch` reports one hardcoded cause for every possible failure.**

```ts
} catch {
  return { ...base, corrections: null, correctionsReason: 'plugin disconnected before the ledger could be read' };
}
```

The `corrections: null` half is exactly right — an honest unknown, never `[]`. The reason is asserted rather than observed: a wire timeout, an `E_WRONG_FILE`, or an older plugin build that does not implement `GET_CORRECTION_MEMORY` all report a confident "plugin disconnected" that never happened. This is the same shape as the `broker-peek.ts` refusal-relabelled-as-timeout finding from slice 1, and it was fixed there by carrying the real error through. Catch the error, and derive the reason from `err instanceof CliError ? err.code/err.message : ...`.

**2. `broker-daemon.ts:1329` — the caller-supplied `timeoutMs` has no ceiling.** `typeof params?.timeoutMs === 'number' && params.timeoutMs > 0` accepts any positive value, so a CLI can register a waiter that effectively never expires. The waiter is still cleaned when its socket closes (`handleClose`, and the `readyState !== OPEN` check in the sweep), so this is not a leak — but it is inconsistent with the repo's own precedent for caller-supplied budgets: `EXEC_JS` caps at `EXEC_JS_MAX_TIMEOUT_MS` (120s) rather than trusting the flag. A `Math.min(..., MAX)` with a named constant would match it.

**3. Quiescence resolves at park-sweep granularity — up to one tick late.** `PARK_SWEEP_INTERVAL_MS` is `min(500, max(100, PLUGIN_WAIT_TIMEOUT_MS/8))`, so the quiet window can overshoot by up to ~500ms. Against a `--wait` measured in seconds (floor 1s, default 3s) that is at most a few percent and piggybacking on the existing interval is the right call — worth stating in the module header so the resolution is a documented property rather than an accident of which interval it was attached to.

**4. The corrections round-trip re-resolves routing independently of the wait.** `handleCowork` deliberately pins one `targetInstanceId` for the whole wait and never re-resolves it (correct, and well argued in the comment). The follow-up `runCommand('GET_CORRECTION_MEMORY', {})` at `cowork.ts:84` is a separate request that routes through the ambient filter again. With an explicit `--file` both agree; if the wait resolved via the standing runtime pin and that pin moves during the wait, the edits and the corrections can describe two different files. Passing the resolved instance back on the COWORK reply and addressing the ledger read to it would close that.

## Verified

- **Queue bypass.** `broker-daemon.ts:1513` intercepts `COWORK` in the same `else if` chain as `PROJECT_BIND`/`JOB`, before `admitRequest`, so a wait never enters the per-file FIFO, never takes the contention lock, and cannot block another command. Correct precedent, correctly followed.
- **No zombie waiters, both directions.** Plugin disconnect (`:1421-1431`) replies `E_NO_PLUGIN` to every waiter pinned to that instance with an actionable reconnect hint, then drops them. CLI disconnect (`:1455`) filters by `from !== ws` silently, which is right — there is nobody left to answer. The sweep also skips any waiter whose socket is no longer `OPEN`.
- **Concurrent waiters from two CLIs both fire.** `st.coworkWaiters` is an array; the `EDIT_FEED` handler feeds every waiter whose `fileIdentity` matches, and the sweep replies to each independently. Neither is dropped.
- **Actor semantics are exactly the owner ruling.** `cowork-waiter.ts:36-45`: `source !== 'live'` returns early (gapfill replay never arms), only `actor === 'owner'` edits arm or re-arm, `actor === 'agent'` never arms, and `ambiguous` is counted into `ambiguousCount` without arming — a refusal to guess kept distinct from a confirmed owner edit, and surfaced only when non-zero.
- **Zero-edit timeout is a success.** `tick` returns `'expire'` with no edits, which replies through `sendReplyOk` with `cycles: 0` — exit 0, not an error. Partial activity past the deadline still fires rather than being discarded.
- **No second memory store.** Corrections come from one read-only `GET_CORRECTION_MEMORY` round-trip against the existing ledger; nothing writes it. The `causalParent` walk in `isPendingForCycle` is cycle-guarded with a `seen` set.
- **Protocol additions are additive and the classification is honest.** `COMMANDS` gains `COWORK`, `COMMAND_TIMEOUTS` gains a fallback entry rather than raising `DEFAULT_TIMEOUT_MS` for everyone. `tests/mutating-commands.test.ts:26` lists `COWORK` as read-only, which is correct on that file's own axis (nothing to undo — it never touches the scene) and is explained against the `PROJECT_BIND`/`JOB` precedent.
- **Plugin bundles rebuilt from `shared/protocol.ts` only** — no `plugin/src` changes in the diff, and `npm run build` left the tree clean, so `plugin/code.js` and `plugin/ui.html` match what the sources compile to.
- Module sizes are fine (`cowork.ts` 93, `cowork-waiter.ts` 63). `SKILL.md` re-emitted; the drift test is green in all three runs.

jangtrinh:
Reworked per review, rebased onto the reworked #62.

- **`lint:comments` blocker fixed** — reworded the 3 comment/test-name citations of "issue #54" (`cowork-waiter.ts:31`, `tests/cowork-harness.test.ts:185`, `tests/cowork-waiter.test.ts:44`) to state the invariant directly instead.
- **`cowork.ts`'s bare `catch` fixed** — it collapsed every ledger-read failure (wire timeout, `E_WRONG_FILE`, an older plugin build lacking `GET_CORRECTION_MEMORY`, a dropped broker) into one hardcoded reason. Now derives `correctionsReason` from the actual thrown error (`CliError.code`/`.message`, or a generic message for anything else), keeping the honest `corrections: null` on any failure.
- **`--timeout` capped** — `handleCowork` accepted any positive value with no ceiling. A cowork waiter holds a live broker-side slot for its whole duration (unlike a one-shot `EXEC_JS`), so an unbounded value is a standing resource hold, not just a UX nuisance. Added `COWORK_MAX_TIMEOUT_MS` (30min) in `shared/protocol.ts`, same "CLI --timeout may raise, capped" shape as `EXEC_JS_MAX_TIMEOUT_MS`.
- **Rebased onto the reworked `fix/harness-target-pin-flake`** — picks up the recency tie-break fix, so this branch no longer carries the flake either.

**Verified on the rebased tip**: `typecheck` clean, `lint:comments` run explicitly and clean (0 new citations), `build` clean, `npm test` x2 → 1536/1536 both times, `emit:skill` re-run — no diff (command surface unchanged, so `SKILL.md` didn't need updating).

jangtrinh (commented):
**Follow-up on `b745c54` — verdict: approve-equivalent. Blocker cleared.** (Comment rather than an approve state: same-account limitation.)

Gates re-run in a fresh detached worktree at `b745c54` (`npm ci`, submodule initialised):

```
typecheck      EXIT=0
lint:comments  EXIT=0      <-- was 1
build          EXIT=0      (git status --porcelain empty afterwards — no rebuild drift)
npm test  x2   EXIT=0, 0 — 120 files / 1536 tests passed both runs, ~10s each
```

`comment-hygiene: clean (55 file(s) with baselined citations, 0 new)`. All three citations are gone, and the replacements state the invariant instead of pointing at a tracker — `cowork-waiter.ts:31` now explains *why* an agent's own writes must not arm the timer, and both test names describe the behavior. That is the better outcome than a baseline entry.

Now that the base PR's registry fix has landed underneath it, the suite is genuinely clean here — no retries absorbing anything, which is what made the previous round's green untrustworthy.

## Delta spot-checks

**`correctionsReason` is derived, not asserted.** `cowork.ts:91-99`:

```ts
const correctionsReason = err instanceof CliError
  ? `${err.code}: ${err.message}`
  : `unexpected error reading corrections: ${err instanceof Error ? err.message : String(err)}`;
```

`E_NO_PLUGIN`, `E_NO_BROKER` and a wire timeout now each produce their own reason carrying the real code and message, and a non-`CliError` throw is labelled distinctly rather than being folded into the same sentence. This is no longer the relabelling defect — it is the same shape the `broker-peek.ts` fix settled on. `corrections: null` still stands for unknown, never `[]`.

**The timeout cap is enforced broker-side.** `broker-daemon.ts:1346-1350` applies `Math.min(requestedTimeoutMs, COWORK_MAX_TIMEOUT_MS)` inside `handleCowork`, which is the right place — `params` arrives from a client the broker does not trust, so a CLI-side clamp alone would be advisory. `COWORK_MAX_TIMEOUT_MS = 1_800_000` clears the 600s default with real headroom, and the constant's comment correctly distinguishes this from `EXEC_JS_MAX_TIMEOUT_MS` (a standing waiter slot versus a one-shot call).

Checked the interaction with the CLI's own wire budget: `cowork.ts` computes `wireTimeoutMs` from the *uncapped* request, so the wire always outlives the broker's capped deadline. That is the safe direction — the broker replies first and the caller never hangs.

## One non-blocking note

The cap applies silently. A caller passing `--timeout 3600` gets a reply at 30 minutes with no field saying the ceiling was applied; `waitedMs` lets them infer it, but only if they think to compare. Given this repo's "nothing vanishes silently" rule, a `timeoutCappedMs` on the reply when `requestedTimeoutMs > COWORK_MAX_TIMEOUT_MS` would close it — same present-only-when-relevant contract as `ambiguousCount`. Not worth holding the merge.

Everything I verified in the previous round still holds: queue bypass before `admitRequest`, waiter cleanup on both plugin and CLI disconnect, concurrent waiters from two CLIs all fed, owner-only arming with agent edits never arming and ambiguous counted, `cycles: 0` as a success, read-only ledger passthrough, and `COWORK` classified read-only on the undo axis.

Issue #64, opened 2026-08-12
cowork: disclose when --timeout is capped (timeoutCappedMs)


jangtrinh:
Follow-up from #63's review: `cowork --timeout 3600` silently returns at the 30-minute `COWORK_MAX_TIMEOUT_MS` ceiling. Add a `timeoutCappedMs` field present only when the cap applies, matching the `ambiguousCount` disclosure contract (nothing vanishes silently).

jangtrinh:
Claimed for the final auto-connect completion pass. Scope: add `timeoutCappedMs` only when the broker clamps the requested cowork timeout, prove the pre-fix failure at the CLI/broker seam, run full gates, stage-4 review, and merge before closing #56.

jangtrinh:
Final evidence — shipped and audited.

- PR #66 merged as `0182b967`.
- Contract: requested cowork timeout above the 30-minute ceiling returns `timeoutCappedMs: 1800000`; at or below the ceiling the field is absent.
- Proven at both seams: real broker WebSocket reply and final cowork CLI JSON output. The red check failed pre-fix with `expected undefined to be 1800000`.
- Stage-4 Opus + Codex review: no blocker/high/medium.
- Exact integrated gates: typecheck/build/lint clean; 123 suites / 1547 tests passed at the PR head.
- Fable final audit verdict: PASS TO CLOSE.

The pre-existing oversized CLI wire deadline remains a non-blocking LOW; broker replies at the applied cap and socket closure already fails immediately.

PR #66, opened 2026-08-13
fix(cli): disclose capped cowork timeout

jangtrinh:
## Summary

- disclose `timeoutCappedMs: 1800000` when the broker clamps an oversized cowork timeout
- omit the field when the requested timeout is at or below the cap
- preserve the disclosure through the final CLI JSON output

## Verification

- `npx vitest run --config vitest.config.ts tests/cowork-command.test.ts tests/cowork-harness.test.ts tests/cowork-waiter.test.ts` — 24 passed
- `npm run typecheck` — passed
- `npm run build` — passed
- `npm test` — 123 files, 1547 tests passed
- `npm run lint:comments` — passed, 0 new citations

Tracks #64. Issue closure remains gated on the final auto-connect live audit.

### DIFF
--- a/cli/src/commands/cowork.ts
+++ b/cli/src/commands/cowork.ts
@@ -19,6 +19,7 @@ interface CoworkWireResult {
   file: string | null;
   edits: EditInput[];
   ambiguousCount?: number;
+  timeoutCappedMs?: number;
 }
 
 interface CoworkEditOutput {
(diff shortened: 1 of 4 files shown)


jangtrinh (commented):
Opus stage-4 review: NO BLOCKER / HIGH / MEDIUM. Verified strict > cap semantics, absent-at-boundary shape, broker seam, final CLI seam, and shared fire/expire reply composition. LOW, non-blocking and pre-existing: CLI wire deadline retains the requested timeout rather than the broker cap; broker replies at the cap and socket close already refuses immediately. No scope change requested.

jangtrinh (commented):
Codex stage-4 cross-check: NO BLOCKER / HIGH / MEDIUM. Re-read full diff against #64; confirmed timeoutCappedMs is created only from COWORK_MAX_TIMEOUT_MS after a strict cap comparison, survives both normal and corrections-read-failure output paths, and cannot be fabricated when absent. Reproduced focused 24/24 and orchestrator full gates: 123 suites / 1547 tests, typecheck/build/lint clean. Plugin generated artifacts unchanged.

PR #72, opened 2026-08-14
fix(cli): dispatch commands through executable symlinks

### LINKED ISSUE #71: fix(cli): dispatch commands through executable symlinks
## Outcome

Running the CLI through a symlink dispatches the requested command exactly like running the real bundled entrypoint.

## Symptom and root cause

The entrypoint guard compares `import.meta.url` with `pathToFileURL(process.argv[1]).href`. Node resolves the imported module to its real path while `process.argv[1]` may retain the npm/global symlink path, so the guard can evaluate false and silently skip `main()`.

## Constraints

- Preserve import-without-dispatch behavior used by tests.
- Change no command, transport, wire, or output contract.
- Keep the fix local to entrypoint identity.
- Test first against an actual filesystem symlink.

## Acceptance criteria

- A regression test fails on current `main` when the bundled CLI is invoked through a symlink.
- The same test passes after canonicalizing the executable path.
- Direct CLI invocation and module import behavior remain unchanged.
- Typecheck, comment hygiene, full tests, and build pass.

## Non-goals

Installer changes, package publication, version bump, broker behavior, or Figma plugin behavior.

jangtrinh:
## End-to-end work summary

The bundled CLI now recognizes its process entrypoint through a real filesystem symlink. The guard compares canonical paths on both sides and treats an unresolvable process entry path as a plain import, preserving import-without-dispatch behavior.

## Subagent delegation

- Implementation worker added the failing regression and initial minimal fix.
- Debug review identified `--preserve-symlinks-main` and missing `argv[1]` path edges.
- Production review verified the hardened implementation and regression coverage.
- Git manager committed and pushed the reviewed two-file scope.

## Technical decisions

- Canonicalize both `import.meta.url` and `process.argv[1]` with `realpathSync`.
- Keep canonicalization inside a non-throwing entrypoint predicate so programmatic imports remain safe.
- Exercise a real filesystem symlink, Node preserved-main-symlink mode, and a missing process entry path.

## Deviations from plan

The initial one-sided canonicalization was expanded after review showed it could throw during a plain import and could still skip dispatch with `--preserve-symlinks-main`. The accepted outcome and external contracts are unchanged.

## Completion evidence

- Test-first proof: the original regression failed on unmodified `main` with exit 0 and empty stdout through the symlink.
- Focused tests: 2 files, 7 tests passed.
- Full suite: 126 files, 1,619 tests passed.
- Typecheck: passed.
- Comment hygiene: clean, zero new findings.
- Build: passed.
- `git diff --check`: passed.
- Independent spec and production re-review: PASS, no findings.

## Checklist

- [x] Change is scoped to CLI entrypoint identity.
- [x] Direct invocation behavior is preserved.
- [x] Import-without-dispatch behavior is preserved.
- [x] No command, broker, transport, wire, or Figma plugin contract changed.
- [x] No installer, publication, or version changes.
- [x] Tests and build pass locally.

## Human actions required

None.

## Linked Issues

Closes #71

## Ship Mode

English fallback.

### DIFF
--- a/cli/src/figma-agent.ts
+++ b/cli/src/figma-agent.ts
@@ -8,7 +8,8 @@
 // this module (e.g. for COMMAND_MODULES) without accidentally dispatching a
 // command from the TEST RUNNER's own argv.
 import { resolve } from 'node:path';
-import { pathToFileURL } from 'node:url';
+import { realpathSync } from 'node:fs';
+import { fileURLToPath } from 'node:url';
 import { runBrokerDaemon } from './transport/broker-daemon.ts';
 import { parseArgs, type CommandArgs } from './arg-parse.ts';
 import { CliError } from './transport/protocol-helpers.ts';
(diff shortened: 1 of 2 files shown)


jangtrinh (commented):
Formal review: PASS.

Verified the accepted outcome and production behavior:

- canonical path comparison covers direct execution, filesystem symlinks, and `--preserve-symlinks-main`;
- resolution failures return non-entrypoint instead of throwing during import;
- focused regressions pass 7/7;
- full suite passes 126 files / 1,619 tests;
- typecheck, comment hygiene, build, and diff checks pass;
- scope is limited to the entrypoint guard and its regression tests.

No Critical, Important, or Suggestion findings.

PR #104, opened 2026-09-03
fix(plugin): move gap-fill baseline off the document and gate sandbox globals

jangtrinh:
## Why

Opening the plugin on a large file froze the canvas, and the freeze survived the earlier single-walk fix. Probing the owner's 21-page / 418k-node file live showed why: the Figma main-thread sandbox has no `TextEncoder`, the gap-fill snapshot writer used one, so every write threw on its first record and was swallowed. No baseline was ever persisted on any file, while every boot, idle window and plugin close still paid the full document walk. `tsc -p plugin` was green because its lib included DOM. Writing the snapshot into `figma.root` sharedPluginData would also have echoed back as `documentchange` and re-armed the idle timer (verified live), and would have put ~10 MB per cycle into the document.

## What

- `plugin/tsconfig.main.json` (ES2020 lib only) joins `npm run typecheck`: a DOM/Node-only global in `plugin/src/main` or `shared` now fails the gate (proven red on the old line, green after).
- One shared plain-JS utf8 byte counter (`shared/utf8-byte-length.ts`).
- Baseline lives in `figma.clientStorage`, one value per file, stamped with `fileKey`/`fileName`/`writtenAt`; records only for pages under the cap. A foreign or unreadable baseline is reported as missing and never diffed against; a failed read withholds the write so the previous baseline survives.
- Quota refusal evicts the oldest other file's baseline, retries once, then records the failure. Evictions and errors surface in STATUS `gapfill`.
- Chunked sharedPluginData writer, its manifest and the `figma.on('close')` full walk are deleted; the legacy manifest key is cleared once.
- A missing baseline emits exactly one `baseline-missing` feed notice.
- A page whose walk throws no longer aborts boot: its diff is skipped, its previous entry carried forward, and `documentchange` is still subscribed (`boot-capture.ts`).

## Evidence

- Gates after the last edit: typecheck 0 · bundle 0 · lint:comments 0 · test 143 files / 1852 passed.
- Mutation check: 6 targeted source mutations, 6 caught after the fix round.
- Live on the owner's file after reopening: STATUS `gapfill` = pagesDiffed 21 · pagesTruncated 15 · baselineWrittenAt set · 1.6 KB; root `ease_design` keys hold only connections + corrections; legacy manifest cleared; feed shows exactly one `baseline-missing` frame and no self-emitted DOCUMENT/pluginData frames from the boot.

## Not in this PR

Bounded/sliced walker and dirty-page idle (next PR), documentchange noise predicate and per-batch store read (next PR), relay probe/buffer, idle-write retry after a withheld write.


jangtrinh (commented):
## Summary

The four findings from the previous round are genuinely fixed and pinned by tests that fail without the fix; gates, the sandbox gate's teeth, and the stated test totals all reproduce locally. One honest-reporting defect remains: a storage read that *fails* is still announced to the feed as "there was no baseline".

**Risk level:** medium — no data loss and no contract break, but the feed states a false cause on a path the code already distinguishes internally, and that path also pays a full document walk for nothing.

### Verified locally (head 7cf97b7)

- `typecheck` 0 (all three projects) · `lint:comments` clean, 0 new · `npm test` 143 files / 1852 passed — matches the description.
- Sandbox gate has teeth: a temporary `new TextEncoder()` in `plugin/src/main` fails `tsc -p plugin/tsconfig.main.json` (TS2304) while `tsc -p plugin` stays green — i.e. exactly the hole that shipped the original bug. It also covers `plugin/src/ui/panel-model.ts` transitively (adding `typeof document` there fails the gate).
- Mutation reproduced: neutering the `readFailed` guard in `writeBaseline` (`edit-gapfill.ts:319-322`) turns two tests red, including the byte-identity assertion on the preserved baseline. The identity check, the per-page throw skip, and the injected `writtenAt` are each pinned by an assertion that only passes on the fixed behaviour.
- Boot order: `loadAllPages → gapfill → post EDIT_FEED → legacy clear → subscribe` (`main.ts:487-492`, `boot-capture.ts:22-43`). The only path that skips `documentchange` is a `loadAllPagesAsync` failure, which is the dynamic-page precondition — correct, and announced. A gap-fill rejection notifies and still subscribes.
- Idle path: single-flight writer is the sole idle writer, and the boot write completes before the subscription exists, so the two cannot interleave. A withheld write surfaces in STATUS `gapfill.errors`, and the next boot recovers once the read succeeds.
- Contracts: STATUS `gapfill` is additive and reaches the CLI verbatim (`cli/src/commands/status.ts:167`, `plugin` spreads the scene object). `editSentence` is the only renderer of feed frames (CLI-side only; the panel has no competing copy). No importer of the removed `writeSnapshot`/chunk exports remains.

## Findings

### Important

**1. A failed baseline read is reported to the user as "no previous baseline" — `plugin/src/main/edit-gapfill.ts:394`**

```ts
const { baseline: prev } = await readBaseline(store, stats);   // readFailed dropped here
if (!prev) { /* … full walk … */ return [baselineMissingNotice(figma.root.name, …)]; }
```

`readFileBaseline` distinguishes "nothing stored" from "the store refused the read" (`gapfill-baseline-store.ts:141-168`) precisely so the two are never conflated — and `writeBaseline` uses that distinction correctly. The diff path discards it.

Scenario: `clientStorage.getAsync` rejects transiently at boot on a file that has a valid baseline. The write is correctly withheld, so nothing is lost on disk — but the session posts a `baseline-missing` frame that renders as *"Gap-fill had no previous baseline for \"VSF - PCP\" — edits made before this session are unreported"*. The baseline exists and is intact; the stated cause is false, which is the failure mode the house rule ("a wrong fact is worse than an absent one") exists to prevent. The current behaviour is pinned by `tests/edit-gapfill-baseline.test.ts:421-433`, so it is a decision, not an oversight — but the STATUS error already carries the truth (`baseline read failed: …`), and the feed contradicts it.

Second cost on the same path: `!prev` sends boot into the first-run branch, which walks **every page** and then withholds the write — a full walk of the large document this change exists to stop walking, producing nothing.

Suggested fix: return `readFailed` from `readBaseline` in `runGapfillDiff`; on a read failure skip both the notice and the first-run walk (STATUS already reports it), or emit a distinct notice that names the read failure rather than an absent baseline. The existing test then needs its expectation flipped.

### Suggestion

**2. The baseline is read twice per boot — `edit-gapfill.ts:394` and `:314`.** `runGapfillDiff` reads it for the diff, then `writeBaseline` reads it again for the carry-forward map. On a multi-MB baseline that is a second `getAsync` plus a second parse on the main thread for information the caller already holds. Passing the already-read baseline (and its `readFailed` flag) into `writeBaseline` removes the duplicate and fixes finding 1 in the same edit.

**3. The idle write still walks the whole document in one task — `main.ts:323-325`.** `writeBaseline(figma.root.children, snapshotPage, …)` calls `resolveBaselinePage` in a tight loop with no `yieldToHost`, unlike the boot diff which yields between pages. On the 21-page file this is the same main-thread block the change is chasing; it is disclosed as deferred, but the asymmetry with the boot path deserves a comment at the call site so it does not read as an oversight to the next reader.

**4. Persisted-but-unread payload in a quota-constrained store — `gapfill-baseline-store.ts:18`, `:43`, `edit-gapfill.ts:268`, `:332`.** Every record carries a `parent` id that no code reads (`diffSnapshots` compares id/name/x/y; `toGapfillEdit` hardcodes `parentName: null`), roughly 12 bytes per node — about 48 KB for a full 4 000-node page. `writtenBy` is written and parsed but never compared to anything. Both are speculative, and the byte cost is not free here: a quota refusal deletes another file's baseline. Either use them or drop them until a consumer exists.

**5. Eviction deserializes every other baseline — `gapfill-baseline-store.ts:183-197`.** Only on the refusal path, but that is precisely when the main thread is already struggling: each candidate key is fetched and JSON-parsed in full just to read `writtenAt`. A tiny sidecar key holding `{key: writtenAt}` would make the ranking O(1) reads.

**6. Evicted keys expose other files' identifiers through STATUS — `executor-ops.ts:129`, `protocol.ts:379`.** `baselineEvicted` carries `figma-edit-baseline-v2:<fileKey>` for a *different* file, and STATUS flows into CLI output, agent transcripts and logs. A file key is a durable handle to a document the operator may not have meant to name here. Consider reporting the count plus the stored `fileName`, or a truncated key.

**7. Two shared main/UI modules are now implicitly DOM-free — `plugin/tsconfig.main.json:7`.** The include globs list `src/main/**` and `shared/**`, but `plugin/src/ui/panel-model.ts` and `ui/converter/color-utils.ts` also end up in the bundle and are only gated because main imports them (confirmed by probe). That is the right outcome; a one-line note in each of those two files would stop a future UI-side edit from tripping a gate whose name gives no hint.

**8. A slug collision replaces the other file's stored baseline, and the record does not say so — `gapfill-baseline-store.ts:161-166`.** When two keyless files slug to one key, the second file reports the first's value as foreign (correct, and recorded) and then writes over it. The message says "treated as missing"; it does not say the other file's baseline was replaced. One clause closes the ledger.

**9. `gapfill-baseline-store.ts` is 227 lines against the repo's 200-line guidance — no action recommended.** Key derivation, codec, identity check and the write/evict bracket are one contract, each seam already has a test, and splitting it would create a module boundary with no independent consumer.

## Verdict

**Request changes** — finding 1 only. Everything else is a suggestion and can land separately.

*Posted by the installed review-pr skill at 2026-09-03T03:36:40Z*

jangtrinh (commented):
## Summary

Both blocking findings are now closed at the root rather than in the wording: a refused boot read emits its own `baseline-unreadable` notice and marks the session, and every later write in that session is withheld and counted — so the sentence's promise ("reported on the next successful boot") is one the code actually keeps. Nothing left is blocking.

**Risk level:** low — no data loss, no contract break, gates verified independently, and every fix in this round is pinned by a test that is red without it.

### Re-verified at head 0e74b3d

- **The gate — `plugin/src/main/edit-gapfill.ts:331-334`, set at `:422`:**
  ```ts
  // writeBaseline, before any store traffic:
  if (stats.bootBaselineUnreadable) {
    recordGapfillError(stats, 'baseline write withheld: this session could not read the stored baseline at boot');
    return;
  }
  ```
  Neutering it turns the new test red on the stored-baseline identity assertion (`expected {…} to be {…} // Object.is equality`), matching the reported red. The test is honest about the scenario it claims: it clears `getError` between the boot diff and the write, so reads genuinely recover mid-session and the write is refused by the session flag, not by a still-failing store.
- **Nothing bypasses the gate.** `writeFileBaseline` has exactly one caller (`edit-gapfill.ts:358`, inside `writeBaseline`), and `figma.clientStorage.setAsync` is reached only through the store interface at `gapfill-baseline-store.ts:69`, which only `writeFileBaseline` uses. `writeBaseline` itself has three call sites: the idle trigger (`main.ts:324`) and two inside `runGapfillDiff` (`:434`, `:481`), both downstream of the `readFailed` early return and therefore unreachable in an unreadable-boot session. `main.ts` shares one `gapfillStats` between boot and idle, so the flag is visible to the idle path. The gate sits before the read, so such a session performs no store traffic at all.
- **Residual, stated for the record:** the flag is set inside `runGapfillDiff`, so a gap-fill that *throws* before reaching the read (rather than returning `readFailed`) would leave later writes ungated. Only `figma.root`/`figma.fileKey` access can throw there, and `writeBaseline` calls the same accessor at `:327`, so the same fault blocks the write too. No reachable hole found.
- **Gates:** `typecheck` 0 · `lint:comments` clean, 0 new · `npm test` 143 files / 1854 passed. `plugin/code.js` carries the new flag, so the committed bundle is current.
- **Downstream safety of the notices:** gap-fill frames never pass through `coalesceEdits` (`main.ts:460-470`; the only call is the live batch at `:434`), so `gapfill:baseline-unreadable` cannot merge with a real node — and Figma ids are `<num>:<num>`, so collision is impossible regardless. `editSentence` is the single renderer (`shared/edit-vocabulary.ts:100-107`), checked before the generic verb path, so a notice never reads as "Restyled". `cli/src/commands/changes.ts` is the only feed reader; it prints `nodeId` and never resolves it against the document. One caveat in finding 9.

## Findings

### Critical

None.

### Important

None — both prior findings are closed and test-pinned.

### Suggestion

**2. Baseline read twice per boot — `edit-gapfill.ts:417` and `:335`.** The diff reads it, then `writeBaseline` reads it again for the carry-forward map. On a multi-MB value that is a second `getAsync` plus parse for data the caller already holds.

**3. The idle write walks the whole document in one task — `main.ts:323-325`.** No `yieldToHost` between pages, unlike the boot diff. Same main-thread block on the 21-page file; disclosed as deferred, but the asymmetry deserves a comment at the call site so it does not read as an oversight.

**4. Persisted-but-unread payload in a quota-constrained store — `gapfill-baseline-store.ts:18`, `:43`; `edit-gapfill.ts:268`, `:353`.** Every record carries a `parent` id nothing reads (`diffSnapshots` compares id/name/x/y; `toGapfillEdit` hardcodes `parentName: null`) — ~12 bytes/node, ~48 KB for a full 4 000-node page. `writtenBy` is written and parsed, never compared. A quota refusal deletes another file's baseline, so these bytes are not free.

**5. Eviction deserializes every other baseline — `gapfill-baseline-store.ts:183-197`.** Only on the refusal path, i.e. exactly when the main thread is already struggling: each candidate is fetched and parsed in full to read one `writtenAt`. A small index key would make the ranking cheap.

**6. Evicted keys expose other files' identifiers — `executor-ops.ts:129`, `protocol.ts:379`.** `baselineEvicted` carries `figma-edit-baseline-v2:<fileKey>` for a *different* file and flows into CLI output and logs. A count plus the stored `fileName`, or a truncated key, says as much without the handle.

**7. Two shared main/UI modules are now implicitly DOM-free — `plugin/tsconfig.main.json:7`.** `plugin/src/ui/panel-model.ts` and `ui/converter/color-utils.ts` land in the bundle and are gated only because main imports them (verified: `typeof document` in panel-model fails the gate). Right outcome; a one-line note in each stops a future UI-side edit from tripping a gate whose name gives no hint.

**8. A slug collision replaces the other file's baseline without saying so — `gapfill-baseline-store.ts:161-166`.** The foreign value is reported as "treated as missing" and then overwritten; the record does not mention the replacement. One clause closes the ledger.

**9. Notices count as owner edits — `cli/src/commands/changes.ts:170-173`.** `countByActor` counts every frame, so a session whose only gap-fill output is a notice reports `counts.owner: 1` with no owner edit behind it. Pre-existing with the truncation notice, now three instances; excluding `gapfill:` ids, or counting them separately, keeps the one number consumers trust honest.

**10. `gapfill-baseline-store.ts` is 227 lines against the 200-line guidance — no action recommended.** Key derivation, codec, identity check and the write/evict bracket are one contract with a test per seam; splitting it would create a boundary with no independent consumer.

**11. A withheld-write session inflates `errorCount` — `edit-gapfill.ts:332`, `gapfill-status.ts:30-33`.** Every idle fire after an unreadable boot records another error, so a long session shows `errorCount: 12` from a single root cause while `errors[0]` still names the original read failure. Correct as a ledger, but an operator cannot separate "one cause, twelve withheld writes" from "twelve distinct failures". Surfacing the session flag in `GapfillStatus`, or counting withheld writes in their own field, would read straight.

## Verdict

**Approve** — findings 2-11 are non-blocking and can land as follow-ups.

*Posted by the installed review-pr skill at 2026-09-03T03:47:08Z*

PR #105, opened 2026-09-03
perf(plugin): filter self-emitted pluginData changes and read the correction store once per batch

jangtrinh:
## Why

Two costs sat inside every `documentchange` batch. First, the plugin's own bookkeeping writes on `figma.root` (verified live: `PROPERTY_CHANGE ['pluginData']` on the DOCUMENT node) entered the owner-edit feed as an "updated" edit and re-armed the idle timer — a self-feeding loop. Second, `recordDesignerCorrection` read and parsed the whole correction store once per changed node (0.63 ms at 102 events), so a 50-node drag cost a store parse storm per delivery. The handler also lived inside `main.ts`, which calls `figma.showUI` at module load and so could not be tested.

## What

- `onDocumentChange` is extracted into `plugin/src/main/document-change-capture.ts` with injected dependencies; `main.ts` delegates. Real designer edits keep the same actor classification, identity cache, component-scoped vs widened split, connector hook and read-only guard attribution (the guard fires before any filtering).
- Pure predicate `isPluginBookkeepingChange` in `shared/figma-changes.ts`: a `PROPERTY_CHANGE` whose every property is `pluginData` is dropped before bookkeeping, never enters the feed, never arms idle. Mixed lists, PAGE and node changes are kept. Drops are counted in STATUS `pluginDataChangesDropped` (present only when non-zero).
- Correction store read once and written at most once per batch (`CorrectionBatch`); a store read that throws at batch start is recorded (STATUS `captureErrors`) and the batch's feed edits still go out.
- One uncapped page walk (`page-of-node.ts`) shared by capture and the connector modules; the feed never substitutes the current page for a node it can resolve, and a genuine fallback is counted (`pageFallbacks`).

## Evidence

- Gates after the last edit: typecheck 0 · bundle 0 · lint:comments 0 · test 146 files / 1904 passed.
- Counting test: a 200-change batch now costs 2 `getSharedPluginData` calls (was 400) and 2 writes for one landed correction (was 402 reads).
- Mutation check: 10 targeted mutations, 8 caught; the 2 survivors are `main.ts` wiring lines that no test can import (module-load `figma.showUI`).
- Review round: capped page walk (wrong page in the feed) found and replaced by the shared uncapped walk, pinned by a 26-hop cross-page test that fails on the capped code.

## Not in this PR

Bounded/sliced walker, dirty-page idle and `perf.*` STATUS (next PR); relay probe/buffer.


jangtrinh (commented):
## Summary

Round-2 review. The four round-1 findings are genuinely closed, and each one is pinned by a
test that fails when the fix is reverted (verified by restoring the pre-fix shape byte-exact,
running the suite, then restoring the file):

- capped page walk → `plugin/src/main/page-of-node.ts:16` is uncapped; re-capping it at 20 hops
  fails `tests/page-of-node.test.ts:39` **and** `tests/document-change-capture.test.ts:196`
  (26-hop cross-page node). The deleted `connector-render.ts` copy was byte-identical to the new
  one, so connector reroute/verify semantics are unchanged — no drift.
- module-level cache → the capture module holds no module state; `createEditIdentityCache()` is
  injected (`plugin/src/main/main.ts:186`), session state stays in `main.ts`.
- over-broad predicate → node type is out of the test (`shared/figma-changes.ts:112`); flipping
  `.every` to `.some` fails `tests/figma-changes-coalesce.test.ts` and the mixed-list capture test.
- the tester's mid-batch store-read throw → removing the per-node try/catch
  (`document-change-capture.ts:153-160`) fails `tests/document-change-capture.test.ts:291`.

Behaviour vs `origin/main`'s handler is preserved where it matters: actor classification, the
DOC_CHANGE/EDIT_FEED split, the connector hook before the posts, the identity cache, idle arming,
and the read-only guard firing once per batch **before** any filtering (`document-change-capture.ts:107`)
— that ordering is right and should stay, since filtering its input would blind it to an EXEC_JS
script that legitimately writes `pluginData`. Gates re-run here after the last commit: typecheck
clean, `npm run build` reproduces `plugin/code.js`/`plugin/ui.html` byte-identically,
`npm run lint:comments` clean, `npm test` 146 files / 1904 passed.

What blocks: the "every guess and every drop leaves a counter" contract this PR introduces has
two holes, both one-liners, both contradicting the PR/commit line *"a fallback to the current
page is counted, never silent."*

## Risk level

Medium. No data loss or trust-boundary issue; two paths write a guess or lose a bookkeeping
record silently — exactly the class this PR exists to close.

## Findings

### Important

**1. A correction-store WRITE that throws loses the batch's corrections with no counter and
escapes the handler** — `plugin/src/main/document-change-capture.ts:245`

```ts
if (correctionsUsable) deps.corrections.flush(correctionBatch);
```

The read path directly above records every failure into `stats` (`errorCount`/`firstError` →
STATUS `captureErrors`). The flush has no guard, and
`tests/document-change-capture.test.ts:264-266` pins the throw as propagating.

Scenario: a view-only file (Figma refuses `setSharedPluginData`), or an entry at the byte cap. A
remote collaborator edits a node that already has an `agent-operation` parent in the document
store → the correction is appended in memory → `flushCorrectionBatch` → `writeEdgeCorrections`
(`correction-edge-store.ts:180`) throws. The correction is dropped with zero trace (`captureErrors`
stays absent, so STATUS reports a healthy session) and the exception leaves the `documentchange`
listener. `document-change-capture.ts:53-58` claims the opposite ("each leaves a counter behind").

Fix: wrap the flush in the same `try/catch` → `recordCaptureError(error)`. Update the pinning
test from `.toThrow(...)` to "feed posted, idle armed, `errorCount === 1`, first message verbatim".

**2. The DELETE page fallback is a silent guess** — `plugin/src/main/document-change-capture.ts:186-188`

```ts
const page = removed
  ? known?.page ?? figma.currentPage.name
  : resolvedPage(node, known?.page);
```

The live-node branch counts its current-page substitution; the removed branch does not.
Reproduced locally with a throwaway test: a `DELETE` for a node this session never saw emits
`{ nodeId: 'gone', page: 'Cover', nodeName: null }` while `stats.pageFallbacks` stays `0`.

Scenario: the plugin connects mid-session and the designer deletes a frame on a page they are not
viewing. The feed files that delete under the viewport's page — a wrong fact about which page lost
a node — and STATUS shows no guess was ever made. The field's own docstring says "live nodes", so
it is self-consistent; the PR/commit claim and the house rule are not.

Fix: count it (`if (known?.page === undefined) stats.pageFallbacks += 1;` on the removed branch),
or carry `page: null` for an unknown delete and let the sentence layer degrade honestly, the way
`nodeName` already does.

### Suggestion

**3. Nothing pins "own chain beats the remembered page"** — `document-change-capture.ts:88-94`.
Swapping the first two lines of `resolvedPage` (cache first, chain second) survives the entire
1904-test suite. That is the wrong-page class already fixed once this round, one line away from
returning. Add: remember a node on `Page 1`, then deliver a change for the same id whose parent
chain ends at `Page 2`; expect `page: 'Page 2'` and `pageFallbacks === 0`. (This also corrects the
PR's "the 2 survivors are `main.ts` wiring lines" — at least one semantic mutation inside the
tested module survives.)

**4. DOCUMENT-node non-pluginData changes now inflate `pageFallbacks`** —
`document-change-capture.ts:194-204`. Verified: a file rename (`PROPERTY_CHANGE ['name']` on
`0:0`) enters the feed as an `updated` edit typed `DOCUMENT` with `page: <current page>`, and is
counted. Not a fabricated edit and not silent, so: acceptable under-report, not a wrong fact. But
a document-level change has no page by construction, so it dilutes a counter whose job is to flag
*node* edits filed under a guess. Consider `page: null` for the DOCUMENT node. Not blocking.

**5. Comment asserts I/O that does not happen** — `document-change-capture.ts:108-115` and `:66`:
"Opening the batch and every per-node record run inside a guard because both reach
`sharedPluginData`". `beginCorrectionBatch` (`correction-edge-store.ts:255-257`) returns
`{ events: null, appended: 0 }` and touches nothing; the read is lazy, inside
`recordDesignerCorrectionInBatch`, so the `begin()` catch is dead today. Same wording in the PR
body ("a store read that throws **at batch start**"). Reword to the lazy read; keep the guard.

**6. `recordDesignerCorrection` has no production caller left** —
`correction-edge-store.ts:314`. Only `tests/correction-edge-store.test.ts` uses it. Either port
those tests to `beginCorrectionBatch`/`recordDesignerCorrectionInBatch` and delete it, or say in
the docstring why the single-shot API is kept.

**7. Trailers** — the commit message and PR body carry authoring-tool trailers and a session
link; `CLAUDE.md`'s house rules ask for none in either. Also worth running
`npm run lint:comments:update-baseline` in a follow-up: the run reports 7 stale baseline entries,
which are the citations this PR legitimately retired by restating the invariants in the moved
comments.

## Verdict

**Request changes** — for findings 1 and 2 only. Both are small, and both are contradictions of
the counting contract this PR introduces rather than regressions from `main`; everything else
(page walk, predicate, batch store I/O, read-failure resilience, STATUS shape and its verbatim
CLI pass-through, rebuilt bundles) checks out. Fix those two test-first (each test must fail
against the current code), fold in 3 if cheap, and this lands.

*Posted by the installed review-pr skill at 2026-09-03T04:39:15Z*

jangtrinh (commented):
## Summary

Re-review at `c7b211e`. Both blocking findings are closed, each pinned by a test that goes red
when the fix is reverted (mutations applied, run, then restored byte-exact — working tree back to
`c7b211e`, verified by checksum).

**1. Correction flush no longer escapes the listener** — `plugin/src/main/document-change-capture.ts:257-263`

```ts
if (correctionsUsable) {
  try {
    deps.corrections.flush(correctionBatch);
  } catch (error) {
    recordCaptureError(error);
  }
}
```

Removing that `try/catch` fails `tests/document-change-capture.test.ts:294`. The rewritten test
asserts the whole contract, not just the absence of a throw: feed posted, `idleArmed === 1`,
`errorCount === 1`, `firstError` verbatim. A view-only file or a byte-cap refusal now loses this
batch's corrections *with a counter*, and the comment states plainly that there is no scoped copy
to retry with — an honest limit rather than a silent one.

**2. A DELETE's page fallback is counted** — `document-change-capture.ts:97-103`, used at `:195`

```ts
function resolvedRemovedPage(remembered: string | undefined): string {
  if (remembered !== undefined) return remembered;
  stats.pageFallbacks += 1;
  return figma.currentPage.name;
}
```

Reverting the call site to `known?.page ?? figma.currentPage.name` fails
`tests/document-change-capture.test.ts:262` (unseen delete → `page: 'Cover'`, `pageFallbacks: 1`)
while the companion at `:272` (cached delete → `page: 'Page 1'`, `pageFallbacks: 0`) still passes,
so the pair separates "guessed" from "remembered" rather than just counting every delete.

**3. Page-resolution order is now pinned** — `tests/document-change-capture.test.ts:217`. Swapping
the first two lines of `resolvedPage` (cache before own chain) fails it. That closes the one
surviving semantic mutation from the last round: a node reparented to another page is filed by its
chain, not by the stale cache entry.

**4. The stale `begin()` comment is fixed** — `document-change-capture.ts:118-131` now describes
the actual shape (allocation only, lazy read at the first `record`, one guarded write at the end),
and the `errorCount` docstring at `:66-68` was updated to include the flush write.

Gates re-run after the last commit: typecheck clean · `npm run build` reproduces `plugin/code.js`
and `plugin/ui.html` byte-identically · `lint:comments` clean · `npm test` 146 files / 1907 passed.
Everything confirmed in round 1 still holds: uncapped shared page walk, the pluginData-only
predicate, one store read and at most one write per batch, read-failure resilience, actor
classification, the DOC_CHANGE / EDIT_FEED split, the connector hook, the read-only guard firing
before any filtering, and STATUS fields present only when non-zero, passed through by the CLI.

## Risk level

Low. Behaviour for real designer edits is unchanged from `main`; the new code paths are the two
counters and one guard, all directly covered.

## Findings

### Critical / Important

None.

### Suggestion

**S1. Two docstrings still say "live nodes" for a counter that now counts deletes too** —
`document-change-capture.ts:62-65` ("How many live nodes this session had NO page for…") and the
STATUS comment at `executor-ops.ts:141-142` ("how many live nodes had no resolvable page"). After
fix 2 the counter also covers an unseen delete. Two-line reword; the repo's "no wrong facts" rule
applies to comments as much as to the feed.

**S2. `CorrectionRecorder.begin` is now called unguarded** — `document-change-capture.ts:131`. Correct
today (`beginCorrectionBatch` only allocates), but this is an injection seam: the "no I/O, must not
throw" requirement lives in a body comment, not in the interface docstring at `:29`. Put it on the
`begin` member so a future recorder cannot quietly reintroduce a throw that escapes the listener.

**S3. DOCUMENT-node changes inflate `pageFallbacks`** — `document-change-capture.ts:196-213`. A file
rename (`PROPERTY_CHANGE ['name']` on `0:0`) enters the feed as an `updated` edit typed `DOCUMENT`
with `page: <current page>` and bumps the counter. Counted and not fabricated, so this is an
acceptable under-report — but a document-level change has no page by construction, so it dilutes a
signal meant for node edits. Consider `page: null` for the DOCUMENT node. Not blocking.

**S4. `recordDesignerCorrection` has no production caller** — `correction-edge-store.ts:314`; only
`tests/correction-edge-store.test.ts` uses it. Port those tests to the batch API and delete it, or
state in the docstring why the single-shot API stays.

**S5. Shrink the comment-hygiene baseline in a follow-up** — the lint run reports 7 stale entries,
which are the citations this PR legitimately retired by restating their invariants in the moved
comments. `npm run lint:comments:update-baseline` is exactly the sanctioned use here.

**Accepted, not a finding:** the commit and PR-body trailers are required by the environment this
branch was authored in and stay as-is.

## Verdict

**Approve.** Both blockers are fixed at the root with tests that fail against the pre-fix code, the
new counters separate a remembered page from a guessed one, and every gate passes on the final
commit. S1-S5 are follow-ups and need not hold the merge.

*Posted by the installed review-pr skill at 2026-09-03T04:48:21Z*

PR #106, opened 2026-09-03
perf(plugin): bound and slice the gap-fill walk, diff oversized pages by top-level frames

jangtrinh:
## Why

The remaining boot stall was the walk itself: `page.findAll(() => true)` is atomic and materialises every node before the 4000-record cap applies, so an 80k-node page held the editor thread for over a second, and the idle rewrite re-walked every page in one tick. Gap-fill was also blind on every page over the cap: a truncated page suppressed its whole diff, which on the owner's file meant 15 of 21 pages could lose closed-window edits silently.

## What

- Bounded, sliced walker (`page-walk-bounded.ts`): manual DFS that stops at cap+1 (the cap now bounds work, not just output), takes `parentId` from its own stack instead of the expensive `node.parent` read, and yields to the host every 500 nodes. Per-slice timings feed STATUS `perf` (`bootLoadAllPagesMs`, `bootWalkMs`, `bootWalkMaxSliceMs`, `bootSlices`, idle equivalents, `propertyReadErrors`).
- Every page also stores a top-level fingerprint (`[id, name, type, x, y, w, h, childCount]` per `page.children` entry). A page over the cap diffs that fingerprint instead of skipping: top-level frames report created/deleted/renamed/moved/resized, and a child-count change reports as `subtree` ("Contents of … changed while the plugin was closed"). STATUS `gapfill.pagesTopLevelOnly` says how many pages are covered this way.
- Idle no longer re-walks the document: the capture layer marks each edited page dirty; the idle write walks only dirty pages and carries the rest forward (`idle-baseline-write.ts`, testable outside `main.ts`).
- Honesty guards: a page with a property-read error carries its previous entry forward and emits one `walk-errors` notice rather than fabricating deletes; every `deleted` candidate is re-checked with `getNodeByIdAsync` because the walk spans yields (survivors counted in `gapfill.deletedRechecked`). `skipInvisibleInstanceChildren` was tried and dropped: under a capped DFS it changed which nodes were seen, not the stall budget, and turned instance visibility toggles into fake created/deleted frames.
- Baseline key bumped to v3 (new `top` payload); a v2 value reports as missing and is cleared. Pure diff core split into `gapfill-diff.ts`.

## Evidence

- Gates after the last edit: typecheck 0 · bundle 0 · lint:comments 0 · test 148 files / 1988 passed.
- Live, in the owner's real sandbox (walker bundle run read-only over all 21 pages of the 418k-node file): total 3.7 s wall, 141 slices, worst slice 45 ms, 80k-node page 143 ms, 0 property-read errors. Before: 0.3–1.5 s per page at boot and 10–15 s in one tick at idle/close.
- Mutation check: 12 targeted mutations; the 4 that survived the first pass (flag restore, idle dirty-only, dirty-set timing, max-slice metric) are now each killed by a dedicated test after the fix round.
- Review round: two wrong-fact paths (instance-visibility flag, throwing-node subtree reported as deleted) found and closed at the root.

## Not in this PR

Relay probe/buffer (next PR). Progressive page load stays gated on the cold-open `perf` numbers. Lowering the slice size to 400 if a slower machine breaches 50 ms.


jangtrinh (commented):
## Summary

Second-round review of the bounded/sliced walker, the top-level fingerprint signal, and the
dirty-page idle write. Every finding from the first round is closed at the root, and each fix
is held by a test that dies when the fix is reverted (verified by mutation, sources restored
byte-exact):

| Prior finding | Closed at | Mutation that kills a test |
|---|---|---|
| B1 instance-visibility flag → fake created/deleted | flag removed; `page-walk-bounded.ts:206-220` documents why | `page-walk-bounded.test.ts:276,287` pin "walks hidden children, writes no global" |
| B2 throwing subtree reported as deleted | `edit-gapfill.ts:404-409` (page diff skipped, prev entry kept) | `if (walk.propertyReadErrors > 0)` → `if (false)`: 3 failures |
| H1 vacuous max-slice assertion | `edit-gapfill-baseline.test.ts:916` asserts exact ms | `maxSliceMs = sliceMs` (last, not worst): 1 failure |
| H2 reparent mid-walk → fake deleted | `gapfill-diff.ts:143-154` + `edit-gapfill.ts:281-289` | drop the filter in `withoutSurvivingNodes`: 2 failures |
| H3 `subtree` conflating resize | `gapfill-diff.ts:247-251`, `edit-vocabulary.ts:136` | resize → `['subtree']`: 5 failures |
| M1 module split | `gapfill-diff.ts` (pure) / `edit-gapfill.ts` (figma) | — |

Also verified independently: the re-check runs on both record-level and top-level `deleted`
candidates (`edit-gapfill.ts:426,431`); a whole deleted page is still ONE notice with zero
node lookups (`edit-gapfill.ts:372-374`); an oversized page with read errors emits only
`walk-errors`, never the truncation notice (`edit-gapfill.ts:398-409`); the idle writer claims
and clears the dirty set before the first `await` (`idle-baseline-write.ts:38-41`), walks only
dirty pages, carries the rest forward by reference, and still walks a page the baseline has
never seen (`edit-gapfill.ts:211-214`); `bootBaselineUnreadable` withholds every later write
(`edit-gapfill.ts:194-197`); the v2→v3 key bump degrades to `baseline-missing` and the stale
value is cleared only after the new one lands, counted (`gapfill-baseline-store.ts:200-209`).
Under-cap behaviour is pinned against a verbatim copy of the old walker
(`page-walk-bounded.test.ts:182`, real oracle over one shared fixture tree, not two fixtures
agreeing). Gates re-run here after the last commit: typecheck 0, `lint:comments` clean (7
stale baseline entries noted, 0 new), 148 files / 1988 tests passed. No new `any`, no lint
suppression, no swallowed error without a counter. Bundle rebuilt (`plugin/code.js` carries
`walk-errors`, `deletedRechecked`, `figma-edit-baseline-v3`, `dirtyPageIds`); the vocabulary
sentences are CLI-side only, so their absence from the bundle is correct.

PR body claims check out against the diff, gate numbers included.
`GapfillStatus.pagesTopLevelOnly` is a new required field, but no CLI or panel code reads
these fields (STATUS passes through as an opaque blob), so there is no runtime break.

## Risk level

Medium — the design is sound and the honesty guards are real; three items below can each put
a wrong fact, or silence, in front of the owner.

## Findings

### Important

**1. `nodeStillExists` is the one new trust boundary in this PR and has no test at all.**
`plugin/src/main/edit-gapfill.ts:281-289` decides whether a `deleted` candidate is reported or
suppressed, and it accepts any non-null resolution as "still exists". The figma stub in
`tests/edit-gapfill-baseline.test.ts:51-58` defines no `getNodeByIdAsync`, so every test that
does not inject `deps.nodeExists` silently rides the "host has no getter" branch with nothing
asserted; the throw branch and the null branch are equally uncovered. That is precisely the
permissive-mock case CLAUDE.md rules out ("Figma's dynamic-page getters throw; permissive
mocks are green lights that mean nothing").
Failure scenario: if the host ever resolves a removed node to a handle rather than to `null`
— the repo's own convention elsewhere treats `.removed` as "the honest check, not a stale
local" (`executor-exec-js.ts:64-67`) — then EVERY closed-window deletion is suppressed and
survives only as an anonymous `gapfill.deletedRechecked` counter. The deleted category
vanishes from the feed with no per-page trace, which is the exact failure this phase exists to
prevent. Fix is small: `return node !== null && !(node as { removed?: boolean }).removed;`
plus one test driving the real predicate through a figma stub with four shapes (getter absent,
getter throws, id resolves, id resolves to a removed handle).

**2. The truncation notice states a reason that is false for a page that shrank below the
cap.** `plugin/src/main/edit-gapfill.ts:411-415` emits `truncatedNotice` whenever EITHER side
was truncated, and `shared/edit-vocabulary.ts:90` renders it as `Gap-fill covers only
top-level frames on "X" while it exceeds the scan cap`.
Scenario: page had 6,000 nodes last session (`truncated: true`, no records stored); the owner
deletes 3,000 while the plugin is closed. This boot walks 3,000 nodes, `nextTruncated` is
false, and the feed still tells the owner the page exceeds the scan cap — it does not.
`gapfill.pagesTruncated` counts it too, and `protocol.ts:370` defines that field as "the page
exceeds the scan cap", so the wrong fact reaches STATUS as well. Suppressing the per-node diff
here is correct (the previous side stored no records); only the stated reason is wrong. The
trigger predates this PR, but the sentence was rewritten in it — worth one more clause, e.g.
splitting the "previous baseline was over the cap" case from the "this page is over the cap"
case.

**3. The slice budget's own comment is contradicted by the PR's live measurement.**
`plugin/src/main/edit-gapfill.ts:51-54` claims `SNAPSHOT_SLICE_SIZE = 500` is "≈ 25–30 ms of
held thread — inside the 50 ms 'no visible stall' budget with room for a slower machine". The
PR's own evidence measured a 45 ms worst slice on the owner's machine, i.e. 90% of the budget
and ~1.5× the comment, on the fastest environment this will run in. The PR defers the fix
("lowering the slice size to 400 if a slower machine breaches 50 ms"), but the number that
would justify that decision is already in hand. Either drop the constant now, or cut slices on
elapsed time rather than node count (per-node cost varies by node type), or at minimum restate
the comment with the measured worst case — a stale estimate here is what the next reader will
size the budget from.

### Suggestion

**4. Two comments still blame the deleted visibility flag.** `shared/protocol.ts:419-421`
documents `propertyReadErrors` as "a reference invalidated by the scoped visibility flag", and
`page-walk-bounded.ts:64` says "invalidated by the flag flip". This PR removed that flag; the
cause is now a stale reference or a host refusal.

**5. The survivor re-check is an unbounded serial round trip on the boot path.**
`gapfill-diff.ts:143-154` awaits one lookup per `deleted` candidate; the comment says "normally
a handful", which the live run (0 read errors, small diff) did not test. A 2,000-node subtree
deleted on an under-cap page produces 2,000 sequential awaits before the batch posts. No
thread stall (each await yields), but boot gap-fill is delayed by an amount nothing counts.
Consider a counter for candidates checked, and a threshold above which a mass deletion is
taken at face value.

**6. Size and position use different comparison rules.** `gapfill-diff.ts:249-250` compares
width/height with `!==` while moves use `MOVE_EPSILON = 0.5` (`gapfill-diff.ts:63`). Both
sides are half-pixel normalized, so an auto-layout reflow of 100.2 → 100.4 crosses a bucket
and reports "Resized" while an identical positional jitter is correctly suppressed.

**7. "Resized" now outranks the residual bucket for LIVE edits too.**
`shared/edit-vocabulary.ts:136` is not scoped to gap-fill frames, so a live batch with
`changedProps: ['width','height','fills']` changes from "Restyled" to "Resized" and stops
mentioning the paint change. Neither sentence is wrong, but this is a vocabulary ranking
change beyond the stated H3 scope; the comment above it justifies only the fingerprint case.
Worth an explicit note (or a test) that the live path was meant to change too.

**8. Coverage is now inverted between small and oversized pages, and `edit-gapfill.ts` is
still over the size guidance.** A frame resize or child-count change is reported on a page
OVER the cap (top-level diff) and not on a page UNDER it (`edit-gapfill.ts:430` diffs records
only), so the owner gets less signal from the smaller page. Deliberate given the
existence/name/position non-goal, but worth one line saying so where the asymmetry is created.
Separately, `edit-gapfill.ts` is 438 lines / 253 non-comment lines against the repo's 200-line
guidance; `gapfill-diff.ts` (193) and `page-walk-bounded.ts` (124) are inside it.

## Verdict

**Request changes** — for findings 1–3 only. The core of the phase is right: the cap now
bounds work, the honesty guards are real and independently reproducible, and every first-round
finding is closed with a test that actually fails without its fix. Finding 1 is the one that
matters (an untested boundary whose failure mode is silent loss of a whole fact category);
2 and 3 are each a one-line correction to a fact this repo has decided not to get wrong.

*Posted by the installed review-pr skill at 2026-09-03T06:01:46Z*

jangtrinh (commented):
## Summary

Re-review at 41b2f64. All three blocking findings are closed at the root, each held by a test
that dies when the fix is reverted (mutations run here, sources restored, tree clean):

| Finding | Closed at | Mutation → result |
|---|---|---|
| 1 `nodeStillExists` untested, no `removed` guard | `edit-gapfill.ts:301-310` — exported, `node !== null && !node.removed`, four host shapes driven through a stub (`edit-gapfill-baseline.test.ts`) | drop the `.removed` clause → 1 failure |
| 2 truncation notice stated a false reason | `edit-gapfill.ts:434-450` splits by direction; `previouslyTruncatedNotice` (`gapfill-diff.ts:331-348`) + its own sentence (`edit-vocabulary.ts:92-99`) | force the `nextTruncated` branch → 1 failure |
| 3 slice budget was a node count | `page-walk-bounded.ts:104-125,199-208` — cut on elapsed time, node count as upper bound | remove the time cut → 3 failures; remove the count cut → 6 failures |

The direction split is complete rather than cosmetic: `pagesTruncated` now counts only pages
over the cap AS OF this walk, `pagesTopLevelOnly` counts both suppressed cases and is
explicitly documented as not a subset (`protocol.ts:370-380`, `gapfill-status.ts:12-22`), and
the 2×2 prev/next table is pinned as a table-driven test. The shrunk page still stores fresh
records, so the next session gets its per-node diff back. `budgetClock` is deliberately a
different seam from `SlicedWalkTiming.now` with the reason stated at
`page-walk-bounded.ts:118-125` — the exact-stamp STATUS assertions from the last round still
pass, which is the evidence that separating them was necessary. Both flag-blaming comments are
gone (`protocol.ts:426-428`, `page-walk-bounded.ts:65-67`).

**Post-`git checkout` integrity.** I diffed both affected files across 975412e..41b2f64 in full
and compared their declaration/control-flow skeletons: the only deltas are
`SNAPSHOT_SLICE_BUDGET_MS`, `DEFAULT_SLICE_BUDGET_MS`, the `nodeStillExists` export + guard,
the `if (nextTruncated)` split, and the new slice-cut condition. Nothing dropped, duplicated or
half-applied. To confirm the earlier round survived the revert, I re-ran the round-1 mutation
set against this head: read-error gate → 3 failures, idle dirty-only → 2, survivor filter → 2,
resize→subtree → 4, max-slice metric → 1. All still bite. One weaker variant survives — moving
`dirtyPageIds.clear()` one microtask later keeps `idle-baseline-write.test.ts` green — but
`documentchange` arrives as a host callback, so no edit can land in a microtask gap; the
load-bearing rule (clear before the awaited walk) is pinned. Not a defect, just the exact scope
of what that test proves.

Gates re-run here after the last commit: typecheck 0, `lint:comments` clean (7 stale baseline
entries noted, 0 new), 148 files / 2003 tests passed — matching the PR's claim. `plugin/code.js`
is current and carries all three fixes (`node !== null && !node.removed`,
`budgetClock = Date.now`, `prev-truncated`); `ui.html` differs only by the version stamp. The
live sandbox result (worst slice 21 ms at a 20 ms budget, i.e. budget + one node, at the cost of
251 slices and +0.5 s wall) is consistent with the mechanism and is a good trade for the target.

## Risk level

Low — the closed-window honesty guards are now each backed by a test that fails without them,
and the stall target is held by a bound that does not depend on machine speed.

## Findings

### Critical

None.

### Important

None.

### Suggestion

**1. The survivor re-check is an unbounded serial round trip on the boot path.**
`gapfill-diff.ts:143-154` awaits one lookup per `deleted` candidate; the comment says "normally
a handful", which no run has yet contradicted or confirmed at scale. A 2,000-node subtree
deleted on an under-cap page produces 2,000 sequential awaits before the batch posts. No thread
stall (each await yields), but boot gap-fill is delayed by an amount nothing counts. Consider
counting candidates checked, and a threshold above which a mass deletion is taken at face value.

**2. Size and position use different comparison rules.** `gapfill-diff.ts:249-250` compares
width/height with `!==` while moves use `MOVE_EPSILON = 0.5` (`gapfill-diff.ts:63`). Both sides
are half-pixel normalized, so an auto-layout reflow of 100.2 → 100.4 crosses a bucket and
reports "Resized" while identical positional jitter is correctly suppressed.

**3. "Resized" now outranks the residual bucket for LIVE edits too.**
`shared/edit-vocabulary.ts:136` is not scoped to gap-fill frames, so a live batch with
`changedProps: ['width','height','fills']` changes from "Restyled" to "Resized" and stops
mentioning the paint change. Neither sentence is wrong, but this is a vocabulary ranking change
beyond the top-level-fingerprint case the comment above it justifies. Worth an explicit note or
a test that the live path was meant to change too.

**4. The top-level fingerprint is counted against the first slice but not bounded by it.**
`page-walk-bounded.ts:163-175` starts the budget clock, then runs `topLevelFingerprint` over
every direct child in one uninterruptible block before the first per-node budget check. On the
owner's file (≤ 458 frames) the measured 21 ms max slice proves this is cheap; on a slower
machine or a wider page it is the one path that can exceed the budget by an unmeasured amount.
Cheap hedge: report the fingerprint's own elapsed time, or cut the first slice after it.

**5. Coverage is inverted between small and oversized pages, and `edit-gapfill.ts` is over the
size guidance.** A frame resize or child-count change is reported on a page OVER the cap
(top-level diff) and not on a page UNDER it (`edit-gapfill.ts:465` diffs records only), so the
owner gets less signal from the smaller page — deliberate given the existence/name/position
non-goal, but worth one line where the asymmetry is created. Separately `edit-gapfill.ts` is
473 lines / 262 non-comment against the repo's 200-line guidance, and `gapfill-diff.ts` has
just crossed it at 201; both are cohesive today, but the next addition to either should land in
a new module.

## Verdict

**Approve.** The three blocking findings are fixed at the root, not worked around: a resolved
handle is no longer assumed live, the two truncation reasons are now two facts with two
sentences and two counters, and the stall budget is expressed in the unit the target is written
in. The revert-and-reapply left no residue — the full file diff and the re-run of the previous
round's mutations both come back clean. The five remaining items are non-blocking; item 1 is
the one worth a follow-up issue.

*Posted by the installed review-pr skill at 2026-09-03T06:23:48Z*

PR #107, opened 2026-09-03
perf(plugin): probe broker ports concurrently and buffer captures until the socket opens

jangtrinh:
## Why

Two relay costs sat between "plugin opened" and "agent can talk to it". Broker discovery probed the ten ports one after another with a 1.2 s timeout each, so a cold open with no broker took up to 12 s to say so, and a broker on a late port cost ~11 s on every reconnect. And any frame the main thread posted before the socket was open — the boot gap-fill notice, an owner edit captured during a reconnect gap — was dropped by `wsSend` with no trace, against the house rule that nothing vanishes silently.

## What

- Concurrent probing (`broker-scan.ts`): all ten ports are probed at once, the first `BROKER_HELLO` wins, every other socket is closed, and a pass with no broker resolves after one timeout window instead of ten. Backoff and the correlated heartbeat/app-probe handshake are untouched.
- Bounded pre-OPEN outbound buffer (`outbound-buffer.ts`): `EDIT_FEED`/`DOC_CHANGE` frames posted while no socket is open are queued (200 frames / 1M chars), flushed in order after `PLUGIN_HELLO`, and an overflow drops the oldest and counts it. Replies, heartbeat/lease frames and `FILE_INFO` (which fires on every selection change and would evict real captures) are never buffered.
- Replayed captures are honest: each buffered event is stamped `capturedAt` at enqueue and marked `replayed`; the broker dates the feed frame at capture time (rejecting an implausible stamp and counting the rejection) and counts a replayed batch as live activity for `cowork` waiters only while it is younger than the quiet window. The feed frame schema gains an optional `replayed` flag; older feeds still parse.
- The drop tally travels in a separate `PLUGIN_RELAY_STATS` event after the handshake, only when non-zero; the relay accounts the delta since its last delivered report, the broker logs what arrives and holds no per-instance state, and an older broker re-broadcasts the event to CLI clients, which ignore it. `PLUGIN_HELLO` and scene identity are byte-identical to before, so the background-file watchdog is unaffected.
- Socket adoption logic (`socket-adoption.ts`) is DOM-free and testable: duplicate adoption is refused, a socket that is no longer open at adopt time keeps the buffer intact, and the flush order is HELLO → RELAY_STATS → captures → heartbeat start.

## Evidence

- Gates after the last edit: typecheck 0 · bundle 0 · lint:comments 0 · test 153 files / 2076 passed (panel gate included).
- Mutation check: 13 targeted mutations, 11 caught on the first pass; the 2 survivors (duplicate-adoption guard, broker drop-log line) are now each pinned by a test after extracting the adoption module and adding daemon-harness coverage.
- Review round: replayed frames being dated at broker receipt, an unreachable retention guard, and a mixed-version identity leak were found and closed at the root.
- `PROTOCOL_VERSION` and `EDIT_FEED_SCHEMA_VERSION` unchanged; every addition is optional.

## Not in this PR

Lazy-loading the html-to-figma converter: esbuild refuses code splitting under the single-file iife build the plugin requires, and the converter has no top-level side effects, so deferral would buy nothing. Progressive page load stays NO-GO on the cold-open numbers (`bootLoadAllPagesMs` 0).


jangtrinh (commented):
**Summary**

Second-round review of the concurrent broker probe + pre-OPEN capture buffer, at `0365e8c`. All three
round-1 blockers are genuinely closed at the root, and I verified each by mutation rather than by
reading:

- **Replay dating** — captures are stamped `capturedAt`/`replayed` at enqueue
  (`plugin/src/ui/outbound-buffer.ts:59`), and the broker files the batch at capture time
  (`shared/protocol.ts:81`, `cli/src/transport/broker-daemon.ts:337`). Neutering `readCapturedAt` to
  always return `now` fails 6 tests across `tests/replayed-capture-time.test.ts` and the daemon harness.
  Downstream readers check out: `readEditFeed` sorts by `ts` (`cli/src/commands/changes.ts:93`) and the
  change-log cursor is a line count (`cli/src/transport/change-log.ts:2`), so an older `ts` appended
  later hides nothing from reconcile.
- **Dead retention guard** — the drain now happens only after an OPEN check
  (`plugin/src/ui/socket-adoption.ts:47`, wired at `plugin/src/ui/ui-relay.ts:550`). Removing the
  `already-adopted` branch fails 2 tests; the dead-winner case keeps the whole gap buffered.
- **Mixed-version identity** — `PROTOCOL_KEYS` is byte-identical to base
  (`cli/src/transport/plugin-registry.ts:106`), the HELLO payload is unchanged field-for-field
  (`ui-relay.ts:583`), and the tally moved to its own `PLUGIN_RELAY_STATS` event.
- Gates reproduced here after the last edit: `typecheck` 0 · `build` 0 (rebuild is byte-identical to the
  committed `plugin/ui.html`; `plugin/code.js` untouched) · `npm test` 153 files / 2074 passed ·
  `lint:comments` clean. `PROTOCOL_VERSION` and `EDIT_FEED_SCHEMA_VERSION` unchanged; the emitted
  `skills/figma-agent/SKILL.md` command table is untouched.

One finding blocks: the drop-tally de-duplication does not survive the reconnect it was written for, and
the two tests that appear to pin it exercise paths the daemon never takes.

**Risk level:** medium-low. No data-loss or trust-boundary defect found. The blocker is log honesty plus
a phantom pin; everything else is a follow-up.

---

## Findings

### Important

**1 — the per-instance drop-log delta is erased by the very reconnect it exists for**

`cli/src/transport/plugin-registry.ts:245` carries `preOpenDroppedLogged` across a re-`registerPlugin`,
and its comment states the guarantee: *"Carried across a reconnect on purpose … so a reconnect
re-reporting an already-logged total must stay silent."* The PR body repeats it ("the broker logs the
delta per plugin instance"). But a reconnect does not normally go through supersede — the socket closes
first, and `handleClose` calls `st.registry.removeByWs(ws)` (`cli/src/transport/broker-daemon.ts:1952`),
which does `this.plugins.delete(entry.instanceId)` (`plugin-registry.ts:344-347`). The entry, and the
baseline with it, is gone before the new HELLO arrives.

Scenario: the panel drops 7 frames during an outage, connects, broker logs `lost 7 capture frame(s) … 7
this session`. Figma backgrounds the editor tab, the socket closes, the panel reconnects and re-sends its
session-cumulative tally (the report is deliberately idempotent because nothing acks it). `existing` is
`undefined`, so the delta is `7 - 0`, and the same 7 is logged again — and again on every subsequent
reconnect. That is round-1 finding 5 unfixed on the only path that occurs in production, and the log now
over-states real loss to anyone summing it.

Reproduced against the shipped registry (temporary test, since removed): register → `reportPreOpenDropped
{frames:7}` → `removeByWs` → register same `instanceId` → `reportPreOpenDropped {frames:7}` returns
`{frames:7,chars:70}`, not `null`.

Both pinning tests miss it, and each misses it in the same direction:
- `tests/broker-daemon-harness.test.ts:2355` sends three `PLUGIN_RELAY_STATS` frames down **one** socket.
  Production sends exactly one per connection, so that sequence never happens.
- `tests/plugin-relay-stats-drop-report.test.ts` ("a reconnect of the SAME instance keeps its recorded
  position") calls `registerPlugin` twice with no `removeByWs` in between — the supersede race, not a
  reconnect.

Fix, either way is fine: keep the logged tally in a small `instanceId → PreOpenDroppedReport` map owned by
the daemon (outliving the registry entry, pruned with the plugin set), **or** drop the delta and log the
honest sentence ("session total, re-reported on every reconnect"). Whichever you pick, the pinning test has
to close the socket and reconnect, and the comment at `plugin-registry.ts:245` must match the code.

### Suggestions

**2 — cowork: decide by capture age, not by the `replayed` flag alone.** `broker-daemon.ts:2248` excludes
every replayed batch from live waiters. For a sub-second blip the designer *is* active, so a waiter that
survives the gap is told less than the truth. The window is narrow — a plugin disconnect already fails
every waiter on that instance with `E_NO_PLUGIN` (`broker-daemon.ts:1986-1991`), so only the half-open
supersede race reaches it — which is why this is a follow-up, not a blocker. Cheap upgrade: count a
replayed frame as activity when `now - capturedAt <= quietMs`; history stays history, a blip stays live.

**3 — `readCapturedAt` bounds the future but not the past** (`shared/protocol.ts:86`). A stamp is accepted
whenever it is finite and `> 0`, so a backwards clock correction (or a `capturedAt: 1`) files an edit
arbitrarily far in the past — the same permanent `changes --since` invisibility the forward skew guard was
added to prevent, reached from the other side, and uncounted. A generous floor (e.g. refuse older than the
broker's start minus a few days, count the refusal) closes it without breaking legitimate long outages.

**4 — a replayed frame is now invisible to a `--since` cursor.** Correct dating means an agent that polled
*during* the outage and advanced its cursor past the capture time never sees those frames: the frame no
longer records when it was appended, so nothing can recover them. Consider an additive optional
`receivedAt` (append time) alongside `ts`, so `--since` can offer late arrivals without lying about when
the edit happened.

**5 — "an older broker ignores the event" is not what happens.** An older daemon has no
`PLUGIN_RELAY_STATS` branch, so the frame falls into `else if (isPlugin) broadcastToClients(text)`
(`broker-daemon.ts:2288`) and is fanned out to every CLI client. Harmless in practice (clients ignore
unknown event types, `cli/src/transport/broker-client.ts:173`), and `tests/broker-daemon-harness.test.ts:2383`
does prove the connection survives — but the claim in the PR body and in the comments at
`shared/protocol.ts:33` / `plugin/src/ui/outbound-buffer.ts:100` ("costs that same broker nothing") should
say re-broadcast-and-ignored.

**6 — ten per-port PROBE transitions in one tick.** `ui-relay.ts:534` fires one transition per port, so the
detail line only ever shows `:9419` while all ten are in flight, and `since` is reset ten times per pass
(`ui-relay.ts:136`). With `RECONNECT_BACKOFF_MAX_MS` 8 s + jitter, the panel's honest branch (`ageMs >=
10_000`, `panel-model.ts:34`) stays effectively unreachable, so the phase's "≤ 1.5 s to 'no broker'"
criterion is delivered only in the legacy status DOM, not in the sentence a designer reads. One aggregated
`probing figma-agent.localhost:9410-9419…` transition fixes both halves.

**7 — the drop clause is never rendered by the panel gate.** `tests/panel-model.test.ts` covers the string;
`tests/figma-plugin-panel.test.ts` never sets a non-zero count, so `Connected · 12 edits lost while offline`
has not been laid out at the 200 px rail width. If it truncates, the loss is invisible exactly when it
matters.

**8 — flushed captures are still unacknowledged.** `flushHandshake` treats a returned `send()` as delivery;
a socket dying between the OPEN check and the OS write loses the batch with nothing counted. Unchanged from
before this PR and much narrower thanks to the OPEN check — worth one sentence in the module header so the
"nothing vanishes silently" claim stays exact.

**9 — structure.** `ui-relay.ts` grew 618 → 643 lines *after* 415 lines were extracted into three modules;
the next seam (inbound routing) would finally make the adopt→flush wiring importable — today those three
lines are the only part of this change with no execution coverage. `droppedNote` (`panel-model.ts:17`) is
exported for one in-file caller plus its test. Broker selection is now a race rather than lowest-port; in a
transient two-broker state the plugin can register with the non-advertised daemon (behaviour change only).

**10 — informational, no action.** The `lint:comments` note about 7 stale baseline entries is pre-existing:
the stale entries are in `panel-model.ts`, `panel-ui.ts`, `shared/safe-cleanup.ts` and
`tests/broker-daemon-harness.test.ts`, and `main` already lacks those citations. Correct not to launder
them through `--update-baseline` in this PR.

---

**Verdict: Request changes** — finding 1 only (a stated guarantee the code does not provide, pinned by two
tests that exercise paths the daemon never takes). Findings 2-10 are follow-ups and none of them needs to
land here.

*Posted by the installed review-pr skill at 2026-09-03T07:24:55Z*

jangtrinh (commented):
**Summary**

Re-review at `86974a8`. The round-2 blocker is closed at the root, not patched around: the delta now lives
where the state actually survives a reconnect — in the relay — and the broker computes nothing. I verified
each guard by mutation rather than by reading, and all gates reproduce here after the last edit.

**Risk level:** low. No open Critical or Important finding.

---

## Blocker closed — drop-tally accounting moved to the relay

**The accumulator and its clear condition.** `plugin/src/ui/outbound-buffer.ts:159` bumps `unreported`
alongside the session total on every eviction; `:130` emits a stats frame only while `unreported.frames >
0`; `:173` `confirmReported` *subtracts* rather than zeroing, so a drop caused by re-buffering during the
same flush is not silently absorbed by the frame that preceded it. The clear happens in exactly one place,
and only after delivery is as confirmed as the relay can make it:

- `plugin/src/ui/socket-adoption.ts:98` — `write()` refuses a non-OPEN socket (`send()` discards in silence
  on CLOSING/CLOSED, so a returned call is not evidence on its own);
- `plugin/src/ui/socket-adoption.ts:111` — `if (batch.reported) buffer.confirmReported(batch.reported);`
  sits *inside* the `batch.control.every(write)` branch, so the delta survives a failed or swallowed write
  and rides the next handshake.

Mutation-checked, mutants reverted byte-exact: deleting the `readyState` check fails *"keeps the
accumulator when the socket is not OPEN at send time"*; hoisting `confirmReported` above the write fails
that test plus *"keeps the accumulator when the host throws out of send"*.

**Broker statelessness.** `cli/src/transport/broker-daemon.ts:2134` reads `readRelayDropStats(msg.data)`
and logs exactly what arrived — no registry lookup beyond the instance id for the log line, no baseline, no
arithmetic. `cli/src/transport/plugin-registry.ts` is byte-identical to `main` again (`git diff main --
cli/src/transport/plugin-registry.ts` is empty), so `preOpenDroppedLogged` and `reportPreOpenDropped` are
gone rather than left as dead state. `readRelayDropStats` (`shared/protocol.ts:72`) still refuses a zero,
negative, non-numeric or non-object delta, and falls back to the delta when `sessionTotal` is missing or
smaller — the log states a number the frame actually supports.

**The harness path is the production one now.** `tests/broker-daemon-harness.test.ts:2361` registers, reports
`[5,40]`, then does a *real* `first.close()` and waits for `plugin [inst-drops-1] disconnected` (`:2381-2382`)
before re-registering the same `instanceId` on a new socket and reporting a fresh delta (`:2384-2387`). It
asserts exactly two delta lines with no repeated total (`:2391-2393`), plus that the handshake itself never
carried the tally (`:2396`). That is the sequence my last review reproduced as broken; it is the sequence
the test now drives.

The trade-off this inverts is the right way round: the old design could over-count (re-log the same total),
the new one can under-count (a report written to a socket that dies before the OS flush is cleared and never
re-sent) — and the panel still shows the session total while the frame carries `sessionTotal`, so a log
reader sees where the session stands. Documented at `socket-adoption.ts:77-84`.

## Cowork age gate — spot-checked

`broker-daemon.ts:2257-2262` computes `replayAgeMs` only for a replayed batch and skips a waiter only when
that age exceeds *that waiter's own* `quietMs`, so the window is per-waiter rather than a constant. Using
`readCapturedAt` (not `captureTimestamp`) is right: the append already logged any refusal, and a refused
stamp reads back as broker-now, i.e. the waiter treats the batch exactly as the feed dated it.

Both directions are pinned, and each mutation fails a different test: forcing flag-only exclusion fails
*"counts a replay captured inside the quiet window as live activity"* (`tests/cowork-harness.test.ts:534`,
500 ms replay against `waitMs` 800), and deleting the age gate entirely fails *"ignores the replayed gap,
then fires on the first genuinely live edit"* (`:513`, a 10-minute-old capture).

## Gates reproduced (after the last edit)

`typecheck` 0 · `build` 0 — a fresh build is byte-identical to the committed `plugin/ui.html`, `plugin/code.js`
untouched · `npm test` 153 files / 2076 passed · `lint:comments` clean. `PROTOCOL_VERSION` and
`EDIT_FEED_SCHEMA_VERSION` unchanged; `PLUGIN_HELLO` and scene identity byte-identical to base; the emitted
`skills/figma-agent/SKILL.md` command table untouched.

---

## Remaining suggestions (none blocking)

**1 — `readCapturedAt` bounds the future but not the past** (`shared/protocol.ts:114`). A stamp is accepted
whenever it is finite and `> 0`, so a backwards clock correction (or a `capturedAt: 1`) files an edit
arbitrarily far in the past — the same permanent `changes --since` invisibility the forward skew guard
exists to prevent, reached from the other side and uncounted. A generous floor, counted like the forward
refusal, closes it without breaking legitimate long outages.

**2 — a replayed frame is invisible to a `--since` cursor.** Correct dating means an agent that polled
*during* the outage and advanced its cursor past the capture time never sees those frames, and the frame no
longer records when it was appended, so nothing can recover them. An additive optional `receivedAt` would
let `--since` offer late arrivals without lying about when the edit happened.

**3 — a fresh replay re-arms the quiet window from arrival, not from capture.** `feedCoworkWaiter(…,
editNow)` (`broker-daemon.ts:2263`) stamps `armedAt` at replay time, so a batch captured `quietMs - ε` ago
extends the wait by a further full `quietMs`. Harmless, but arming at `capturedAt` would be the honest
number.

**4 — ten per-port PROBE transitions in one tick.** `plugin/src/ui/ui-relay.ts:534` fires one transition per
port, so the detail line only ever shows `:9419` while all ten are in flight, and `since` is reset ten times
per pass (`:136`). With `RECONNECT_BACKOFF_MAX_MS` 8 s + jitter, the panel's honest branch (`ageMs >=
10_000`, `panel-model.ts:34`) stays effectively unreachable, so the phase's "≤ 1.5 s to 'no broker'"
criterion lands only in the legacy status DOM, not in the sentence a designer reads. One aggregated
`probing figma-agent.localhost:9410-9419…` transition fixes both halves.

**5 — the drop clause is never rendered by the panel gate.** `tests/panel-model.test.ts` covers the string,
but `tests/figma-plugin-panel.test.ts` never sets a non-zero count, so `Connected · 12 edits lost while
offline` has never been laid out at the 200 px rail. If it truncates, the loss is invisible when it matters.

**6 — structure.** `ui-relay.ts` is 643 lines *after* 415 were extracted into three modules; the next seam
(inbound routing) would finally make the adopt→flush wiring importable — today those three lines are the
only part of this change with no execution coverage. `PreOpenDroppedReport` (`shared/protocol.ts:38`) and
`droppedNote` (`panel-model.ts:17`) are exported with no consumer outside their own file plus tests. Broker
selection is now a race rather than lowest-port; in a transient two-broker state the plugin can register
with the non-advertised daemon (behaviour change only).

**7 — informational, no action.** The `lint:comments` note about 7 stale baseline entries is pre-existing
(`panel-model.ts`, `panel-ui.ts`, `shared/safe-cleanup.ts`, `tests/broker-daemon-harness.test.ts`; `main`
already lacks those citations). Correct not to launder them through `--update-baseline` here.

---

**Verdict: Approve.** The blocker is closed at the root, the accounting lives where the state survives, and
both new guards plus the cowork age gate fail under mutation. Suggestions 1-7 are follow-ups.

*Posted by the installed review-pr skill at 2026-09-03T07:42:59Z*

PR #108, opened 2026-09-03
feat(panel): single-row rail that hugs its content, no expanded state

jangtrinh:
## Why

The owner asked for the panel to be one row that hugs its content, with no expanded state, and with the host window title "design:os by JANG" always shown in full. The previous 200 px rail truncated the title and hid the sync prompt, the connection sentence and the file target behind a chevron.

## What

- The inspector (tabs, onboarding, version row) is gone. The panel is one row: the semantic orb, the sentence, and the contextual target/sync buttons that appear only when they apply.
- The iframe measures the row's content width after every render and asks the main thread for it; main clamps to 240–560 px and never trusts anything else. 240 keeps the host title whole; past 560 the sentence gets an ellipsis with the full text in its tooltip.
- The sentence ranks what matters: a lost-edit note leads and is structurally unclippable (its own non-shrinking span), then connection trouble, then the sync result, then the current activity, then Idle.
- The sync button carries the pending count and runs the sync on click; it stays for a retry on failure. Clicking the sentence acknowledges unresolved failures (chip and orb clear); the record stays in the feed and `figma-agent errors`.
- The comment-hygiene baseline is untouched.

## Evidence

- Gates after the last edit: typecheck 0 · bundle 0 · lint:comments 0 · test 154 files / 2091 passed (panel design-lint gate included and net stricter).
- Chromium against the built `ui.html`: Idle measures 112 px → clamped to 240; a long sentence stops at 524 px; the hug is posted once per width change and not on the 1 s tick; the sync click dispatches exactly one request and reads "Syncing"; a dropped-edit note survives a reconnect; a 9-digit lost-edit count still renders unclipped (224 px).
- Review round: the lost-edit note was being ellipsised away behind a long connection message; fixed by ranking and structure, pinned by a browser measurement that was red before the fix.

## Live check for the owner

Reopen the plugin: the title should read "design:os by JANG" in full; the row should grow on a long sentence and shrink back; the sync button should run a sync in a bound file.


jangtrinh (commented):
## Summary

Re-review of `3472da4` on `feat/rail-hug-single-row` (base `ce6f811`). The three blocking/high items
raised in the first pass are closed, each with a test that fails against the pre-fix code. I re-ran the
gates and four mutations myself rather than taking the report's word; working tree restored byte-exact
after every mutation.

Gates re-run here: `npm run typecheck` exit 0 · `npm run lint:comments` exit 0 (clean, 0 new; it now
prints the expected "7 baseline entries are stale" note) · `npx vitest run` 154 files / 2091 tests passed.

**Closure, by code path**

- **Ranking.** `plugin/src/ui/panel-model.ts` now pushes the lost-edit layer before trouble
  (`if (dropped > 0) layers.push({ text: droppedNote(dropped), … }); if (trouble) layers.push(TROUBLE_SENTENCE[trouble]);`)
  and returns `{ lead, rest, text, tone, title }` with `text = ${lead}${rest}`. `lead` is non-empty only
  when a lost edit exists, so only lower-priority text can ever be cut.
- **Flex structure.** `plugin/src/ui/panel.html`: `.rail-sentence { display: flex; max-width: 400px;
  overflow: hidden }`, `.sentence-lead { flex: 0 0 auto }`, `.sentence-rest { min-width: 0; overflow:
  hidden; text-overflow: ellipsis; white-space: nowrap }`, with the two spans written by
  `panel-ui.ts` `render()` (`sentenceLead.textContent = view.lead` / `sentenceRest.textContent = view.rest`).
  The lead cannot shrink, so the ellipsis can only reach the tail.
- **Acknowledge.** `panel-ui.ts` `sentence.onclick = () => { activityView.acknowledgeFailures(); render(); }`
  → `panel-activity-view.ts` `acknowledgeFailures() { this.failures.acknowledge(); }` →
  `panel-view-state.ts` `acknowledge(): void { this.keys.clear(); this.overflow = 0; }`. The chip and the
  orb both read `failures.unresolvedCount`, so both clear; the next failure re-adds a key and re-arms.
- **Baseline.** `scripts/comment-hygiene-baseline.json` is absent from the commit
  (`git show --stat 3472da4` lists 16 files, none of them the baseline) and the working copy is
  byte-identical to `ce6f811`. Nothing was laundered and nothing was retired.

**Mutations I ran (all restored byte-exact)**

| mutation | result |
|---|---|
| rank trouble before the lost edit (the original defect) | 2 fail — the pure priority case, and Chromium: `expected 'Broker not running — run figma-agent …' to be '2 edits lost while offline'` |
| `.sentence-lead` → `flex: 0 1 auto` with its own ellipsis | 2 fail — including Chromium `a lost edit may never be ellipsed away: expected true to be false` |
| no-op the acknowledge click, patched into the BUILT `plugin/ui.html` | 1 fail — `the chip clears once the user has seen it: expected { hidden: false, text: '1' } to deeply equal { hidden: true, text: '' }` |
| drop `text-overflow: ellipsis` from `.rail-sentence` | 2 fail — but both are source-string assertions; rendering is unchanged (see F2) |

`tests/panel-ui-browser.test.ts` is the right kind of test: it drives the built artifact, stubs `WebSocket`
inert and records outbound events, so a broker running on the machine cannot change the result — and it
watches 2.6 s of ticks to prove the hug does not re-post an unchanged width.

## Risk level

**Low.** No trust-boundary change: `resolveViewportRequest` / `clampRailWidth` are untouched and still
reject strings, objects, `null`, `NaN`, `Infinity`, negatives and 100 000; height is never read off the
wire. No schema, protocol or exported CLI contract changed. `SYNC_DONE { commit }` semantics and
`main.ts`'s `changesSinceCommit = 0` are unchanged, and the browser test now pins the refusal path
(`commit: false`, button stays). Remaining risk is presentational and confined to the panel iframe.

## Findings

None blocking. In rough priority:

**F1 — the acknowledge affordance is pointer-only (answers question 1: tooltip alone is not enough).**
`#fga-sentence` is a `<button>` whose accessible name is its own status text; the only statement of what
activating it does is `acknowledgeHint(n)` appended to `title` in `render()`. `title` is hover-only for a
mouse and is announced inconsistently as a description by screen readers, so a keyboard user focuses a
control whose name is a status string and gets no hint that it acknowledges anything. Worse, the button
is focusable and clickable even at zero failures, where the click is a no-op — the same "a control that
does nothing" argument used one file over to demote the orb to `role="img"`. The README documents the
gesture, which helps, but the product should not depend on that.
Smallest fix, still one row and no expanded state: make `#fga-failure-count` the button — it already
carries `aria-label="N unresolved failures"` and is already hidden at zero, so the control exists exactly
when there is something to acknowledge and labels itself; extend its label with `acknowledgeHint(n)`. If
you keep the sentence clickable instead, give it an `aria-description` from `acknowledgeHint(n)` and stop
it being focusable at zero — but avoid toggling `disabled` on a focused element (focus jumps to `<body>`).

**F2 — remove the inert `text-overflow: ellipsis` from `.rail-sentence` (answers question 2).**
`text-overflow` applies to block containers; `.rail-sentence` is `display: flex`, so the declaration has
no rendering effect — proven: with it deleted, the Chromium geometry test still measures
`restClipped === true` for the over-long sentence, i.e. the ellipsis comes entirely from `.sentence-rest`.
What does fail are two source-string assertions (`tests/figma-plugin-panel.test.ts:95` and
`tests/panel-craft-regression.test.ts:75`), which means the suite is currently guarding a no-op and
blocking its removal. Delete the declaration and both assertions. Keep `overflow: hidden` (it clips a
lead longer than the cap) and `white-space: nowrap` (both spans inherit it).

**F3 — the browser test trusts a committed build output.** `tests/panel-ui-browser.test.ts` reads
`plugin/ui.html`, and `tests/plugin-build-id.test.ts` only proves `code.js` + `manifest.json` + `ui.html`
are self-consistent — not that they were built from the current `plugin/src`. A stale but self-consistent
artifact would let the new test validate old wiring. Worth a freshness check as a follow-up.

**F4 — carried forward from the first pass, unchanged in this commit.**
- `renderSync()` calls `replaceIcon(syncRailBtn, …, run !== null)` on every 1 s tick and `replaceIcon`
  does `replaceChildren`, so the fresh SVG restarts `animation: fga-spin 1s linear infinite` from 0 each
  tick. Re-icon only when the name or spinning state changed.
- `render()` assigns `sentenceLead.textContent` / `sentenceRest.textContent` unconditionally each second
  inside an `aria-live="polite" aria-atomic="true"` region, so screen readers may re-read the row every
  tick. Guard the assignments on a real change.
- Target pinning is unreachable with one connected file (`targetRailBtn.hidden = peersCount <= 1`, and no
  CLI command sends `SET_TARGET`/`CLEAR_TARGET`), so a pin set with two files open cannot be cleared once
  the other disconnects.
- `main.ts:140-143` still explains `FILE_INFO` as feeding "the panel's Context block", and justifies the
  undebounced `selectionchange` by keeping the panel current — the panel no longer shows selection. The
  surviving reason is that it enriches `PLUGIN_HELLO` / `figma-agent status`.
- `docs/images/adaptive-agent-rail.gif` still depicts the pre-change surface under a caption that now
  claims one hugging row.
- `targetButtonLabel` (`panel-model.ts:133`) and `activity-feed.ts`'s `timeAgo` / `formatTimestamp` /
  `diffRowKeys` have no production caller left; only tests reach them.

None of F1–F4 changes behavior under load or crosses a trust boundary; all are fine as follow-ups.

## Verdict

**APPROVE.** The lost-edit note is now structurally un-hidable and proven so in Chromium, the baseline
file is untouched, the acknowledge round-trip is real and tested against the built artifact, and the
panel's four shared linters plus the `ease-design@0.5.0` pin are intact. F1 and F2 are the two I would
land next — both are small, and F2 also removes a test that guards nothing.

**Not verified:** byte-identity of a fresh rebuild (`npm run build` is blocked in my environment; the
artifacts were checked structurally and by the build-id gate), the live Figma host title at 240 px, real
`figma.ui.resize` behavior on canvas, light-theme rendering, and end-to-end sync against a live broker.

*Posted by the installed review-pr skill at 2026-09-03T09:55:39Z*

PR #109, opened 2026-09-03
feat(cli): craft bundle v1.1.0 and the five group-6 CLI behaviours

jangtrinh:
## Why

The VSF-PCP campaigns (backlog group 6, 2026-09-03) patch five figma-agent behaviours with project-side hooks and helpers, and the repo's bundled `es-figma-craft` skill lags the canonical v1.1.0 (three new references, gates 9–10, Class G gotchas). Plugin and CLI are meant to stay generic — those behaviours belong in the tool.

## What

One commit per item, all additive on the wire (`PROTOCOL_VERSION` unchanged):

1. **`skills/es-figma-craft` = canonical v1.1.0** — copied wholesale (`screen-composition`, `drawer-decision-tree`, `section-presentation`, operating gates 9–10, Class G). The only content the old bundle had that the canonical lacks is the `es-figma-craft` spelling of the skill name (canonical: `es:figma-craft`).
2. **Pre-dispatch plugin admission** — every command the broker classifies as a mutation (everything outside the safe-read allowlist, so EXEC_JS/BATCH/AUDIT_DS too) waits up to 60 s for a plugin matching `--file`/`--instance` (else any) to register before it is sent; announced once on stderr; a plugin that never comes → `E_NO_PLUGIN` naming the bound and the opt-out. Global `--no-wait` restores immediate dispatch. Safe reads, `--target-file-key` requests and broker-terminal commands never wait. Replaces the project's `status --wait --timeout 60 &&` PreToolUse hook.
3. **`resolve-component --name <n> [--page <p>]`** — read-only (rides `SCAN_DESIGN_SYSTEM`, a safe read that bypasses the mutation FIFO); exact-after-trim, case-insensitive match; `--page` filters first; otherwise a tie is broken only when exactly one hit sits on a design-system-looking page (`preferred: "design-system-page"`); `E_AMBIGUOUS` lists every candidate; `E_NOT_FOUND` when none. Plugin side: each scan entry now names its page (`{id,name}` or `null`).
4. **exec-js lint** — two new warnings (`find-all-without-visible-filter`, `component-properties-on-set`; the sync `mainComponent` getter was already one) and `--strict`, which promotes warnings to a refusal. Rules gain a `requires` clause matched against the raw source (node types are spelled inside string literals).
5. **`export-png --assert <script.js>`** — the assert runs first as a plugin-enforced read-only EXEC_JS (the plugin's own guard refuses a script that writes; broker admission still treats it as a mutation, so the public `--read-only` refusal on EXEC_JS is unchanged); the PNG is written only on a truthy / `{ok:true}` result; `E_ASSERT_FAILED` otherwise, no file. Craft gate 9 in one command.
6. **`changes --png <dir> [--scale]`** — an after PNG per unique node in the listed window (read-only EXPORT_PNG), every path in the JSON; a before only when a prior export in `<dir>` predates the earliest edit (`beforeSource: "prior-export"`), otherwise `before: null` with the reason; deleted / vanished nodes listed in `skipped`; `--limit` bounds the count.

## Evidence

Gates after the last edit (this branch, local): `npm run typecheck` 0 · `npm run build` 0 · `npm test` 0 — 158 files / 2148 tests (main: 153 / 2080) · `npm run lint:comments` 0 · `npm run emit:skill` 0 (committed SKILL.md == emitter, drift test green).

Test-first, each with an observed red: plugin-admission (module absent; and the `--no-wait` path shows the pre-fix `E_FILE_KEY_UNAVAILABLE` against a real scratch broker), resolve-component + serializer page field (missing exports), lint rules (3 positives + `--strict` red), export-png assert (16/18 red), changes `--png` run() (3 red against HEAD `changes.ts` via file backup).

Attacked: duplicate names across pages / on one page / page-less candidates; plugin never registering (fake clock) and registering only after the mutation was issued (real broker); assert scripts that throw, mutate (E_READONLY_VIOLATION), fail lint, or return `{ok:false}`; a deleted node and a vanished node in the changes window; prior after.png older / newer than the edit.

## Not in this PR

- No end-to-end run against the live Figma file (the owner's plugin on 9410 was not used); the real 60 s admission bound was exercised only with an injected clock.
- `resolve-component` reuses the full design-system scan rather than a dedicated wire command — one scan per call; a dedicated `RESOLVE_COMPONENT` would be a follow-up if scan cost matters on the real file.
- `broker-client.ts` (395 lines) and `changes.ts` (259) were already over the 200-line guideline; the new logic lives in new modules, only the wiring landed in them.


jangtrinh (commented):
## Summary

Reviewed at `origin/feat/craft-v1.1.0-and-cli-group-6` in a clean worktree (33 files, +1684/−90 — matches the PR body). Six well-separated commits, additive on the wire, tests that actually exercise behaviour rather than shape.

Independently re-run after checkout: `npm run typecheck` 0 · `npm run build` 0 · `npm test` **158 files / 2148 tests pass** · `npm run lint:comments` clean (0 new citations, `scripts/comment-hygiene-baseline.json` untouched). `npm run build` leaves the tree clean, so the committed `plugin/code.js` / `plugin/ui.html` match their sources. `PROTOCOL_VERSION` is still `1` (`shared/protocol.ts:5`); `E_AMBIGUOUS`, `E_NOT_FOUND`, `E_ASSERT_FAILED` and the `page` field on scan entries are purely additive.

Two things I verified specifically because they are the load-bearing claims:

- **The mutation gate is not weakened.** `broker-daemon.ts:1170` recomputes `isReadOnly = cmd !== undefined && isBrokerSafeRead(cmd)`; the envelope flag arrives as `_readOnly` and is referenced nowhere in the function. So `pluginEnforcedReadOnly` cannot bypass admission, the per-file FIFO, or the per-file gate. `resolveWireReadOnly` (`broker-client.ts:125-130`) refuses the internal option on any non-`EXEC_JS` command, and the option is reachable from exactly one call site (`export-png.ts:41`) — never from a CLI flag. The stage-4 ruling refusing public `--read-only` on `EXEC_JS` stands.
- **The craft bundle is byte-equal to canonical** (`diff -rq skills/es-figma-craft ~/.claude/skills/es-figma-craft` → no output). The `es-figma-craft` spelling survives where it is actually load-bearing: the *directory* name, which `install-skill.ts:16` and `tests/install-skill.test.ts` key on. The frontmatter `name: es:figma-craft` is read by no code or test in this repo and matches the installed canonical, so the lost spelling is cosmetic. No file keys, node ids, local paths, or credentials in the three new references.

## Risk level

Medium. Nothing corrupts a file or breaks a contract, but two defects make a new guarantee weaker than advertised.

## Findings

### Important

**1. The admission wait can be satisfied by a plugin the broker will not route to — `cli/src/transport/plugin-admission.ts:53`, `cli/src/transport/plugin-wait.ts:40-51`**

The wait matches `--file` with `fileIdentity(null, name)` slug equality (`file-identity.ts:20-23`: lowercase, every non-alphanumeric run → `-`). Dispatch routes with `fileMatches(actual, filter, exact=true)` (`shared/file-match.ts:5-9`: trim + lowercase, punctuation-significant). Slug equality is strictly looser, so the mismatch is one-directional: the wait can succeed where the broker refuses, never the reverse.

Scenario: `figma-agent set-text --file "VSF PCP" …` while the plugin reports `figma.root.name = "VSF - PCP"`. Both slug to `vsf-pcp`, so the very first poll reports "registered", `needsPluginAdmission` is satisfied, the request is dispatched — and the broker's exact route filter finds no plugin and answers `E_FILE_KEY_UNAVAILABLE`. The user gets exactly the error this PR exists to remove, with no wait and no `E_NO_PLUGIN` hint. The same happens for any name differing only in punctuation or spacing (`Design-System` vs `Design System`, an en dash vs a hyphen).

This is the failure mode `plugin-wait.ts:13-19` already warns about ("these two halves of one command must never disagree about which file was meant") — the comment reasons about the deep-link marker, but the half that decides dispatch here is `fileMatches`.

Fix: give `waitForPlugin` an opt-in exact matcher and have `awaitPluginAdmission` pass the broker's own comparison, `fileMatches(p.fileName, fileFilter, true)`, leaving `status --wait`'s slug semantics untouched. `tests/plugin-admission.test.ts:177` covers "keeps waiting while a different file is registered" but only with names that differ wholesale; add the punctuation case.

**2. `changes --png` discards the record of work it already did — `cli/src/commands/changes-png-export.ts:110-114`, `cli/src/commands/changes.ts:241-243`**

An aborting code (`E_TIMEOUT`, `E_NO_BROKER`, `E_NO_PLUGIN`, `E_VERSION_MISMATCH`) rethrows out of `exportChangePngs`, out of `run()`, and the command exits 1 with only that error.

Scenario: a window of 8 nodes; node 5's `EXPORT_PNG` hits `E_TIMEOUT`. Nodes 1–4 already have a new `<stem>.after.png` on disk, and up to 4 prior `.after.png` files were already renamed to `.before.png` (line 121). None of those paths, and none of those renames, appear anywhere in the output. The caller cannot tell which nodes were exported or that the directory was reshuffled — work performed, zero record, which is the case the "nothing vanishes silently" rule exists for.

Fix: catch the aborting error, and surface the partial result — either a `png:{…, aborted:{code,message}}` block, or the already-written paths inside the error message. The existing test (`tests/changes-png-export.test.ts`, "a transport failure … aborts the whole export") only fails on the first node, so the partial-work case is untested.

### Suggestion

**3. The `--assert` doc comment overstates the plugin guard — `cli/src/commands/export-png.ts:5-8`.** The header says "the plugin's own read-only guard refuses a script that writes". `plugin/src/main/main.ts:440-456` shows detection is post-hoc: the script has already run and already mutated; the guard throws `E_READONLY_VIOLATION` afterwards and the write is sealed into its own undo step. The catalog entry (`cli/src/command-catalog.ts:144-146`) states this honestly — the module header and the PR body do not. Related: `readonly-guard.ts:14-31` only attributes a change while `activeCount === 1`, so a mutating assert racing another dispatch is never flagged and the PNG *is* written. Worth one clause in the `--assert` help.

**4. `--read-only` does not block `--assert` — `broker-client.ts:125-130`.** With `pluginEnforced` true the global flag is bypassed entirely, so `export-png --read-only --node … --assert bad.js` runs an `EXEC_JS` that can land a write in a session the caller declared read-only. Either refuse `--assert` under a global `--read-only`, or name the exception in the flag help (which already frames `--read-only` as a partial hardening).

**5. The design-system page heuristic can prefer an archive — `cli/src/commands/resolve-component.ts:26`.** `DESIGN_SYSTEM_PAGE_RE` matches `component` anywhere, so "Deprecated Components" or "Component Graveyard" reads as the library home. With one candidate there and one on a screens page, the archived copy wins and the reply asserts `preferred:"design-system-page"` — a confident wrong answer, which is worse than `E_AMBIGUOUS`. Consider excluding an archive token (`deprecated|archive|old|graveyard`), or treating a hit on such a page as no preference at all. The `\bds\b` tightening is the right call and correctly rejects "Cards"/"Threads".

**6. `resolve-component` scan cost — follow-up, not blocking.** Measured on the owner's file (21 pages, ≥418k nodes, pages resident): `scan-design-system` = **1.52 s** end-to-end, 2,061 components, 498 KB. So each call blocks the plugin main thread for up to ~1.5 s of atomic `figma.root.findAllWithCriteria` (`serialize-node.ts:74`, no yields) and ships half a megabyte to resolve one id. That is per call and on demand, not a boot stall, so it does not undo the boot-perf campaign — but it is a real freeze on an interactive canvas. Two notes: (a) the new `page` field grows *every* `SCAN_DESIGN_SYSTEM` reply for every consumer, ~2k entries × ~40 B ≈ +80 KB (~16%); (b) on a cold plugin the first call also pays `loadAllPagesAsync` across 21 pages, which is unmeasured — though that is pre-existing for `scan-design-system`. The bounded alternative (page-scoped `findAllWithCriteria({types:['COMPONENT','COMPONENT_SET']})` per page with a yield between pages, or a dedicated `RESOLVE_COMPONENT` wire command returning only matches) is the right follow-up if this lands in a hot loop. `pageOf` reuse is correct and cheap — one shared O(depth) walk, no second implementation.

**7. `component-properties-on-set` false positives are untested — `exec-js-lint.ts:109-114`, `:246`.** `requires` is matched against the RAW source, so a script that legitimately reads `instance.componentProperties` and merely *mentions* `COMPONENT_SET` — in a comment, or in `findAllWithCriteria({types:['COMPONENT_SET']})` — warns. The trade-off is documented in code, but the "false-positive guard" test (`tests/exec-js-lint.test.ts:147-150`) uses `componentPropertyDefinitions`, which the pattern `\.\s*componentProperties\b` cannot match under any circumstances — it does not exercise the actual FP mode. Add the mixed-script case. Note that `--strict` has no per-rule waiver, so such a script can never pass `--strict`. `find-all-without-visible-filter` is cleanly guarded (`findAllWithCriteria`, `findOne` and masked strings all verified).

**8. Filename stems can collide in principle — `changes-png-export.ts:51-53`.** Every non-`[A-Za-z0-9_.-]` char maps to `-`, so ids differing only in separator (`:` vs `;`) produce one stem. Figma's id grammar makes that unreachable today, but nothing enforces the grammar; a short hash suffix would close the class. Path safety itself is fine — `/` is stripped, so nothing can be written outside `--png <dir>`, and `changes.ts` resolves the dir to an absolute path.

**9. Flag-ordering trap for the two new boolean flags — `cli/src/arg-parse.ts:26-29`.** A value-less flag followed by a non-`--` token swallows it, so `exec-js --strict script.js` loses the positional and fails with a misleading "missing file". Pre-existing parser behaviour, not introduced here, but `--strict` and `--no-wait` widen the surface; the catalog examples should show them after positionals.

**10. Declared and accepted:** `broker-client.ts` (395) and `changes.ts` (259) were already over the 200-line guideline and only took wiring; new logic lives in new modules. `AUDIT_DS` now waits up to 60 s despite being read-shaped — correct per `isBrokerSafeRead`, and the `--no-wait` help says so.

## Verdict

**Request changes** — findings 1 and 2 are Important. Both are small, local fixes with a test each: use the broker's own file comparison in the admission wait, and report the PNGs already written when the export aborts. Everything else is a suggestion and can land as follow-ups.

*Posted by the installed review-pr skill at 2026-09-03T10:01:02Z*

jangtrinh (commented):
## Summary

Re-reviewed at head `784c230` in a fresh throwaway worktree. Both Important findings from the previous round are fixed correctly, each pinned by a test that fails when the fix is reverted. The `E_AMBIGUOUS` suggestion landed too.

Gates re-run independently after checkout: `npm run typecheck` 0 · `npm run build` 0 · `npm test` **158 files / 2153 tests pass** · `npm run lint:comments` clean (0 new, baseline untouched) · `tests/skill-emitter-drift.test.ts` 4/4 — the committed `skills/figma-agent/SKILL.md` matches the emitter. `npm run build` leaves the tree clean, so the committed `plugin/code.js` / `plugin/ui.html` still match their sources.

## Risk level

Low. No blocking defect remains.

## Verified fixes

**F1 — the wait and dispatch now share one predicate (`8d77729`).** Not a re-implementation: `cli/src/transport/plugin-admission.ts:58` calls

```
waitForPlugin({ matchFile: (actual, filter) => fileMatches(actual, filter, true), ...opts })
```

importing `fileMatches` from `shared/file-match.ts` — the same function object the dispatch path uses at `cli/src/transport/plugin-registry.ts:365` (`fileMatches(e.scene.fileName, f, opts?.exact === true)`), reached with `exact: true` for the `--file` flag (`cli/src/transport/route-filter.ts:51`). `plugin-wait.ts:56` keeps `slugEquals` as the default parameter, so `status --wait`'s slug semantics — paired with the deep-link marker lookup — are untouched, and the diff to that file is purely additive.

Mutation check: reverting line 58 to `waitForPlugin(opts)` makes `tests/plugin-admission.test.ts:195` fail ("`--file "VSF PCP"` is NOT satisfied by a plugin named "VSF - PCP"") — 1 failed / 14 passed. Restored byte-exact. The companion test at `:204` proves the trimmed, case-insensitive spelling still IS satisfied, so the fix did not simply tighten the matcher into uselessness.

**F2 — partial work is reported, and no rename precedes a successful export (`c32e1b5`).** `cli/src/commands/changes-png-export.ts:115-125`: a non-`CliError` still rethrows (a bug, not a transport fact), a non-aborting code still becomes a `skipped` entry, and an aborting code now records `error = {code, message, atNodeId}` and `break`s. Line 139 returns it beside `exported`/`skipped`. `cli/src/commands/changes.ts:262-266` prints the one JSON object and exits 1, so a caller sees both the partial work and that the window is not fully covered.

Ordering is correct for the failure this addresses: the `EXPORT_PNG` round-trip is awaited at `:112` and the base64 validated at `:126` **before** `resolveBefore` / `renameSync` at `:130-131`, so a node whose export fails never leaves a half-done rename. `tests/changes-png-export.test.ts` pins this directly — the mid-window test seeds a qualifying `2-2.after.png`, fails node `2:2` with `E_TIMEOUT`, and asserts the prior file is byte-unchanged, `2-2.before.png` does not exist, node `3:3` was never attempted, and node `1:23`'s PNG is on disk and named in the result.

Mutation check: replacing the `error = …; break;` pair with `throw err` fails 3 tests in that file (12 passed), including the `run()`-level test that parses real stdout and asserts `exit(1)`. Restored byte-exact; worktree confirmed clean at `784c230`.

The behaviour is documented where it is contractual — the `changes` row of the command table and the emitted `skills/figma-agent/SKILL.md:98` both spell out `png.error:{code,message,atNodeId}` and the non-zero exit, and the drift test is green.

**Suggestion landed (`784c230`).** `CliError` gains an optional readonly `candidates` (`protocol-helpers.ts:29-38`), `json-out.ts:43` prints the key only when set, and `resolve-component.ts:90-92` attaches it only when the list is non-empty — so `E_NOT_FOUND` (candidates `[]`) is unchanged. Additive for every other error.

## Remaining suggestions

None blocking; all carried forward from the first round unless marked new.

**1. (New) Spread order makes the exact matcher overridable — `plugin-admission.ts:58`.** `{ matchFile: …, ...opts }` puts the spread last, so a caller passing `matchFile` (or even an explicit `matchFile: undefined`, which would fall through to `slugEquals` via the default parameter at `plugin-wait.ts:56`) silently reverts the fix. No live caller does — `awaitPluginAdmission` has exactly one call site, `broker-client.ts:376`, which builds the literal without that key — so this is latent, not a defect. Writing `{ ...opts, matchFile: … }` makes the invariant unbreakable for one character's cost.

**2. (New) `candidates` is not announced in the command table.** The `resolve-component` row still says only "E_AMBIGUOUS listing every candidate", which now reads as prose-only while the error JSON also carries structured `candidates:[{id,name,type,page}]`. One clause in `command-catalog.ts` would let an agent discover it without reading the source.

**3. A write failure after the rename still loses the record — `changes-png-export.ts:131-132`.** `renameSync` precedes `writeFileSync`, so an `ENOSPC` / `EACCES` / dir-removed-mid-run failure throws out of the loop uncaught: the rename already happened and the partial record is dropped again. Much narrower than the transport case now fixed, and the resulting state is recoverable (the next run finds `.before.png` with its old mtime and re-exports), but the same `error`/`break` treatment would close it.

**4. The `--assert` doc comment overstates the plugin guard — `cli/src/commands/export-png.ts:5-8`.** It says the guard "refuses a script that writes"; `plugin/src/main/main.ts:440-456` shows detection is post-hoc — the script already ran and already mutated, and the write is sealed into its own undo step. The catalog entry (`command-catalog.ts:144-146`) states this honestly; the module header does not. Related: `readonly-guard.ts:14-31` only attributes a change while `activeCount === 1`, so a mutating assert racing another dispatch is never flagged and the PNG *is* written.

**5. `--read-only` does not block `--assert` — `broker-client.ts:125-130`.** With `pluginEnforced` true the global flag is bypassed, so `export-png --read-only --assert bad.js` can land a write in a session that declared read-only. Refuse the combination, or name the exception in the flag help.

**6. The design-system page heuristic can prefer an archive — `resolve-component.ts:26`.** `DESIGN_SYSTEM_PAGE_RE` matches `component` anywhere, so "Deprecated Components" reads as the library home; with one candidate there and one on a screens page the archived copy wins and the reply asserts `preferred:"design-system-page"` — a confident wrong answer, worse than `E_AMBIGUOUS`. Consider excluding an archive token, or treating such a hit as no preference.

**7. `resolve-component` scan cost — follow-up.** Measured on the owner's file (21 pages, ≥418k nodes, pages resident): `scan-design-system` = 1.52 s end-to-end, 2,061 components, 498 KB. Each call blocks the plugin main thread for up to ~1.5 s of atomic `figma.root.findAllWithCriteria` (`serialize-node.ts:74`, no yields) and ships half a megabyte to resolve one id. Per call and on demand, so it does not undo the boot-perf work — but the new `page` field also grows *every* `SCAN_DESIGN_SYSTEM` reply for every consumer (~2k entries × ~40 B ≈ +80 KB). A page-scoped scan with a yield between pages, or a dedicated `RESOLVE_COMPONENT` wire command, is the fix if this ever lands in a loop.

**8. `component-properties-on-set` false positives are untested — `exec-js-lint.ts:109-114`, `:246`.** `requires` matches the RAW source, so a script legitimately reading `instance.componentProperties` that merely mentions `COMPONENT_SET` (a comment, or `findAllWithCriteria({types:['COMPONENT_SET']})`) warns — and `--strict` has no per-rule waiver, so it can never pass. The "false-positive guard" test uses `componentPropertyDefinitions`, which the pattern cannot match under any circumstances; add the mixed-script case.

**9. Stem collisions are unreachable but unenforced — `changes-png-export.ts:51-53`.** Ids differing only in separator (`:` vs `;`) map to one stem. Figma's id grammar rules that out today; a short hash suffix would close the class. Path safety itself is fine — `/` is stripped, nothing can be written outside `--png <dir>`.

**10. Flag-ordering trap — `cli/src/arg-parse.ts:26-29`.** A value-less flag followed by a non-`--` token swallows it, so `exec-js --strict script.js` loses the positional and fails with a misleading "missing file". Pre-existing, but `--strict` and `--no-wait` widen the surface; show them after positionals in the catalog examples.

## Verdict

**Approve.** Both blockers are fixed at the root with the broker's own predicate and a partial-result contract, each proven load-bearing by reverting it. The ten items above are non-blocking; items 1 and 2 are one-line follow-ups worth folding in before merge if convenient.

*Posted by the installed review-pr skill at 2026-09-03T10:14:07Z*

PR #114, opened 2026-09-03
feat(cli): add context command for budgeted per-node code context

jangtrinh:
## Why

The README concedes "no code context" to Figma's official MCP. That concession was half wrong: the Plugin API exposes, on every plan, a node's Inspect-panel CSS declarations (`getCSSAsync`), the variables and styles it binds, its text and component properties. Measured live on the owner's 418k-node Free file: `getCSSAsync` costs 7–8 ms per node and answers for FRAME, TEXT and INSTANCE alike. What was missing was a command that reads it honestly — bounded before the wire, counting everything it leaves out.

## What

- New typed safe-read wire command `GET_CONTEXT` (`COMMANDS`, `BROKER_SAFE_READ_COMMANDS`, explicit `COMMAND_TIMEOUTS`); it never enters the per-file mutation FIFO and takes no undo step.
- `figma-agent context [nodeId] [--budget KB=64] [--depth N] [--no-css] [--timeout ms]`: breadth-first walk of one subtree, byte-budgeted in the plugin with `utf8ByteLength` (no `TextEncoder` in the sandbox), `getCSSAsync` in batches of 16 with a macrotask hop between batches, a soft deadline that returns a partial with counts instead of an empty timeout.
- Every reply carries `schema: "context/1"`, per-node records (identity, layout summary, CSS verbatim, bindings, styles, text, component properties, `isAsset` collapse with `collapsed.{descendants,types}`), identity tables `refs.{variables,styles,components}` resolved once per id with `modeCount` and no mode-0 value, and a budget block with the conservation law `visited = emitted + omitted.budget + omitted.deadline`, `partial`, a stateless `frontier[]` cursor and `complete`.
- A refused structural read (a page whose `children` getter throws) leaves a counted trace; a DOCUMENT target is refused; a valueless `--depth/--budget/--timeout` is refused rather than defaulted.
- `--format` is reserved and refused, not silently accepted.

## Evidence

Gates after the last edit (45e8516): typecheck 0 · build 0 (committed plugin artifacts fresh) · lint:comments 0 (0 new) · emit:skill drift green · `npm test` 166 files / 2251 passed (+80 for the five new modules across two review rounds).

Live on the owner's file (plugin hot-reloaded from this build): a 121-node frame answers in 773 ms wall (`--no-css` 124 ms), `--budget 8` emits 15 of 71 visited with 56 counted as omitted and named in `frontier`, re-issuing `context <frontier[0].id>` completes; an INSTANCE with library variables resolves 13/13 bound ids into `refs.variables`; CSS arrives verbatim with token fallbacks where Figma emits them (`border: 4px solid var(--color-global-indigo-indigo-10, #E0E7FF)`).

Review: stage-4 adversarial review, two rounds (round 1: one Critical + six Important; round 2: three Important — an unreadable child identity shipping as a ghost record, an unconditional deadline promise, uncapped flags), every finding fixed test-first with its RED reproduced by the reviewer; final verdict Approve. Mutation testing: 18/18 targeted wrong-fact mutants killed by the existing suite, 0 survivors. Caps: `--budget` refused above 512 KB (the wire's chunk seam), `--timeout` clamped to 120 s and reported as `budget.timeoutMs`, both re-enforced in the plugin. Live after round 2: `budget.timeoutMs` reads 120000 for `--timeout 3600000`; `--budget 9999` refused before any broker traffic.

## Not in this PR

Dedup (`--dedup`) and design intent (annotations, devStatus, dev resources) — next phase; the README/deep-dive concession correction lands after that; a cursor for frontier entries beyond the first 50 is a follow-up.


jangtrinh:
## Stage-4 adversarial review — round 2

**Summary.** Every finding from round 1 (C1 + I1–I6) and both controller rulings are fixed in the committed diff, each behind a test that fails against the pre-fix code — I named them below and re-ran all five context files (70/70) plus the six drift/freshness gates. Three new findings: one reproduced silent hole in the same invariant C1 was about (an identity read that refuses leaves no counted trace inside `complete: true`), one unconditional doc promise the code cannot keep (ref resolution runs outside every bound), and two uncapped wire numbers that diverge from this CLI's own clamp precedent.

**Risk level: medium.** The wire path, the conservation law, the refusal semantics and the drift gates are all sound; the residual risk is concentrated in one 3-line record-builder gap and in two doc sentences that overclaim.

**Gates re-run here** (per the read-only mandate: no `npm test`, no `npm run build`)
- `npm run typecheck` → 0.
- `npx vitest run tests/context-{command,executor,node-record,refs,walk}.test.ts` → **5 files / 70 tests passed** (confirms the PR's `+70`; `2171 + 70 = 2241` is arithmetically consistent but was not independently re-run).
- `npx vitest run tests/{plugin-build-freshness,skill-emitter-drift,mutating-commands,orb-command-state,broker-client-read-only,plugin-admission}.test.ts` → 6 files / 50 passed. `plugin-build-freshness` re-runs esbuild in memory and diffs the committed artifacts, so the PR's "build 0" is **proven for the plugin half** (`cli/dist` is untracked, so nothing else is at stake).
- `npm run lint:comments` → `clean (53 file(s) with baselined citations, 0 new)`. It also warns `7 baseline entries are stale` — pre-existing, not this PR.
- `166` test files on disk ✓. No `any`, no `as any`, no `@ts-expect-error`, no `eslint-disable`, no `console.`, no TODO/FIXME in the five new source files or their tests.
- No DOM/Node global in the four new `plugin/src/main` files — only `Date.now` and `setTimeout`; `plugin/tsconfig.main.json` (DOM-less) typechecks clean, which is the stronger proof.

---

### Round-1 findings — verified fixed, with the test that fails without the fix

| # | fix in the committed diff | test that goes red without it |
|---|---|---|
| **C1** | `childrenOf(node) → {children, refused}` at `plugin/src/main/context-node-record.ts:105-116`, consumed at **all three** call sites: `buildContextRecord:208-217` (sets `childrenError`, `childCount: null`, `incomplete`), `countCollapsed:141-142` (`readErrors`), `context-walk.ts:140-143` (`childCountOf → number \| null`) | `context-node-record.test.ts` "a throwing children getter is NOT a childless node"; "an asset collapse counts the refusals it hit instead of undercounting silently"; "a node with no children FIELD at all is honestly childless"; `context-walk.test.ts` "counts a refused children read as partial, and its frontier childCount is null" |
| **I1** | `numericFlag` `cli/src/commands/context.ts:115-120`; `resolveContextParams` hoisted to `contextTarget`'s first statement (`:91`) so a refused flag costs no `GET_SELECTION`; sub-1-byte budget refused at `:61-63` | `context-command.test.ts` "refuses a valueless --%s by name" (budget/depth/timeout); "refuses a --budget that floors to zero bytes before spending a round trip" |
| **I2** | `modeCount: Array.isArray(modes) ? modes.length : null` `context-refs.ts:94` | `context-refs.test.ts` "a modes read that refuses reports null, never a fabricated count of 0" |
| **I3** | `budget.refsBytes` `executor-context.ts:150`; catalog `command-catalog.ts:64-67` and `SKILL.md:39-40` now say `--budget` bounds the node records and `refsBytes` is reported separately | `context-executor.test.ts` "assembles refs from the walked records and reports both byte numbers" (`refsBytes` was `undefined`) |
| **I4** | `omitted.readError` gone; law `visited = emitted + omitted.budget + omitted.deadline` (`context-walk.ts:121-132`); `emitted: nodes.length` (`:266`); `partial` (`:237`, `:268`); un-buildable / un-serialisable records ship as `{id, readError}` (`:201-205`, `:212-223`) | `context-walk.test.ts` "visited === emitted + budget + deadline, and emitted === nodes.length"; "a record that cannot be built or serialised still ships as a minimal identified node"; "refuses to report accounting that does not add up" |
| **I5** | `expect(peak).toBe(8)` `context-walk.test.ts:253` — a sequential `for … await` rewrite now fails | (the assertion itself) |
| **I6** | `refuseNonSubtree` `executor-context.ts:89-96`, applied on **both** the id path (`:103`) and the selection path (`:110`) | `context-executor.test.ts` "refuses a DOCUMENT target — a document is not a subtree"; "allows a PAGE target" |
| **cssMs** (controller) | per-**batch** wall time `context-walk.ts:195-206`; per-record `cssMs` and the injected `now` are gone | `context-walk.test.ts` "reports cssMs as WALL time in the css batches, never the sum of overlapping latencies" (16 overlapping 10 ms reads → **20 ms across two batches, not 160**) + `cssMs <= walkMs`; "spends no css wall time at all under --no-css" |
| **frontier reason** (controller) | catalog `command-catalog.ts:71-74` + `SKILL.md:39-40`: `budget`/`deadline` → re-run on that id; `depth` → re-run with a larger `--depth`; `childCount: null` = unreadable, not a leaf | `skill-emitter-drift` (green) |
| suggestions | dead `countVerb: 'Read'` removed with the reason kept as a comment (`activity-sentence.ts:68-70`); plugin no longer emits `targetSource`; per-record `JSON.stringify` wrapped; `changesDuringWalk` → `changeBatchesDuringWalk` | `context-executor.test.ts` "resolves params.nodeId through the async getter" (`targetSource` undefined); "counts DOCUMENT-WIDE change batches during this dispatch, under a name that says so" (asserts the **old name is absent**) |

Drift gates all consistent: `COMMANDS` (`protocol.ts:167-173`), `BROKER_SAFE_READ_COMMANDS` (`mutating-commands.ts:21`), explicit `COMMAND_TIMEOUTS.GET_CONTEXT = 45_000` (`protocol.ts:705-711`), absent from `MUTATING_COMMANDS`, orb `searching`, activity verb pair, catalog row, emitted `SKILL.md`. `broker-daemon.ts:1170` still derives `isReadOnly` from the **command name**, so the CLI's `readOnly: true` assertion cannot launder a mutation past the FIFO — and `:1361-1366` still creates a job row + `JOB_STATE` for a read-only request, so an `E_TIMEOUT` stays pollable.

---

### Findings

#### Critical
None.

#### Important

**1 · An identity read that refuses leaves no counted trace, inside `complete: true` — the same class as C1, at the one reader C1 did not touch.**
`plugin/src/main/context-node-record.ts:151-157`. `id`, `name` and `type` are read through `safe()` and degraded to `''` / `''` / `'UNKNOWN'` with no `incomplete`, no error field, and no counter — while the *other two* readers in this module treat exactly this contingency as a refusal and say so in their comments: `childrenOf` at `:102-103` ("a fully invalidated reference refuses even that, and 'cannot tell' is a refusal, not an absence") and `idOf` at `context-walk.ts:145-146` ("a fully invalidated reference refuses even its id").

Reproduced (fixture: a child whose `id` and `type` getters throw `The node with id "9:9" does not exist`; walk run out-of-tree so nothing in the repo was touched):

```
nodes      => [ {"id":"1:1","name":"root","type":"FRAME","depth":0,"parentId":null,"childCount":1},
                {"id":"","name":"ghost","type":"UNKNOWN","depth":1,"parentId":"1:1","childCount":0} ]
accounting => {"visited":2,"emitted":2,"omitted":{"budget":0,"deadline":0},"partial":0,
               "frontier":[],"frontierTotal":0,"complete":true}
```

Wrong output: `complete: true, partial: 0` over a record the caller **cannot re-issue on** (empty id) and has no frontier entry for. `SKILL.md:38-41` tells agents `complete: false` is how they learn something is missing, so this is a wrong `complete`, not a cosmetic gap. Secondary effect of the same read: a TEXT whose `type` refuses skips the `characters`/`segments` branch (`:172-184`) silently — an agent renders no string.

Confidence: the code path is FACT (above); that the live host refuses an `id` read is an ASSUMPTION — asserted by this module's own two comments, not verified live. Either way one of the two has to change.

Smallest fix (3 lines, at the same site the other two refusals already use):

```ts
const rawId = safe(() => node.id);
if (typeof rawId !== 'string' || rawId === '') { record.readError = 'id could not be read'; incomplete = true; }
```

Test that fails without it: the fixture above, asserting `partial === 1` and `complete === false`. If the owner rules the trigger unreachable instead, the two comments at `context-node-record.ts:102-103` and `context-walk.ts:145-146` are overclaims and should go with it — this finding is then non-blocking.

**2 · The soft-deadline promise is unconditional in the docs, but ref resolution runs outside every bound.**
`cli/src/command-catalog.ts:66-68` and `skills/figma-agent/SKILL.md:75` state, without qualification: *"A soft deadline 2s inside `--timeout` means a big subtree answers with a partial AND its counts instead of a bare timeout."* The walk honours that (`context-walk.ts:187`), but `executor-context.ts:129-131` then calls `resolveContextRefs`, which is a **sequential** `await` per distinct variable id (`context-refs.ts:78-109`) with no deadline, no cap, and no omission bucket — the comment at `:125-128` concedes it runs outside the deadline and measures `refsMs`, but nothing bounds it.

Failing input: `context <ds-subtree> --budget 512` (the budget is uncapped — see #3). ~1 500 records at ~350 B; a token-documentation or DS-audit subtree where rows bind distinct tokens yields on the order of 10³ distinct ids; at the ~3 ms/lookup `context-executor.test.ts:126-139` itself models (same order as `getCSSAsync`'s measured 7-8 ms) that is 3–5 s against a 2 000 ms headroom. Wrong output: the wire timeout fires at 45 s and the caller gets `E_TIMEOUT` — recoverable only via `figma-agent job <id> --wait` — on the one command documented to answer with a partial instead. Both prior reports list `refsMs` on a real DS subtree as still unmeasured, so nothing rules this out empirically either.

Smallest fix, doc-only and honest: append to that sentence in both places — *"…unless ref resolution itself outruns the headroom: it runs after the walk, is not bounded, and reports `refsMs`."* Better: hand `resolveContextRefs` the deadline and mark the remainder `{unresolved: 'deadline'}` with a count, so the reply stays inside the bound it advertises.

**3 · `--budget` and `--timeout` are uncapped, unlike every other timeout in this CLI, and the plugin has no upper bound either.**
`cli/src/commands/context.ts:50-72` checks positivity only; `executor-context.ts:63-70` (`positive`) checks `> 0` only. The house pattern is the opposite: `cli/src/commands/scan-node.ts:56-58` clamps with `Math.min(requested ?? COMMAND_TIMEOUTS.EXEC_JS, EXEC_JS_MAX_TIMEOUT_MS)`, and `protocol.ts` records the policy in the very table this PR edits (`EXEC_JS: 30_000, // CLI --timeout may raise, capped at 120s`).

Failing input on the owner's 418k-node file: `context <pageId> --budget 1000000 --timeout 3600000 --no-css` → the wire carries `budgetBytes: 1_073_741_824`, `deadlineMs: 3_598_000`, and the plugin accepts both. Wrong output: the sandbox accumulates an unbounded `nodes[]` for up to an hour and either OOMs the plugin (the designer's plugin dies mid-session) or chunks hundreds of MB over the loopback WS — on the command whose stated selling point is that it is bounded before the wire. Smallest fix: refuse above a cap at both boundaries (refuse, not clamp — `executor-context.ts:61-62` already argues that a wire number is "validated at the boundary, never silently corrected"), e.g. `MAX_CONTEXT_BUDGET_KB` in `resolveContextParams` plus an upper bound in `positive()` so a wire client cannot route around the CLI.

#### Suggestions

- `plugin/src/main/main.ts:538` and `tests/context-executor.test.ts:3` still name `changesDuringWalk`; the field is `changeBatchesDuringWalk`. `:152` correctly asserts the old name is gone, so only the prose drifted.
- `countCollapsed` (`context-node-record.ts:126-146`) is synchronous and unbounded — no deadline check, no hop, no cap. A large `isAsset` subtree (an illustration frame marked as an asset) runs its whole descendant count inside one CSS batch, blocking the plugin thread past the soft deadline, and that cost is invisible in `estimatedBytes` (the `collapsed` block is a few bytes). Nothing vanishes — it is a latency bound, not a wrong fact.
- The catalog's `partial` sentence (`command-catalog.ts:70`) enumerates `cssError/mainComponentError/childrenError` but not the two other things that set it: `collapsed.readErrors > 0` and the minimal `{id, readError}` record. A caller reading only that sentence does not know a `{id, readError}` node counts as `emitted`.
- `visited` is never defined in the emitted docs. It counts nodes **enqueued**, not nodes in the subtree — the live PAGE run reported `visited 274` on a 23.7k-node page. The conservation-law sentence is exact but unreadable without that definition; one clause fixes it.
- `tests/context-command.test.ts:98-109` drives `run(parseArgs(...))`, whose `contextTarget` defaults to the real `runCommand`. Today every case throws before any transport, but if one of those refusals regresses the test performs **live broker I/O** — which is literally how the implementer's RED was produced ("the pre-fix run reached the owner's live plugin"). In CI that spawns a daemon. One optional `runner` param on `run()`, stubbed to throw, makes the failure mode a failed assertion instead.
- The budget check (`context-walk.ts:228-234`) stops the **whole** walk at the first record that would overrun rather than skipping it, so a single oversized record early in BFS order (a TEXT with a large `characters`, which has no length cap) strands the rest of the budget. Documented behaviour, but the recovery is only "ask for more bytes".
- PR body, Evidence: the `773 ms` wall time and the `13/13` ref resolution come from the **round-0** live pass, before the fix round; the post-fix pass re-measured `walkMs 675 / cssMs 652 / refsBytes 560` and did not re-take wall time or re-run the INSTANCE call. Everything else in that paragraph (`--budget 8` → 15/71/56, the DOCUMENT refusal, the valueless `--depth` refusal, the PAGE 274/64/210) is post-fix. Worth one clause, per this repo's own reporting law.

---

### Verdict: **Request changes** — blocking on Important #1 (or on removing the two comments that assert its trigger) and on the one-clause doc correction in #2. #3 is strongly recommended before this command is advertised for page-level reads.

Unresolved questions
1. Does the live host actually refuse an `id`/`type` read on an invalidated reference? That decides whether #1 is a hole or two dead comments.
2. `refsMs` on a genuinely token-dense DS subtree — still unmeasured after two rounds, and it is the only leg outside the deadline.
3. Should a `--budget` large enough to make a PAGE target expensive be refused outright, given the phase's "no file-wide load" rule?

jangtrinh:
## Stage-4 re-review — round 2 (fix commit `45e8516`)

All three round-2 blockers are fixed, and I reproduced each RED myself by running the **new** tests against the **pre-fix** sources (`git archive 6f673cd` into a scratchpad, post-fix test files dropped in — nothing in the tree touched, no live broker reached).

| round-2 finding | state | RED reproduced here |
|---|---|---|
| **1 · refused identity read left no counted trace** | **fixed** — `readIdentity()` (`plugin/src/main/context-node-record.ts:176`) makes `id`/`type` *read* (a throw **or** a non-string is a refusal); such a node ships as the minimal located record and `walkContext` counts it in `partial`, so `complete` goes false | `context-node-record.test.ts` "a child whose id getter throws ships as a minimal LOCATED record, not as id \"\"" / "…type getter throws is minimal too — never type \"UNKNOWN\" with text skipped" / "…located as the target rather than as a child" / "…only the NAME read refuses", and `context-walk.test.ts` "a child whose identity read refuses is counted partial, so complete goes false" → pre-fix: `expected { id: '', name: 'ghost', …(4) } to deeply equal { …(2) }` and `expected +0 to be 1` |
| **2 · unconditional soft-deadline promise** | **fixed** — the catalog row (`cli/src/command-catalog.ts:65-77`), `skill-template.ts:46` and the emitted `skills/figma-agent/SKILL.md:42,78` all now say the deadline covers the **walk** only, ref resolution runs after it unbounded (`refsMs`/`refsBytes`), and a hit wire timeout is `E_TIMEOUT` + `jobId` recoverable with `job <id> --wait`. `visited` is now defined ("nodes ENQUEUED, not subtree size") and the `partial` sentence names all five triggers | `skill-emitter-drift` green — catalog, template and emitted SKILL.md carry the identical clause; only one copy of the sentence exists in the repo |
| **3 · uncapped `--budget` / `--timeout`** | **fixed at both boundaries, same numbers** — CLI *refuses* `--budget > 512 KB` (`MAX_CONTEXT_BUDGET_KB = CHUNK_LIMIT/1024`, after the finite/positive check so `NaN` still refuses first) and *clamps* `--timeout` to `EXEC_JS_MAX_TIMEOUT_MS = 120_000`, reporting the used value as `budget.timeoutMs`; the plugin re-enforces both in `bounded()` (`executor-context.ts:68`, `CHUNK_LIMIT` / `EXEC_JS_MAX_TIMEOUT_MS`). CLI can only ever send `deadlineMs ≤ 118_000`, so no CLI call can trip the wire bound | `context-command.test.ts` "refuses a --budget past the 512 KB chunk seam…" / "accepts a --budget exactly at the seam" / "CLAMPS --timeout to the CLI-wide 120s max and reports the clamped value" / "reports the timeout it used even when the caller named none" (all four red pre-fix), and `context-executor.test.ts` "refuses a wire budget or deadline past the caps…" → pre-fix `promise resolved "{ schema: 'context/1', …(4) }" instead of rejecting` |

The `runner` param on `run()` is load-bearing, not cosmetic: driving the pre-fix `run()` (which ignores it) sent every non-refusing case straight at the transport — exactly the CI hazard flagged in round 1.

**No regression to the conservation law.** `visited = emitted + omitted.budget + omitted.deadline` still holds through the new path (the ghost is enqueued and emitted), `assertConservation` unchanged, `partial` gained one honest trigger. The located id is prose no Figma id can be confused with — `(unreadable child 0 of 1:1)`, or `(unreadable target)` for the requested root; a real id is preserved when only a non-`id` read refused. Nothing fabricated.

**Gates re-run here** (read-only mandate: no `npm test`, no `npm run build`): `npm run typecheck` → 0 · `npx vitest run tests/context-{command,executor,node-record,refs,walk}.test.ts` → **5 files / 80 tests** (was 70; +10 this round) · `tests/{plugin-build-freshness,skill-emitter-drift,mutating-commands,orb-command-state,broker-client-read-only,plugin-admission}.test.ts` → 6 files / 50 passed, so the committed `plugin/code.js` + `ui.html` are a fresh build and the wire/orb/admission drift gates agree · `npm run lint:comments` → clean, 0 new · no `any`, `as any`, `@ts-expect-error`, `eslint-disable`, `console.` or TODO in the touched sources · no DOM/Node global in `plugin/src/main` (only `Date.now`/`setTimeout`, and DOM-less `tsconfig.main.json` typechecks clean).

### New findings — all informational, none blocking

1. **PR body is stale after this round** (honest-reporting law, and the merge audit reads it): Evidence still says `2241 passed (+70 … +13 from the fix round)` — it is `+10` more now (context files alone went 70 → 80); the What list mentions neither cap nor the counted identity refusal; and the round-1 note still stands that `773 ms` / `13/13` come from the round-0 live pass, not post-fix. One paragraph.
2. `plugin/src/main/context-walk.ts:205` — the build-throw fallback still emits `id: idOf(pending.node)`, i.e. `''` when the id cannot be read, instead of the new locator, though `pending.parentId` and `pending.childIndex` are both in hand. Counted `partial`, so nothing is silent; it is the one place the round-2 finding's shape survives, and unifying it is ~2 lines. (Effectively unreachable now that `buildContextRecord` returns rather than throws on a refused identity.)
3. `plugin/src/main/context-node-record.ts:205-217` — a refusal from the **`name`** read alone collapses a node whose `id` and `type` read fine to `{id, readError}` and returns `children: []`, so a readable subtree is dropped with a `partial` counter but **no frontier cursor**. Honest (`complete: false`) and practically unreachable (an invalidated reference refuses everything), but the id is real in that case, so a `frontier` entry would be actionable if it ever fires.
4. `plugin/src/main/context-node-record.ts` is now 306 lines, past this repo's 200-line modularization guideline. The three refusal-aware readers (`childrenOf`, `countCollapsed`, `readIdentity` + `locate`) are a natural split when it is next touched.

### Verdict: **Approve**

Blockers cleared; #1 above is worth a one-paragraph PR-body refresh before the merge audit, and #2–#4 are follow-up material, not merge conditions.

Unresolved (unchanged from round 1, both accepted as out of scope): whether the live host really refuses an `id` read on an invalidated reference — the guard is correct either way now; and `refsMs` on a genuinely token-dense DS subtree, still the one leg outside the deadline and still unmeasured.

PR #115, opened 2026-09-03
feat(cli): add context dedup and design-intent channel

jangtrinh:
## Why

`figma-agent context` (#114) returns one record per node, and on a text-heavy frame most of those records repeat the same CSS and layout blocks: 40 identical list rows cost 40 copies. The Framelink study showed dedup is the strongest token lever there is, and also where a wrong fact is easiest to ship (two named tokens with equal values merged into one). And the Plugin API exposes, free on every plan, what the designer *meant*: `devStatus`, annotations, dev-resource links, a component's description. Nothing read them.

## What

- `--dedup` (opt-in): identity-first — a node that binds a named variable or style keeps its id ref; only unbound literal `css`/`layout` blocks are interned into `refs.literals`, and repeated subtrees (same everything except id/name/position) become `refs.templates` with occurrences carrying `templateRef` + `rootMap`, so no identity is lost. The reply says what happened: `dedup.{applied, savedBytes (only when > 0), foldedNodes, reason (when not applied)}`, and `budget.emitted === nodes.length + dedup.foldedNodes` is enforced in the plugin. A pure `inflateContextReply` in the CLI restores the raw form; `inflate(dedup(x)) deepEquals x` runs in `npm test` over a fixture corpus (same-value-different-name styles, same-component-different-props instances, visible/hidden twins, bound-vs-literal twins, nested templates).
- `intent` (always on, present only when non-empty): `devStatus` verbatim, annotations, dev resources from one subtree-wide read mapped by node id (opt-in with `--dev-resources`: a fixed ≈2 s server round trip measured on the owner's file, so it never taxes the default path; reported unconditionally under the flag as `budget.devResources {found, attached, readMs}`), and the main component's `description` / `descriptionMarkdown` / `documentationLinks` resolved once per key into `refs.components`. Every read is in a try/catch: a refusal keeps the node, adds `intentError`, counts it `partial`.
- The canonical-JSON FNV-1a hash moves out of `supervised-memory.ts` into one shared helper (pinned by a known-hash test) instead of a second copy.
- Schema stays `context/1`; a call without `--dedup` returns exactly the #114 shape.

## Evidence

Gates after the last edit (233e228): typecheck 0 · build 0 (committed plugin artifacts fresh) · lint:comments 0 (0 new) · emit:skill drift green · `npm test` 168 files / 2341 passed.

Live on the owner's file (plugin hot-reloaded from this build): the 121-node frame drops from 60997 to 39212 bytes with `--dedup` (`savedBytes 21785`, 14 literals), `inflate(dedup(x))` deep-equals the raw reply on three live subtrees (frame, INSTANCE, PAGE); a subtree with nothing repeated answers `applied:false` with its reason; the INSTANCE's main component resolves once with `description` and a differing `descriptionMarkdown`; `devStatus` null and 0 dev resources on this file read as honest absence. With real links (a temporary frame given two dev resources plus one on a child, written through `exec-js` on a Free seat and deleted afterwards): `--dev-resources` reports `{found:3, attached:3}` with `intent.devResources` on exactly the two linked nodes, `--depth 0` reports `{found:2, attached:2}` (root only), and the no-flag call carries neither the key nor any `intent`.

Review: stage-4 adversarial review (Request changes: one Critical — `--dedup` folded records out of `nodes[]` with no counter while the emitted skill still said `emitted is always nodes.length` — plus three Important), fixed test-first; a second controller round made dev resources opt-in after the live measurement above (`getDevResourcesAsync` is a fixed ≈2 s server round trip regardless of subtree size, 11 to 453 nodes). Rounds three and four fixed the `devResources` counter twice more (`attached` counted records, not links; then counted records the budget later discarded) — each with a failing-first test and the live run above. Mutation testing: 20 wrong-fact mutants, 18 killed by the existing suite, 2 survivors now covered (a subtree signature that ignored `bindings`; a "reduction" of zero bytes announced as applied).

## Not in this PR

Flipping `--dedup` on by default (owner decision; live templating never fired on the probed subtrees); annotation category names; META-1 metadata change capture; the README/deep-dive concession correction lands next.


jangtrinh:
## Stage-4 review — PR #115 `feat(cli): add context dedup and design-intent channel`

Reviewed `81de6b9` against `4256e3b` (21 files, +3002/−296). Read-only session: `npm test` and
`npm run build` are off-limits here (they kill the owner's live broker / hot-reload the live
plugin), so the suite total is reported as unverified below and the build claim is checked
through `tests/plugin-build-freshness.test.ts` instead.

**Gates re-run after the last edit:** `npm run typecheck` pass · `npm run lint:comments` clean
(0 new, 53 baselined files) · `context-dedup` + `context-intent` + `context-command` +
`supervised-memory` + `plugin-build-freshness` + `skill-emitter-drift` = 113 pass ·
`mutating-commands` + `orb-command-state` + `broker-client-read-only` + `plugin-admission` +
`context-walk` + `context-refs` + `scan-node-walker-bundle` = 75 pass. Emitted `SKILL.md`
matches the catalog (drift gate green).

### Summary

The round-1 Critical is genuinely fixed at the root, not papered over: `dedup.foldedNodes` is
derived from the folding pass, the identity `emitted === nodes.length + foldedNodes` is asserted
in **both** the transform and the executor, and both agent-facing documents now carry the
exception *and* the frontier→`rootMap` caveat. The hash duplication is gone with two literals
pinned so the move is provable, and the shared-object bug is closed on both expansion paths.
Scale and correctness hold under an independent probe: 601 records dedup in 13 ms, fold 450
records, and `inflate(dedup(x))` came back deep-equal.

Two Important findings remain, both narrow and both in the same place the round-1 I3 finding
was: the `devResources` accounting. One of them is a wrong fact now emitted into `SKILL.md`,
which is the failure mode this repo rates worst.

**Risk level: medium** — no data loss, no auth surface, no schema break; one incorrect
instruction to agents and one flag-gated timeout regression.

---

### Findings

#### Important 1 — `budget.devResources.attached` counts RECORDS, `found` counts LINKS, and the catalog draws a false conclusion from the gap

`plugin/src/main/context-intent.ts:127` — `attached += 1` per record carrying at least one
resource. `plugin/src/main/context-intent-readers.ts:100` — `found: list.length`, i.e. per
resource. Two different units compared in one sentence.

`cli/src/command-catalog.ts:102-105` (and so `skills/figma-agent/SKILL.md`):

> attached counts the EMITTED records a resource landed on, so attached below found means the
> rest belong to nodes absent from this reply: descendants the budget or deadline never
> enqueued, nodes outside --depth, or a record whose identity read refused…

**Failing input → wrong output** (probed against the real modules, not a mock of them):
a frame `1:1` with child `1:2`, both emitted, `getDevResourcesAsync` returning two links on
`1:1` and one on `1:2` →

```
budget.devResources { found: 3, attached: 2 }
```

Every node is in the reply and every link is attached, yet `attached < found`. An agent
following the sentence re-runs `context` with a larger `--budget`/`--depth` chasing a node that
does not exist. Two links on one layer is the ordinary case (a Figma layer takes many dev
resources), so this fires on the first real file that uses the feature — and the existing test
at `tests/context-intent.test.ts:98-113` bakes the same reasoning into a comment while using a
fixture that has exactly one link per node, so it cannot catch it.

Same finding as round-1 I3: the sentence was reworded, but the reworded sentence is still
false because the unit mismatch underneath it was never addressed.

**Smallest fix:** `attached += resources.length` at `context-intent.ts:127` — same unit as
`found`, and the catalog sentence becomes true as written. (If the record count is the number
worth reporting, rename it `attachedNodes` and reword all three documents instead; do not keep
`attached` next to `found` in different units.)

Second, smaller leak in the same gap: `context-intent-readers.ts:89` drops a resource whose
`nodeId` is empty/unreadable from `byNode` with no counter, and `found` still counts it. That is
a silent vanish under a house rule that forbids exactly that — either count it
(`skipped`/`unaddressed`) or say in the doc that it lands in the gap.

#### Important 2 — `--dev-resources` spends the entire wire-timeout headroom before the walk's soft deadline starts, so the "partial instead of a timeout" guarantee stops holding

`cli/src/commands/context.ts:95-97` sets `deadlineMs = timeoutMs − DEADLINE_HEADROOM_MS`
(45 000 − 2 000 = 43 000 by default, `shared/protocol.ts:711`). That 2 s exists so the plugin can
finish ref resolution and get the reply on the wire before the CLI gives up.

`plugin/src/main/executor-context.ts:84-93`: the ~2.1 s dev-resource read is awaited **before**
`deadlineAt: env.now() + deadlineMs` is computed, so the walk still receives its full 43 s.

**Failing input → wrong output:** `figma-agent context <large page> --dev-resources` on a subtree
big enough to run the soft deadline out. Plugin wall clock ≈ 2.1 s + 43 s = 45.1 s against a
45 s wire timeout, with ref resolution (`refsMs`, explicitly unbounded) still to come. The caller
gets `E_TIMEOUT` + `jobId` instead of the partial-with-counts reply the soft deadline is there to
produce. The catalog documents that the read is unbounded by `--budget`/`--depth`, but not that
it consumes the timeout headroom — and the read is a *fixed* cost, so this is deterministic, not
a tail case.

**Smallest fix** (one line, keeps the contract the flag-less path has):

```ts
deadlineAt: env.now() + Math.max(1, deadlineMs - devResourcesMs),
```

`devResourcesMs` is already computed at line 88.

#### Suggestion 1 — the walk's empty component name can overwrite a real one

`plugin/src/main/executor-context.ts:106-108`: `refs.components[key] = { ...component, ...refs.components[key] }`.
`plugin/src/main/context-refs.ts:126` writes `{ name }` where `name` falls back to `''` when
`main.name` was unreadable. The merge then replaces the intent reader's real name with `''`.
Merge the name only when it is non-empty.

#### Suggestion 2 — a repeated subtree nested inside a non-folding candidate is never offered

`shared/context-dedup-templates.ts:63-68` stops descending as soon as a subtree's *global*
signature count is ≥2, but `:104` may then drop that signature because only one **outermost**
occurrence survived. Both the outer subtree and any repeats inside it stay raw. Compression loss
only — no wrong fact — but worth a comment or a re-offer of the descendants.

#### Suggestion 3 — `intent` has no opt-out, and it can move P1's `partial` / `complete`

The intent reads are wired into every walk (`executor-context.ts:90`, `context-intent.ts:103-142`).
A refused `devStatus`/`annotations` getter sets `intentError`, counts the record `partial` and
drops `complete` to `false` — for a caller who never asked for intent and is reading those two
fields under the #114 contract. Behaviour is honest and the reads are guarded; the point is that
the meaning of a P1 field now depends on a channel the caller cannot decline. Either say so in
the catalog next to `partial`, or add `--no-intent`.

#### Suggestion 4 — `tests/supervised-memory.test.ts:19` says "These three strings" and pins two.

---

### Verification of the round-1 findings (each against the committed diff)

- **C1 (Critical) — fixed at the root.** `shared/context-dedup.ts:26-30` declares `foldedNodes`
  (present at `0` too), `:54-60` `assertDedupConservation`, `:91` asserts inside the transform,
  `plugin/src/main/executor-context.ts:128-131` asserts again on the object that goes on the wire
  against `walk.accounting.emitted`. Named tests: `context-intent.test.ts` "keeps the conservation
  law checkable after folding, with a RECORD counter" (`foldedNodes === 4`,
  `emitted === nodes.length + foldedNodes`, and P1's `visited = emitted + omitted.*` still holds)
  and "reports foldedNodes 0 rather than omitting it when only literals folded";
  `context-dedup.test.ts` "counts every folded record" over the whole corpus. Docs: catalog
  `:77-78` carries the `emitted` exception, `:85-86` the frontier→`rootMap` sentence; both are
  mirrored in `skill-template.ts` and the emitted `SKILL.md` (drift gate green). A true RED could
  not be produced — the pre-fix state was never committed — so this is verified structurally plus
  by mutants #1/#2 in `mutation-phase-02-context.md`.
- **I1 — fixed.** `shared/canonical-content-hash.ts` is now the single canonicaliser + FNV-1a;
  `supervised-memory.ts` delegates and its output is pinned by literals `fnv1a-34a7cae6` /
  `fnv1a-b17f217f` plus a key-order test (`tests/supervised-memory.test.ts:19-35`), which is what
  makes the move provable — a changed hash would read every stored correction event as corrupt.
  The false `Math.imul` sentence is replaced by a build-grounded one, and it checks out:
  `git show 4256e3b:plugin/code.js | grep -c Math.imul` → 1, so `imul` already shipped in the
  sandbox bundle before this PR.
- **I2 — fixed on both paths.** `context-dedup-literals.ts:156` (literal expansion) and
  `cli/src/commands/context-inflate.ts:90` (template expansion) both deep-clone. The test "gives
  every inflated record its own nested objects" asserts non-identity on `css` (literal path) *and*
  `bindings` (template path — the field the literal table cannot accidentally clone). Mutant #20
  confirms the template clone is load-bearing.
- **I3 — reworded, not fixed.** See Important 1.
- **S1** — documented (runs before the walk, bounded by neither flag, `includeChildren: false` at
  `--depth 0`) and narrowed in code — but see Important 2 for the part the doc still misses.
  **S2** — fixed: `context-intent.ts:61,66-76` memoises the null answer per key. **S3** — fixed:
  catalog `refsBytes` now says it covers the whole `refs` object under `--dedup`. **S4** —
  documented in the `context-inflate.ts` header ("inverse by DEEP EQUALITY, not byte-for-byte").
  The paint-style caveat is now stated at `context-dedup-literals.ts:13-16`.

### PR-body claims vs the tree

| Claim | Result |
|---|---|
| Schema stays `context/1`, no-flag reply is #114's shape | **Verified** — `CONTEXT_SCHEMA` untouched, `shared/protocol.ts` not in the diff; without flags no `dedup` block, no `budget.devResources`, and `refsBytes` measures the same object (`payload.refs === refs`). Additive only: `intent` (when non-empty) and description fields on `refs.components[*]` — with the caveat in Suggestion 3. |
| 20 mutants / 18 killed / 2 survivors covered | **Consistent** with `mutation-phase-02-context.md` (21 numbered rows, #8/#9 one mutant); both survivor tests are in the committed `tests/context-dedup.test.ts`. |
| 168 test files / 2333 passed | File count **verified** (168). The pass total is **not verified** — the suite may not be run in this session. |
| 60997 → 39212 bytes, `savedBytes 21785` | Arithmetic **consistent**; the live measurement itself is not reproducible read-only. |
| ≈2 s fixed `getDevResourcesAsync` cost, three live subtrees round-tripped | **Not verifiable read-only**; the numbers are used consistently across catalog, code comment and report. |
| build 0 (committed plugin artifacts fresh) | **Verified** via `tests/plugin-build-freshness.test.ts` (re-runs the esbuild steps with `write:false` and diffs the committed `plugin/code.js` / `ui.html`). |

### Regression checks asked for

- **Sandbox globals** — none of `Buffer`, `TextEncoder`, `structuredClone`, `process`, `require`,
  DOM or `Intl` appear in `shared/canonical-content-hash.ts`, `shared/context-dedup*.ts` or
  `plugin/src/main/context-*.ts`. `padStart`, `flatMap`, `Math.imul` and `localeCompare` were all
  already present in the committed `plugin/code.js` at `4256e3b`, so nothing new is asked of the
  sandbox.
- **Mixed `.ts` / extensionless imports** — safe, and the reason is that nothing resolves these
  specifiers at runtime: `cli/dist/figma-agent.js` and `plugin/code.js` are esbuild bundles and
  vitest transforms in-process; esbuild resolves `./x` and `./x.ts` identically for a relative
  path. Both `cli/tsconfig.json` and `plugin/tsconfig.json` set
  `allowImportingTsExtensions: true` with `moduleResolution: "bundler"`, and all three
  `tsc` projects are `noEmit`, so no declaration output can trip on the extension. Confirmed by
  typecheck + the build-freshness gate reproducing the committed bundle.
- **Template search complexity** — linear in payload size, not quadratic: one canonicalisation per
  record, one interned string per subtree (`canonical | childSigs`), iterative post-order.
  Measured on 601 records (150 rows × 3 cells): **13 ms** to dedup, 1.5 ms to inflate,
  450 records folded, `inflate(dedup(x))` deep-equal, conservation identity held.
- **`canonicalContent` key order** — the sort is `localeCompare`, **not** codepoint, and stored
  correction hashes do depend on it for non-ASCII keys. That is pre-existing and was moved
  verbatim (the header says so), and it is the right call here: swapping in `<`/`>` would
  invalidate every event on disk. Worth knowing for later: `localeCompare` is ICU/locale-dependent
  and correction hashes are written in the Figma sandbox (`correction-edge-store.ts`) but
  validated in Node (`sync-corrections.ts`), so a non-ASCII key could in principle hash
  differently across the two engines. Dedup keys cannot diverge — they are computed only in the
  plugin and consumed as opaque strings. If this is ever changed, it needs a hash version, never a
  silent swap.

### Verdict

**Request changes** — Important 1 (a wrong fact already emitted into `SKILL.md`; agents will act
on it) and Important 2 (a documented guarantee that stops holding under `--dev-resources`). Both
are one-line code fixes plus a doc rewording, each needs a test that fails against the current
code: multi-link-per-node for the first, and a fake clock proving the walk deadline shrinks by the
measured read time for the second. Everything else in the round-1 list is fixed at the root and
provable from the committed tests.

jangtrinh:
## Stage-4 re-review — round 3 (`6700c4b` on `81de6b9`)

**Gates re-run after the last edit:** `npm run typecheck` pass · `npm run lint:comments` clean (0 new) · `context-intent` + `context-executor` + `supervised-memory` + `skill-emitter-drift` + `plugin-build-freshness` = 67 pass (committed `plugin/code.js` / `ui.html` reproduce from source).

**RED proof:** the pre-fix tree (`81de6b9`) was extracted and the round-3 test files dropped on top — 6 of the new/changed tests fail against it, 0 against `6700c4b`.

### Requested findings

- **Important 1 — fixed.** `context-intent.ts:128` `attached += resources.length`; the reviewer's probe is now a test: *"counts links, not records: two links on a frame plus one on its child is 3 of 3"* (`{found: 3, attached: 3}`), plus *"attaches the dev resources of the one subtree read, keyed by node id"* (2 links on one node → `attached 2`). Both RED pre-fix.
- **Important 1b (`unaddressed`) — fixed.** `context-intent-readers.ts:94-99,110` (and the three error returns at `:83,89,92`) counts a link with no readable `nodeId` (and a non-object list entry) instead of dropping it; `executor-context.ts:178` emits it only above 0. Tests *"counts a resource with no readable nodeId instead of dropping it"* and *"omits unaddressed when every resource named a node"*. First is RED pre-fix. Conservation holds: `found = attached + unaddressed + links on nodes absent from the reply`.
- **Important 2 — fixed.** `executor-context.ts:101` `deadlineAt: env.now() + Math.max(1, deadlineMs - devResourcesMs)`; since `now()` is read after the read, this is exactly `devResourcesStartedAt + deadlineMs`, so the flag can no longer push the wall clock past the wire timeout. Tests *"charges the dev-resource read against the walk deadline, not on top of it"* (fake clock, 2 s read against a 3 s deadline → `omitted.deadline 2`) and *"leaves the walk a positive deadline even when the read ate all of it"* (root still emitted). Both RED pre-fix. No-flag path byte-identical (`devResourcesMs` stays 0).
- **S1 — fixed** (`executor-context.ts:119-121`, name merge prefers the non-empty one; test *"never lets the walk's empty component name overwrite the one intent read"*, RED pre-fix). **S2 — fixed** (`context-dedup-templates.ts:55-63` states the loss and why it is not chased). **S3 — fixed** (catalog `:80-84` names `intentError` and says the intent reads cannot be declined). **S4 — fixed** (`supervised-memory.test.ts:21` "These two strings").

Catalog/SKILL wording checked line by line for `attached` / `unaddressed` / the charged deadline; drift gate green. No new sandbox-hostile global, `partial` / conservation-assert / no-flag paths untouched.

### New finding

**Important 3 — `attached` counts links on records the budget then threw away, so `attached == found` while every link's node is on the frontier.** `plugin/src/main/context-intent.ts:128` increments at record-BUILD time, but `context-walk.ts:231-235` builds a whole breadth-first batch and only then drops its tail (and the rest of the queue) into `omitted.budget`. Probe against the real modules, frame `p:1` with three `TEXT` children, links on `p:2` and `p:4`, `--budget` small enough to cut:

```
nodes: ['p:1']
budget.devResources {found: 2, attached: 2}
omitted {budget: 3}, frontier: p:2, p:3, p:4
```

Both links are counted as landed, no record in the reply carries a `devResources` key, and both their nodes are on the frontier. That falsifies the counter's own definition (`context-intent.ts:45` "How many dev-resource LINKS landed on an emitted record") and hides the gap the `found`/`attached` pair exists to expose — an agent reads "all links accounted for" and never follows the frontier. Budget exhaustion on a page is the ordinary case for this flag, not a tail. Same shape as Important 1: the number is not the unit/event it claims. Not a regression from `6700c4b` — the identical probe returns `{found: 2, attached: 2}` at `81de6b9` — but it is this PR's own new counter.

Smallest fix: count at emit time instead of build time — sum `record.intent?.devResources?.length` over `walk.nodes` in `executor-context.ts` (pre-dedup, so folded template records still count) and drop `attachedDevResources()`. Needs a test that fails against the current code: links on a budget-dropped child.

### Suggestions

- `plugin/src/main/executor-context.ts:168-171` still carries the round-2 wrong sentence verbatim ("`attached` counts the EMITTED records a resource landed on…") right above the block that now emits links, and it does not mention `unaddressed`. The fixed wording exists three files over.
- `cli/src/skill-template.ts:60` still prints the shape as `budget.devResources {found, attached, readMs}` — no `unaddressed?`, and an agent can now see that key.
- `tests/context-intent.test.ts:21` — the shared `noDevResources` fixture still sets the removed `attached: 0` and omits `unaddressed`; `tests/` is outside all three `tsc` projects, so nothing catches it.

### Verdict

**Request changes** — one Important remains (a devResources count that contradicts its own documented meaning, in the same accounting this PR is being blocked on), plus three one-line doc/fixture leftovers. Everything actually asked for in round 2 landed at the root, with a failing-first test each.

— stage-4 re-review round 3, posted by the controller session

PR #117, opened 2026-09-03
feat(cli): add session coverage statement to status

jangtrinh:
## Why

An agent that reconnects to a file gets a feed, not a coverage claim. Everything it needs to know what this session cannot account for already existed as scattered counters (gap-fill, capture, perf, relay drops in the broker log) and nothing assembled them. The competitor study found no bridge that does this; the house rule that a wrong fact is worse than an absent one makes the assembly the risky part.

## What

- `coverage: {complete: boolean|null, gaps: [{kind, count, see}]}` on the plugin's STATUS reply and on `figma-agent status`'s `plugin` object: one row per KIND with a count, never per event; a closed enum of kinds; `see` names the STATUS field path or command that carries the detail.
- Plugin-known rows (baseline missing, unreadable or first session on this file; pages reported top-level only; page read errors; evicted baselines; gap-fill errors; page fallbacks; capture errors; property read errors) are built in the plugin from what main already knows. Broker-known rows (relay frames dropped before the socket opened, edit batches replayed after an outage, other files connected) are appended by the CLI from BROKER_HELLO, backed by a daemon-scoped per-instance tally that a reconnect replaces rather than sums.
- `complete` is `true` only when every count is zero and boot has finished; `null` while nothing is connected or boot is running; never true by default.
- The emitted skill tells the agent to read `coverage` first on connect and to check `connected` before reading `complete:null`.

## Evidence

Gates after the last edit: typecheck 0 · build 0 (committed plugin artifacts fresh) · lint:comments 0 (0 new) · emit:skill drift green · `npm test` 172 files / 2401 passed (+60 for this phase, after two review rounds).

Live on the owner's 21-page file: `coverage {complete:false, gaps:[{kind:"pages-top-level-only", count:15, see:"status.gapfill"}]}`, matching `gapfill.pagesTopLevelOnly`; during a reconnect `{complete:null, gaps:[]}` beside `connected:false`.

Review: stage-4 adversarial review (Request changes: two Critical — a first-ever session on a file claimed `complete:true` while its own feed carried `baseline-missing`; a replayed batch was counted twice because one batch posts two frames — plus four Important), every finding fixed test-first with its RED quoted; the reviewer's question of which frame is the superset was settled by two characterization tests (every batch posts EDIT_FEED, only component batches also post DOC_CHANGE). Mutation testing: 18 wrong-fact mutants, 17 killed by the existing suite, 1 survivor (a silent tally eviction) now covered by a real-broker test. Round two rooted every `see` pointer at the reply (`status.plugin.gapfill`, type-enforced so a typo fails typecheck), keyed the broker rows by file identity rather than file name (two files named `Untitled` no longer share a relay tally), and made the tally cache least-recently-touched; live, the one row on the owner's file resolves on the actual reply.

## Not in this PR

Per-event detail (the feed and `errors` keep that); retrofitting coverage into other commands; `deletedRechecked` (a node that looked deleted but was not is not a gap).


jangtrinh:
## Stage-4 review — session coverage statement

### Summary

The feature lands as designed: one bounded `{complete, gaps[]}` statement, a closed kind
vocabulary, `complete: null` as the honest pre-boot answer, per-instance broker tallies that
survive the reconnect that deletes the registry entry, and present-only-when-meaningful
preserved on every pre-existing field. All four prior-round findings are fixed, and each fix
has a test that genuinely fails without it (RED reproduced below, not taken on trust).

Two defects remain, both in the same place: the row's `see` pointer and the identity the
broker rows are attributed to. Both produce a fact an agent reads and acts on, and both are
invisible to the current tests. The repo's own rule — a wrong fact is worse than an absent
one — is what makes them blocking rather than polish; the fixes are one to four lines each.

**Risk level: medium.** Diagnostic surface only, no mutation path, no data loss. But the
consumer is an agent that is now instructed to read this block *first* on connect, so a
wrong row is acted on rather than skimmed.

---

### Findings

#### Important 1 — `see` points at a path that does not exist on the reply

`shared/session-coverage.ts:23-25` (`status.gapfill`, `status.captureErrors`, `status.perf`)

Those three targets are rooted at the reply; on a real `figma-agent status` reply they live
under `plugin`. Probed against a live scratch broker (the reply from `cli/src/commands/status.ts`
`run()`):

```
top-level keys: ["broker","plugins","activePlugin","plugin","protocolVersion","mutationGates","mutationGateStoreHealth"]
reply.gapfill        = undefined
reply.plugin.gapfill = {"pagesDiffed":1,"pagesTopLevelOnly":15, ...}
```

So the PR body's own live example — `{kind:"pages-top-level-only", count:15, see:"status.gapfill"}` —
sends the reader to `status.gapfill`, which resolves to nothing, while `see:"status.plugins[]…"`
on the broker rows *is* reply-rooted. Same prefix, two different roots, and the row whose
whole job is "where the detail is" is the one that misses. This fires on every session with a
plugin-side gap, not on an edge case.

The drift guard cannot catch it by construction: `tests/session-coverage-status-command.test.ts:258-273`
resolves each target against `result` **and then** against `result.plugin`, accepting either.
That fallback is exactly the ambiguity being asserted away.

Smallest fix: spell the plugin-scoped targets `status.plugin.gapfill` / `status.plugin.captureErrors`
/ `status.plugin.perf` (one edit each in `shared/session-coverage.ts` and
`plugin/src/main/session-coverage.ts:42,53,55,58,62,66,73`), and drop the second root from
`resolveTarget` so the test pins one meaning. Related, same fix window: `skills/figma-agent/SKILL.md`
step 5 says "read `coverage` first" without saying it lives at `plugin.coverage`, and step 1
sends the agent to `status --peek`, which carries no `coverage` at all.

#### Important 2 — broker rows are attributed by file NAME, so two files sharing a name cross-contaminate

`cli/src/commands/status.ts:177` and `:181-183`

`fileRows` selects by `p.fileName === activePlugin` and `otherFiles` de-duplicates on
`p.fileName ?? p.fileKey ?? '?'`. This package has one canonical identity chain and it is the
other way round — `cli/src/transport/file-identity.ts` rules "raw fileKey is canonical
everywhere… a slug can collide two keys differing only by case", and `fileIdentity` is already
imported in this file (line 15, re-exported from `project-bind.ts`). The comment on line 180
claims the new code "falls back the same way the rest of this CLI does (name, then key)" —
the rest of the CLI does key, then name.

Two probes against a live scratch broker, both on the committed code:

1. Three sessions — `Untitled/KEY_A`, `Untitled/KEY_B`, `Design File/KEY_ACTIVE` (two genuinely
   different other files, one name):
   `gaps: [{kind:"other-files-connected", count:1, …}]` → **should be 2**. One connected file
   whose edits are not in view is silently missing from the statement.
2. `Untitled/KEY_OTHER` reports `PLUGIN_RELAY_STATS sessionTotal.frames: 7`, then
   `Untitled/KEY_ACTIVE` connects clean and is the active target:
   `gaps: [{kind:"relay-dropped-frames", count:7, see:"status.plugins[].relayDroppedFrames"}]`
   → **the active session is told it lost 7 frames it never lost**, its `complete` flips to
   `false`, and the other file disappears from `other-files-connected` at the same time. A
   fabricated count on one row and a vanished one on another, from a single collision.

`Untitled` is Figma's default file name, so the trigger is a plausible desktop state, not a
contrived one. Note the tally layer itself is correct — it is keyed by `instanceId`; only this
CLI-side roll-up loses the key.

Nothing in the suite pins the identity rule. Mutant check: swapping line 182 to
`p.fileKey ?? p.fileName ?? '?'` leaves `tests/session-coverage-status-command.test.ts` and
`tests/session-coverage-broker-tally.test.ts` fully green (15/15) — every fixture connects
without a `fileKey`, so the fallback order is unobserved.

Smallest fix:

```ts
const activeRow = all.find((p) => p.fileName === activePlugin) ?? null;   // or plugins[0] under --file
const activeId = activeRow ? fileIdentity(activeRow.fileKey, activeRow.fileName) : null;
const fileRows = activeId === null ? [] : plugins.filter((p) => fileIdentity(p.fileKey, p.fileName) === activeId);
const otherFiles = new Set(
  all.filter((p) => fileIdentity(p.fileKey, p.fileName) !== activeId)
     .map((p) => fileIdentity(p.fileKey, p.fileName)),
).size;
```

plus a test with two rows sharing a name and differing by key. (Two keyless same-named files
still collapse — that is the honest limit of a nameless identity and worth one sentence in the
comment, replacing the current inaccurate one.)

#### Important 3 — the closed `see` list is closed only in prose

`shared/session-coverage.ts:43` — `coverageRow(kind: CoverageGapKind, count: number, see: string)`.
`kind` is enforced against the enum at runtime and by the type; `see` is a bare `string`. The
drift test then iterates `COVERAGE_SEE_TARGETS` — the *declared* list — so it proves the list
is resolvable, never that an emitted row uses a member of it. A typo in
`plugin/src/main/session-coverage.ts` (`'status.gapFill'`), or `broker-coverage-rows.ts:38-49`
drifting its template, ships green. This is also why Important 1 could land at all.

Smallest fix: `export type CoverageSeeTarget = typeof COVERAGE_SEE_TARGETS[number];` and take
`see: CoverageSeeTarget`. The templated broker rows still typecheck — `pluginsField` is a
two-literal union, so the template resolves to declared members.

#### Suggestion 1 — tally eviction drops by insertion order, not recency

`cli/src/transport/session-tallies.ts:50-64`. `entryFor` never re-inserts on update, so the
victim is the oldest *created* instance, not the least recently touched. One window left open
while another reloads the plugin 128 times evicts the long-lived session's tally; `status` then
reports `complete: true` for a session that demonstrably dropped frames — the exact failure the
module header exists to prevent, downgraded to a log line. The eviction is logged, so nothing
vanishes silently, but the statement itself goes wrong. One line inside the cache-hit branch:
`tallies.delete(instanceId); tallies.set(instanceId, existing);`.

#### Suggestion 2 — a foreign or unparseable baseline reports as two gaps with one cause

`readFileBaseline` (`plugin/src/main/gapfill-baseline-store.ts`) returns `baseline: null` **plus**
an error for a baseline belonging to another file, so `runGapfillDiff` takes the `!prev` path:
`baselineFirstRun = true` → `baseline-missing: 1`, and the recorded error → `gapfill-errors: 1`.
Both rows are individually true and the STATUS type comment on `baselineFirstRun` correctly
covers the foreign case ("no usable baseline"), so this is not the misfire it could have been.
But `gapfill-errors`' own comment in `plugin/src/main/session-coverage.ts:59-61` explains the row
as "that page's edits went unreported" — no page walk failed here. Either widen that comment or
scope the row; and the coverage shape for the foreign case
(`tests/edit-gapfill-baseline.test.ts:…"belongs to another file"`) is currently unpinned — it
asserts the notice and the error string, not the two rows an agent will see.

#### Suggestion 3 — `Math.max(0, otherFiles)` is unreachable

`cli/src/transport/broker-coverage-rows.ts:49`. The argument is a `Set.size`. `coverageRow`
already refuses `<= 0`. Drop the clamp.

#### Suggestion 4 — the branch is behind `main`

`0b6ad05` (#116) is not an ancestor of `56d3fdd`. Mergeable, but merge main before the final
gate rerun so `npm test` and the panel gate run on the tree that will actually land.

---

### Verified

Prior-round findings, each with the test that fails without the fix (RED reproduced on an
extracted copy of `56d3fdd` with the single line reverted, not assumed):

- **C1** first session claiming `complete: true` — reverting `stats.baselineFirstRun = true`
  (`plugin/src/main/edit-gapfill.ts:381`) fails `tests/edit-gapfill-baseline.test.ts` › "marks the
  session a FIRST RUN…" (`expected false to be true`). Reverting the `|| gapfill.baselineFirstRun
  === true` OR in `plugin/src/main/session-coverage.ts:41` fails two more:
  `tests/session-coverage.test.ts` › "a first-ever session on a file is a gap, even though it wrote
  a baseline" and `tests/executor-ops-status.test.ts` › "a FIRST-EVER session on a file cannot be
  complete". Both layers pinned; the flag is threaded through `GapfillStats` → `toGapfillStatus`
  (present-only-when-true) → `GapfillStatus` with a type comment that states what the row means.
- **C2** replayed double count — counted once on EDIT_FEED only
  (`cli/src/transport/broker-daemon.ts:2273`), pinned by `tests/session-coverage-broker-tally.test.ts`
  › "ONE replayed batch counts ONCE, though it arrives as both a DOC_CHANGE and an EDIT_FEED",
  with the superset claim itself characterization-tested (`tests/document-change-capture.test.ts`
  › "a plain node edit posts EDIT_FEED and NO DOC_CHANGE" / "a component edit posts BOTH frames").
  The count sits before any early return in that branch, so no counted batch is later discarded.
  Renamed through `SessionTally` / `PluginStatusEntry.replayedBatches` / the protocol doc.
- **I1** sessions vs files — `other-files-connected` is now a `Set`, pinned by "two windows on ONE
  other file are ONE other file, not two". (Correct for the distinct-name case; see Important 2.)
- **I2** `see` under `--file` — `pluginsField` threaded through `brokerCoverageRows`, pinned by
  "under --file the row points at pluginsAll — plugins[] no longer holds what was counted".
- **I3** skill sentence — identical in `cli/src/skill-template.ts` and `skills/figma-agent/SKILL.md`,
  pinned by the emitter drift test plus "the connect protocol tells an agent to read `coverage`
  first, and what null means".
- **I4** unreachable `see` — a drift test now exists (its limits are Important 1 and 3).

Regressions checked, none found:

- `buildBrokerHelloData` (`cli/src/transport/broker-status.ts:64-88`) keeps today's shape byte for
  byte: `runningJob`/`queueDepth` still ride only when `jobStatusFor` is passed, the two new fields
  only above zero, key order unchanged, and the `jobStatusFor === undefined && tallyFor === undefined`
  short-circuit preserves the raw-list path.
- `status --peek` returns at `cli/src/commands/status.ts:62`, before any coverage code.
- The `plugin` legacy mirror keeps `connected`/`state`/`lastHeartbeatAge` plus the scene spread and
  adds one key; the always-present `coverage` is a deliberate, documented break of the
  present-only-when-meaningful contract on that object only.
- `shared/session-coverage.ts` and `plugin/src/main/session-coverage.ts` are total functions over
  plain data — no clock, no host API, no DOM/Node global in either.

PR-body claims, checked against the tree:

- `typecheck 0` — reproduced (`tsc` × 3 projects, clean).
- `build 0` — `tests/plugin-build-freshness.test.ts` green, i.e. the committed `plugin/code.js` /
  `plugin/ui.html` match a fresh esbuild of the sources; working tree is clean at `56d3fdd`.
- `lint:comments 0 new` — reproduced ("53 file(s) with baselined citations, 0 new"). It also reports
  7 stale baseline entries — pre-existing, not this PR's.
- `emit:skill drift green` — reproduced.
- `+56` — exact: the diff adds 56 `it(` and removes 0.
- `172 files` — exact.
- `2397 passed` — **not verifiable read-only** (`npm test` is off-limits here); 10 PR-touched files
  ran green, 179 tests.
- Mutation report (18/17/1) — not re-run wholesale; the three mutants I did reproduce behaved as
  claimed, and I found one additional survivor outside that set (Important 2's identity order).

### Verdict

**Request changes.** Important 1 is unconditional and hits the row an agent is now told to read
first; Important 2 fabricates a count on a clean session when two files share a name. Both are
small, both need one test each that the current fixtures do not express. Important 3 is the guard
that would have caught the first one. The rest is non-blocking. Nothing here questions the design —
the statement, the closed vocabulary, the `null` verdict, and the instance-keyed tally are right,
and the C1/C2 fixes are properly nailed down.

jangtrinh:
## Stage-4 re-review — round 2 (`bc5c61d` on `4a09430`)

Each fix re-verified by reverting it in-tree and naming the test that goes RED. Read-only otherwise: `npm run typecheck` clean, `npm run lint:comments` clean (0 new), `tests/plugin-build-freshness.test.ts` green (the build claim), 135 targeted tests pass, no C1/C2 regression (114 pass).

| # | Verdict | RED proof |
|---|---|---|
| I1 `see` rooted at the reply + one-root resolver | fixed | vocabulary reverted to `status.gapfill` → `session-coverage-status-command.test.ts` › "each one resolves on a real `status` reply" fails (`see: status.gapfill` resolves to nothing). Rows alone reverted → 5 failures in `session-coverage.test.ts` |
| I2 identity-keyed `fileRows` + `other-files-connected` | fixed | name-keyed roll-up restored → both new tests fail ("two OTHER files that share a name…", "a same-named OTHER file's losses…") |
| I3 typed `see` | fixed | `'status.plugin.gapFill'`-style literal → `tsc` TS2345 ×5 in `plugin/src/main/session-coverage.ts` |
| S1 LRU touch | fixed | touch removed → `session-tallies.test.ts` › "a tally TOUCHED after newer ones survives" fails |
| S2 skill sentence (`plugin.coverage`, `--peek`) | fixed | sentence reverted → 2 failures in `skill-emitter-drift.test.ts`, incl. committed-`SKILL.md` parity |
| S2 foreign-baseline shape pinned · S3 clamp dropped · S4 main merged | fixed | new `edit-gapfill-baseline.test.ts` case; `4a09430` |

**Deviation (`hello.pluginInfo` instead of a name `find`): accepted.** `pluginInfo` is `registry.selectTarget(currentFilter()).scene` (`cli/src/transport/broker-status.ts:75,122`) — the same entry that produced `activePlugin`, captured before the STATUS merge can null it. A name `find` searches `statusList()` (sorted by `lastSeenAt`) while dispatch selects by `lastActiveAt` (`plugin-registry.ts:357-382`), so `pluginInfo` is the better-anchored source. Under `--file`: with one exact match both resolve to the same live entry; with two, nothing is attributed at all (see New 2). Residual, not a regression: a panel target pin routes dispatch (`broker-daemon.ts:1233`) but is not reflected in `pluginInfo`/`activePlugin`, so a pin on file A while B is more recent still splits plugin-side rows from broker-side rows — identical to the pre-fix behavior.

**New 1 (low, comment only) — `tests/session-coverage-status-command.test.ts:262-265`.** "The LOSSY one connects FIRST, so a name-keyed roll-up that happens to pick the first matching row cannot pass by luck of insertion order" claims a guard the test does not give: `plugins` is `lastSeenAt`-DESC, so `plugins.find(p => p.fileName === activePlugin)` as `activeKeySource` passes all 14 tests here (verified). The tests discriminate identity-keying vs name-keying — which is the point — but not the anchor. Trim the claim or pin the anchor.

**New 2 (low, pre-existing, not this PR) — `cli/src/commands/status.ts:161-166`.** Probe with production wiring (`setExpectedFile` as `figma-agent.ts:210` does), two `Untitled` files connected: `status --file Untitled` **throws** `E_INVALID_ARGS: --file "Untitled" matches 2 connected files [...]`. The catch list covers only `E_NO_PLUGIN`/`E_WRONG_FILE`/`E_APP_UNREADY`, so the diagnosis tool exits 1 in exactly the ambiguity this PR is about — against the file's own header contract. Follow-up issue, not a blocker.

**New 3 (medium, follow-up) — `gapfill-errors: 2` for one refusal is a magnitude an agent can misread.** `readFileBaseline` returns the same "belongs to another file … treated as missing" string at boot (`edit-gapfill.ts:174`) and again on `writeBaseline`'s read-back (`edit-gapfill.ts:205`); `recordGapfillError` increments both but keeps only `firstError` (`gapfill-status.ts:60-63`). The reply is `errorCount: 2` with **one** message, so nothing distinguishes one refusal counted twice from two distinct failures. The row's boolean claim ("gap-fill could not do its job") is true; the count is not measured. The new test now hardens that number into a contract — file it before it becomes load-bearing. Fix at source: don't re-record the boot refusal on the write read-back, then re-pin.

**Verdict: Approve.** Two follow-up issues (New 2, New 3) and one comment trim (New 1); none blocks the merge.

PR #120, opened 2026-09-03
fix(cli): diagnose an ambiguous --file in status and count a refused baseline once

### LINKED ISSUE #118: status --file with an ambiguous name exits 1 instead of diagnosing
`figma-agent status --file Untitled` with two connected files named Untitled throws `E_INVALID_ARGS: --file "Untitled" matches 2 connected files` (cli/src/commands/status.ts:161-166 catches only E_NO_PLUGIN / E_WRONG_FILE / E_APP_UNREADY). The file's own header says status is a diagnosis tool that never fails when nobody is listening; the same should hold when two are. Found by the stage-4 re-review of #117 with a production-wired probe (`setExpectedFile`). Pre-existing, not introduced by #117.

Expected: report both matching rows (`pluginsAll`), `plugin.connected:false`, and `coverage.complete:null`, exit 0.

### LINKED ISSUE #119: gapfill-errors counts one refused foreign baseline twice
A baseline that belongs to another file is refused at boot (plugin/src/main/edit-gapfill.ts:174) and again on the write read-back (edit-gapfill.ts:205); `recordGapfillError` increments `errorCount` both times but keeps one `firstError`, so STATUS reports `errorCount: 2` with one message and `coverage` shows `gapfill-errors: 2` for one cause. The boolean claim is true; the magnitude is not measured. tests/edit-gapfill-baseline.test.ts pins the current `2` (added in #117) so the contract does not drift silently.

Fix at source: do not re-record the boot refusal on the write read-back, then re-pin the test at 1. Found by the stage-4 re-review of #117.

jangtrinh:
## Why

Two wrong numbers in `figma-agent status`, both found by the stage-4 review of #117. With two connected files named `Untitled`, `status --file Untitled` exited 1 with the broker's ambiguity refusal, although the command's own contract is a diagnosis tool that never fails when nobody is listening. And a baseline belonging to another file was recorded as a gap-fill error twice for one refusal, so `coverage` said `gapfill-errors: 2` for one cause.

## What

- `status --file X` on an ambiguous name reports `plugin.state: "ambiguous"`, `plugin.ambiguous: N`, the N matching rows in `plugins`, `coverage.complete: null`, exit 0. The refusal keeps its wire text; it now comes from one shared formatter with one predicate, so the CLI never string-matches a loose substring.
- The boot read's refusal is authoritative: the write read-back of the same key no longer re-records it. A genuinely distinct write failure still counts.

## Evidence

Test-first, both REDs quoted in `plans/260904-0616-status-diagnosis-and-gapfill-error-count/report.md`. Gates after the last edit: typecheck 0 · build 0 (committed plugin artifacts fresh) · lint:comments 0 (0 new) · `npm test` 172 files / 2405 passed (+4 across two review rounds: the ambiguous branch keeps no file in view so every connected identity is counted, and the boot read verdict is the one dedupe key so an idle write never re-records it). Not verified live: the two-`Untitled`-windows case (vitest scratch broker only).

Closes #118, closes #119.


jangtrinh:
## Stage-4 review — PR #120

Verified read-only (no `npm test`/`npm run build`): typecheck 0 · `lint:comments` clean (0 new) ·
`plugin-build-freshness` green (committed `code.js`/`ui.html` match a fresh build; the changed
`BUILD_LINE` id is a content hash, not a git sha) · 172 test files, +2 tests, matches the body.
**RED proved**, not taken on trust: pre-fix tree (`a8eb75f`) + this PR's test files → all three fail
(`two true rows, one cause`: gaps differ · `a GENUINE second failure…`: **3** vs 2 ·
`two files sharing a name diagnose as ambiguous`: throws). Both fixes are real and land where #118/#119 asked.
`--peek` short-circuits at status.ts:63, untouched; the non-ambiguous paths are a no-op spread.
`ambiguous = plugins.length` is sound: broker `registry.matching()` and the CLI filter both run
`fileMatches(…, exact:true)` over `liveEntries()`, so the count describes the rows actually printed.

### Blocking

**1. `cli/src/commands/status.ts:233` — the ambiguous reply emits a wrong count, in a PR about wrong counts.**
Probe on this branch (two `Untitled`s, KEY_A/KEY_B):
`coverage = {"complete":null,"gaps":[{"kind":"other-files-connected","count":1,"see":"status.pluginsAll"}]}` with `ambiguous: 2`.
Nothing answered (`scene = null`), so **neither** file's edits are in view — but the roll-up still runs
`activeKeySource = plugins[0]` (status.ts:206), makes one of the two ambiguous matches "the file in view",
and reports the other as the only missing one. Worse, by code read: `fileRows` is then just that one row,
so the *other* match's `relayDroppedFrames`/`replayedBatches` are summed out of the reply entirely — a
dropped-frame count that vanishes without a counter (Law 1). Before this PR the path threw, so the PR
introduces the emission.
Smallest fix: in the `ambiguousRefusal` branch, force the roll-up to "nothing in view" — set a flag and
use `activeId = null` (then `fileRows = []` and `otherFiles` counts both identities → `other-files-connected: 2`).

**2. `plugin/src/main/edit-gapfill.ts:213-218` — the new comment states a fact that is false.**
It claims the idle-debounce caller "has no prior read this session … there is nothing to deduplicate against".
There is. Reachable sequence, measured on this branch: foreign baseline + boot write refused (quota) →
`errorCount 2`; then one idle `writeBaseline(…, dirty)` (main.ts:253, no `bootReadError`) → `errorCount 4`,
because the read-back re-records the *identical* foreign refusal `firstError` already holds. One cause, +1
per idle window — the exact wrong magnitude #119 objects to, just moved off the boot path.
Smallest fix: correct the comment to say the idle path is deliberately out of scope, **or** fix it at the
shared layer (see below) and delete the parameter.

### Non-blocking

- **Mechanism isn't pinned by the tests.** Counter-mutation: revert `edit-gapfill.ts` to `a8eb75f` and add
  3 lines to `recordGapfillError` (`gapfill-status.ts:60`) skipping a message already recorded → all 63
  gapfill tests pass, including both new ones. That variant also fixes finding 2 for free and drops the 7th
  positional parameter threaded through three call sites. The chosen approach is the more conservative
  semantics (two genuinely separate events with one message still count twice) — defensible, but the tests
  do not distinguish them, so the doc comment should not read as if they did.
- `ambiguous` is the hello snapshot's count, not the refusal's own `N`; a window opening between the hello
  and the round-trip makes them disagree. Internally consistent with the printed `plugins[]`, so no action —
  noting it so the next reader doesn't "fix" it by parsing the message.
- The predicate marker (` connected files [`) is safe today: no other `E_INVALID_ARGS` producer echoes a
  file name (checked all 9 in broker-daemon.ts + plugin-side ones). Keep it that way.
- PR body cites `plans/…/report.md` for the quoted REDs; that file does not exist (`plans/` is gitignored
  anyway). The claim is substantively true — I reproduced all three REDs — but the citation points at nothing.
- Pre-existing, not introduced: `edit-gapfill.ts` 494 lines, `status.ts` 260 vs the repo's 200-line guidance.

**Verdict: Request changes** — both blockers are ≤5 lines and are the same class of defect this PR exists to fix.

jangtrinh:
**Re-review of `5f2c8b7` — both blockers fixed, each with a test that is RED against `2d871d2`.**

RED proof method: pre-fix `cli/src/commands/status.ts` and `plugin/src/main/edit-gapfill.ts` (from `2d871d2`) aliased in behind the new tests; the rest of both files' 78 tests stayed green, so the two new tests are the only behavioural pins.

**1. Ambiguous `--file` attributed an active file — FIXED.** `cli/src/commands/status.ts:212` (`ambiguous !== undefined ? null : …`) ⇒ `activeId = null` ⇒ `fileRows = []`, so no identity is excluded and no relay/replayed tally is attributed to a guess. Test `tests/session-coverage-status-command.test.ts:338` — 2 `Untitled` + 1 other now counts **3**; RED on pre-fix source (got `2`). Non-ambiguous paths are untouched (`ambiguous` is set only in the `isAmbiguousFileErrorMessage` branch; the old expression is preserved verbatim under the new ternary).

**2. Idle read-back re-recorded the boot refusal — FIXED.** `GapfillStats.bootReadError` (`plugin/src/main/gapfill-status.ts:55`), set once at `plugin/src/main/edit-gapfill.ts:388`, compared in `readBaseline` at `:185`. Test `tests/edit-gapfill-baseline.test.ts:550` — boot read + boot write refusal + idle write refusal = **3**, not 4; RED on pre-fix source (got `4`).

Extra checks, all pass:
- Write-side failures still count every time (the idle `set` refusal is the 3rd count in the same test).
- A genuinely different later read failure still counts: probed boot foreign refusal (1) → later `baseline read failed: clientStorage offline` + write-skipped = 3. Key is the message, and the foreign message embeds the key + both identities, so a mid-session rename cannot alias two different keys' refusals.
- No payload leak: `toGapfillStatus` is an explicit allowlist, `GapfillStatus` (shared/protocol) unchanged, `tests/executor-ops-status.test.ts` green.
- `plugin/code.js` rebuilt and byte-consistent with the source change, no unrelated bundle drift; `plugin-build-freshness` green. `npm run typecheck`, `npm run lint:comments`, and the 4 named test files: green (106 tests).

New, non-blocking:
- `tests/session-coverage-status-command.test.ts:338` pins the identity count but every fake row has `relayDroppedFrames: 0`, so "no tally attributed" is only proven by construction. Seeding a non-zero drop on one `Untitled` row would pin it directly.
- Dedupe is boot-verdict-only by design: after a CLEAN boot read, an identical repeating read failure still counts on every idle write (probed 2 → 4). Matches the ruling; worth one clause in the `bootReadError` doc so the asymmetry reads as chosen, not missed.
- Three comments cite `main.ts:253`; accurate today, but line numbers rot — naming `createIdleBaselineWriter` would not.

Verdict: **Approve**.

— stage-4 re-review round 1, posted by the controller session

Issue #121, opened 2026-09-03
exec-js --undo-group leaks its sentinel frame deletion into the owner's edit feed


jangtrinh:
Every `figma-agent exec-js --undo-group` run leaves a frame like `{"op":"deleted","nodeType":"FRAME","nodeName":null,"changedProps":["height","name","pluginData",…],"actor":"ambiguous"}` in the bound project's edit feed (observed 2026-09-04 06:31 on the owner's file, two runs → two frames, node ids 15506:314305/314306). That is the plugin's own invisible undo sentinel being removed, not a designer edit; the sentence layer renders it as "Deleted a FRAME node". The bookkeeping filter in plugin/src/main/document-change-capture.ts drops pluginData-ONLY property changes, not the sentinel's CREATE/DELETE.

Expected: the sentinel's own lifecycle never enters the feed (tag the sentinel id at creation and drop its changes, counted in `pluginDataChangesDropped` or a sibling counter — nothing vanishes silently).

PR #122, opened 2026-09-04
feat(plugin): carry designer intent on live edit-feed frames

jangtrinh:
## Why

The edit feed told an agent that a component's `description` or a node's `annotations` changed, and nothing more. What the designer wrote is the part that matters, and Figma's REST API returns an empty description at every version, so the plugin event is the only place it can be read. A live probe on the owner's file showed which property names Figma actually emits (`description`, `annotations`; nothing for `descriptionMarkdown`, `devStatus` or `documentationLinks`), which set the scope.

## What

- Live feed frames whose `changedProps` name `description` or `annotations` carry `intent`: the new description (and its markdown twin), or the annotation list, read at capture time through the same readers `context` uses. A refused read keeps the property name and adds `intentReadError` naming the field; a cleared description or annotation list is a value (`''`, `[]`), not an absence.
- Bounded: text capped at 2 000 characters and annotation lists at 20 entries, each cut leaving `intentTruncated` or `annotationsTotal`.
- Sentences: `Described component "X"`, `Cleared the description on "X"`, `Annotated "X" (2)`, `Cleared the annotations on "X"`; `renamed` keeps precedence when a batch also renames.
- `figma-agent changes` passes `intent` through; a frame without intent is byte-identical to before; older feed lines validate unchanged; the emitted skill states the shape, the caps and the honest limit that closed-window description or annotation edits are not reported at all.

## Evidence

Gates after the last edit: typecheck 0 · build 0 (committed plugin artifacts fresh) · lint:comments 0 (0 new) · emit:skill drift green · `npm test` 173 files / 2486 passed (+81 for this feature across two review rounds).

Live on the owner's file (plugin hot-reloaded from this build): a component given a description, a markdown description and two annotations in one batch produced one feed frame with `intent {description, descriptionMarkdown, annotations[2]}`, and `figma-agent changes --file <key>` returned it intact.

Review: stage-4 adversarial review (Request changes: one Critical — a cleared description shipped as "Described" with no value, indistinguishable from an unreadable one — plus three Important: uncapped annotation text, a wrong gap-fill claim in the skill, a read description discarded when its markdown twin refused), every finding fixed test-first with its RED quoted; the cleared-description path then verified live (`Cleared the description on "…"` with `intent {description:""}`). Mutation testing: 20 wrong-fact mutants, 19 killed by the existing suite, 1 boundary survivor (`annotationsTotal` at exactly 20) now covered. Round two (Approve with one Important) tightened the wire guard so a self-contradicting `annotationsTotal` is refused, applied the caps at the write boundary rather than in one producer only, and made a cleared description read as cleared only when its markdown twin is empty too.

## Not in this PR

`devStatus` and `documentationLinks` changes (Figma emits no documentchange property for them); `cowork` output keeps its sentences but does not carry `intent`; the undo-group sentinel leak in the feed (#121).


jangtrinh:
## Stage-4 review — round 2 (committed diff `ae1772f..3019b79`)

### Summary

Round-1 findings are all fixed, and each fix is nailed down by a test that goes RED when the fix
is removed — verified by reverting in an extracted copy of the commit, never in the working tree:

| Round-1 finding | Fix | RED proof (revert → fail) |
|---|---|---|
| **C1** cleared description indistinguishable from unreadable | `shared/edit-vocabulary.ts:171` cleared branch, driven by the **value** `description === ''` | removing the branch fails `edit-vocabulary.test.ts` › *says the description was cleared rather than claiming it was written* **and** *keeps the location on a cleared description too* (`Described component "Button / Primary"` ≠ `Cleared the description on …`) |
| **I1** uncapped annotation text | `capAnnotation` over `label`/`labelMarkdown` at `INTENT_TEXT_CAP` | stubbing `capAnnotation` to a pass-through fails `edit-intent.test.ts` › *caps a label and its markdown twin…* + *holds a 50 000-character label to a frame a feed can carry* |
| **I2** wrong gap-fill claim in the skill | now "not reported at all" | checked against the code, not the prose: `gapfill-baseline-store` records `[id,name,type,x,y,parent]` and `gapfill-diff` emits only created/deleted/renamed/moved — there is no description or annotation in the baseline to diff. Claim is accurate. |
| **I3** description discarded when the markdown getter refuses | two separate `try` blocks, `intent.description` assigned before the markdown read | folding both reads back into one `try` fails `document-change-capture.test.ts:711` › *a refused markdown read keeps the description that succeeded* (`{intentReadError:"description: not loaded"}` instead of `{description:"The primary action", intentReadError:"descriptionMarkdown: not loaded"}`) |

Both deviations are sound. **S2** (`mergeIntent` treating `annotations`+`annotationsTotal` as one
fact) is the right call — a re-cap at merge time would have deleted a true total; the coupling is
one line, is stated in the comment, and is covered in both directions (*drops a stale
annotationsTotal…* / *keeps the later read's OWN total…*). Shipping a cleared description as
`{description:''}` with no empty markdown twin is also right: an empty twin would carry no fact
(see Suggestion 2 for the one case where the sentence should notice the twin).

Other checks, all clean:

- **`context` output unchanged** by the `context-intent-readers.ts` extraction — `distinctMarkdown`
  is the previous inline condition verbatim, `shapeAnnotations` is the previous `map` body verbatim
  and is still called only after the same `!Array.isArray(raw) || raw.length === 0 → null` guard, so
  `readAnnotations`/`readComponentIntent` are behaviour-identical; `tests/context-intent.test.ts` green.
- **Byte identity of non-intent frames** is asserted against a literal string, not a key set
  (`tests/edit-feed.test.ts:299`), plus `expect(frame).not.toHaveProperty('intent')`.
- **Read side**: strip-after-guard in `readEditFeed`, `intentWarnings` counted separately from
  `warnings` and emitted only when non-zero; the edit itself always survives.
- **No DOM/Node globals** in `shared/edit-intent.ts` or `plugin/src/main/edit-intent-reader.ts`.
- **Claims vs diff**: `+69` is exact (69 `it(` added, 0 removed); 173 test files matches; catalog
  text and `SKILL.md` agree (`skill-emitter-drift` green) and the committed plugin artifacts are
  fresh (`plugin-build-freshness` green); `typecheck` 0, `lint:comments` clean (0 new). The 2 474
  total was not re-run here (full `npm test` is off-limits in this session) — the 8 suites touching
  this feature are green, 236 tests. The mutation table matches the tests actually present, and the
  M6 survivor is covered by *omits annotationsTotal at exactly the cap*.

### Risk level

**Low.** Additive optional field, guarded on write and on read, and the one gap below is not
reachable from the shipped producer.

### Findings

#### Important — 1. `isValidEditIntent` admits an `annotationsTotal` that contradicts the list it counts, and the sentence then states a wrong fact

`shared/edit-intent.ts:136` checks only `typeof number && Number.isFinite`. `shared/edit-vocabulary.ts:182`
prefers that number over the carried list, so an internally inconsistent block becomes a confident
English claim. Probed against the committed code:

```
intent = { annotations: [{label:'a'},{label:'b'}], annotationsTotal: 0 }
isValidEditIntent(intent) === true
editSentence(...) === 'Cleared the annotations on "Search field"'   // 2 annotations ride on the same frame

intent = { annotations: [{}], annotationsTotal: -3.5 }
editSentence(...) === 'Annotated "X" (-3.5)'
```

`capIntent` can never produce either, so this needs a malformed or foreign feed line — which is
precisely the input class the guard's own docstring claims it exists for ("untrusted wire input…
a field it cannot vouch for is not admitted"). Second surface: the broker hands `data.edits` to the
cowork waiters as `EditInput[]` with no validation at all (`broker-daemon.ts:2297`, pre-existing),
and `cowork.ts:73` now renders sentences from `intent` — so a block `buildEditFrame` would refuse to
write to disk can still shape a cowork sentence.

Smallest fix — the count only ever exists because the list was cut, so it must exceed it:

```ts
if (r.annotationsTotal !== undefined
  && (!Number.isInteger(r.annotationsTotal)
    || !Array.isArray(r.annotations)
    || r.annotationsTotal <= r.annotations.length)) return false;
```

#### Suggestion — 2. A cleared description with a markdown twin that still says something reads as "Cleared"

`shared/edit-vocabulary.ts:171` fires on `description === ''` alone. The reader deliberately keeps
both fields in that case (*an empty description with markdown that still says something keeps
both*), so:

```
intent = { description: '', descriptionMarkdown: '[docs](https://x)' }
editSentence(...) === 'Cleared the description on "Button"'   // the frame still carries words
```

Fix: `&& frame.intent.descriptionMarkdown === undefined` on the cleared branch — then it degrades to
`Described component "Button"`, which the frame's own value supports.

#### Suggestion — 3. The 2 000/20 caps are producer-side only; the boundary that validates does not enforce them

`buildEditFrame` (`shared/edit-feed.ts:172`) checks shape, never size, so the cap the catalog states
as a contract to agents holds only because today's plugin applies `capIntent` before sending. A
mis-built or differently-versioned plugin writes an unbounded quote into the append-only feed.
`capIntent` is pure and idempotent on already-capped input (a 20-entry list with `annotationsTotal:31`
survives untouched), so the fix is one call: `...(isValidEditIntent(e.intent) && { intent: capIntent(e.intent) })`.

#### Suggestion — 4. The whole annotation list is shaped before it is capped

`plugin/src/main/edit-intent-reader.ts:75` runs `shapeAnnotations(raw)` over every entry and
`capIntent` then discards all but 20 — inside the `documentchange` pass, which is the hot path.
Slice first and carry the true length (`annotationsTotal: raw.length` when over cap, which `capIntent`
then leaves alone) if this ever shows up on a heavily annotated file.

#### Suggestion — 5. `intentWarnings` is a new `changes` output key documented nowhere

The catalog entry documents `intent`, the caps and `intentReadError` thoroughly but never mentions
`intentWarnings`. (The pre-existing `warnings` is undocumented too, so this matches precedent — but
an agent that sees `intentWarnings: 2` has nothing to read.) One clause in the same sentence.

### Verdict

**Approve — no blockers.** The round-1 fixes are real and pinned by RED-verified tests, the
deviations are justified, and nothing in the non-intent path changed byte-wise. Finding 1 is a
one-line guard tightening plus one test and matches the module's own stated contract — worth doing
before merge; findings 2–5 are fine as follow-ups.

PR #123, opened 2026-09-04
fix(plugin): keep the undo sentinel's own lifecycle out of the edit feed

### LINKED ISSUE #121: exec-js --undo-group leaks its sentinel frame deletion into the owner's edit feed

jangtrinh:
## Why

Every `figma-agent exec-js --undo-group` run left a frame in the bound project's edit feed: `{"op":"deleted","nodeType":"FRAME","nodeName":null,"actor":"ambiguous"}`, rendered as "Deleted a FRAME node". That is the plugin's own invisible undo sentinel being removed, not a designer edit. The bookkeeping filter only dropped pluginData-only property changes; the sentinel's create and delete passed through.

## What

- A bounded registry (last 32 ids) records each sentinel's id at creation; the capture handler drops any change carrying a registered id before the correction store, the connector index and the feed see it. Release is lazy by design: the delete's `documentchange` arrives after the synchronous `remove()`, so no call site can safely forget the id at once; the bound and its reachability are argued in the module comment.
- Every drop is counted: STATUS `capture.sentinelChangesDropped`, present only when non-zero, documented as deliberately unmapped in coverage (the plugin's own node is not a designer edit).
- A frame that merely shares the sentinel's name is still reported: name is not identity.

## Evidence

Test-first, REDs quoted in `plans/260904-0750-undo-sentinel-feed-leak/report.md`. Gates after the last edit: typecheck 0 · build 0 (committed artifacts fresh) · lint:comments 0 (0 new) · `npm test` 175 files / 2500 passed (+14). Stage-4 review found the stray-sweep path removing leftover sentinels without registering their ids (a stranded sentinel after a failed commit or a plugin reload would still leak); fixed with an order-sensitive test proving registration precedes removal. Live on the owner's file: two `--undo-group` runs added no frame to the bound feed (5295 lines before and after) and STATUS reported `sentinelChangesDropped: 6`.

Closes #121.


jangtrinh:
## Stage-4 review — Request changes

The consumer half is right: the drop sits at the top of the loop, counted, id-matched, and STATUS stays byte-identical at zero. But the **producer half misses the sweep path**, which re-opens the exact symptom of #121, and it has no test at all.

### Blocking 1 — the stray sweep removes sentinels it never registered, so their DELETE still reaches the feed

`plugin/src/main/executor-exec-js.ts:52`

```ts
for (const n of page.findChildren((c) => c.getPluginData(SENTINEL_KEY) === '1')) n.remove();
```

`registerSentinel` is only called for the frame this run creates (line 62). A leftover sentinel swept here is removed with its id unknown to the registry, so `isSentinelId(node.id)` is false for its DELETE.

Failing input → wrong output: file carries a leftover frame with `figmaAgentUndoSentinel === '1'` on the page (commit()'s `remove()` is swallowed on throw, `executor-exec-js.ts:70`; a run interrupted by a plugin close/reload never commits at all), plugin closed and reopened — `tracked` is module state, so a new plugin session starts empty — then one `exec-js --undo-group` → the feed gains `{"op":"deleted","nodeType":"FRAME","nodeName":null}` ("Deleted a FRAME node"), and `sentinelChangesDropped` does not count it. Nothing vanishes loudly here; it appears silently, which is the bug.

Smallest fix (one line, same module, still plugin-data-matched so a lookalike is untouched):

```ts
for (const n of page.findChildren((c) => c.getPluginData(SENTINEL_KEY) === '1')) {
  registerSentinel(n.id); n.remove();
}
```

This also closes the eviction question without an unbounded Set: a stranded id that aged out of the 32-deep FIFO is re-registered at the instant it is swept, i.e. before the DELETE it must catch. (The pure eviction race — 32 registrations landing between a sentinel's `remove()` and its own `documentchange` — I agree is not reachable; the reachable path is the stranded/cross-session one above.)

### Blocking 2 — `figmaUndoBracket()` has zero test coverage, so the registration is unproven

`grep -rn figmaUndoBracket tests` → 0 hits; `tests/exec-undo-group.test.ts` only exercises the pure `runInUndoGroup` against a spy bracket. All 10 new tests register ids by hand, so nothing proves the id that gets *created* is the id that gets *registered* — that half rests on the live smoke alone. Please land the fix above with a `figmaUndoBracket()` test on a fake figma asserting (a) the created frame's id is registered, (b) every swept stray's id is registered **before** `remove()`.

### Verified as claimed

- Drop position (`document-change-capture.ts:176`) precedes every consumer: `corrections.record`, `connectorTouched.push`, `resolveComponentIdentity`/`raw`, `notePageDirty`, the new `readEditIntent` reader, `edits.push`, `identity.remember`, and therefore `post` and `armIdle`. Only `onBatchStart` (per-batch prune + read-only-guard record) and the allocation-only `corrections.begin()` run earlier — identical to the pre-existing pluginData drop, so no behavior change there.
- STATUS: `executor-ops.ts:151` gates on `> 0`; both omission tests assert `'sentinelChangesDropped' in status === false`. Coverage no-row documented at `session-coverage.ts:70`.
- Lookalike frame still reported (`tests/document-change-capture.test.ts:163`).
- New module is a plain `Set` — no DOM/Node globals. `npm run typecheck` (3 projects) clean; `undo-sentinel-registry` + `document-change-capture` + `executor-ops-status` + `plugin-build-freshness` = 77 passed; `plugin/code.js` carries the change (registry at 1527, drop at 1580, registration at 6275). PR body's "+10" matches the diff (5+4+1); the 2496 total was not re-run here (`npm test` is off-limits against the live broker).

PR #126, opened 2026-09-04
feat(plugin): fold a typed description into two feed frames

### LINKED ISSUE #124: Typing a description emits one feed frame per keystroke
Observed live 2026-09-04 09:14 on the owner's file: typing "Page content of IDP Review queue" into a component's description produced 17 `documentchange` batches in 5 s, so the bound feed holds 17 frames `Described component "State=Default" …` with `intent.description` growing one letter at a time ("Pa", "Page content", …, "Page content of IDP Review queue"). Every frame is a true fact, but an agent reading `figma-agent changes` sees 17 sentences for one edit, and the cowork waiter wakes 17 times.

The per-batch coalescer cannot merge them (each keystroke is its own batch). Proposed: a short quiet window (≈1.5 s) in the capture handler for intent-only frames of the same node — post one frame with the final value and `coalescedFrames: N` so nothing vanishes silently. Layout drags already show the same repeated-pair pattern (`Moved text` / `Restyled instance` ×4 in the same minute); whether the window should cover all live frames is an owner call, since it delays every frame by the window.

jangtrinh:
## Why

Figma emits one `documentchange` per keystroke, so typing "Page content of IDP Review queue" into a component description put 17 frames into the owner's feed in five seconds, each a true fact and together a wall of `Described component …` an agent has to wade through while `cowork` woke 17 times.

## What

- Frames whose changed properties are only `description` or `annotations` post their first keystroke immediately, exactly as before, so `cowork` and the idle prompt still arm on the first key. Later keystrokes fold into a per-node held frame that posts once, 1.5 s after the last keystroke, with the final value, `coalescedFrames` (total keystrokes, first included) and `capturedAt` dated at that last keystroke; the page is re-read at flush. Two frames per typed description instead of one per letter; every other frame keeps today's timing.
- Held frames are bounded to 64 nodes, the oldest flushed rather than dropped; a later non-intent edit or a delete of a held node flushes it first so order holds. STATUS counts `intentFramesCoalesced` and `intentFramesParked` (present only when non-zero); a description typed in the last 1.5 s before the panel closes is never reported, and the emitted skill says so.
- `coalescedFrames` rides the feed and `figma-agent changes`; a malformed count is stripped and counted, never the edit.

## Evidence

Gates after the last edit: typecheck 0 · build 0 (committed plugin artifacts fresh) · lint:comments 0 (0 new) · emit:skill drift green · `npm test` 175 files / 2538 passed (+38 for this feature). Round two of review (Approve) added two guards in the same PR: a producer-set `capturedAt` survives the relay buffer during a broker outage, and a refusal while posting a held follow-up is counted instead of escaping the timer.

Live: the 17-frame storm was observed on the owner's file before the change (feed lines with `intent.description` growing one letter at a time). The folded shape is pinned by tests with a fake clock and a fake timer with cancel semantics (17 batches → 2 frames, `coalescedFrames: 17`, counter 16); the live re-type on the owner's active file is still owed and will be added as a comment.

Review: stage-4 adversarial review, round one on the uncommitted tree (Request changes): the pure trailing window left `cowork` blind to a designer mid-sentence (fixed by the leading-edge fold above), `coalescedFrames` never reached the CLI, held frames were dated at flush, and a malformed count rejected the whole edit — all fixed test-first with the REDs quoted in the plan report. Mutation testing: 21 wrong-fact and wrong-timing mutants, 21 killed by the existing suite, 0 survivors.

Closes #124.


jangtrinh:
## Stage-4 re-review — round 2 (commit `2816951`, base `ab61650`)

### Summary

Every round-1 ruling landed, each with a test that fails without it. The leading-edge fold is real (first intent-only frame posts in-batch, later keystrokes fold into one follow-up 1.5 s after the last, `coalescedFrames` counting all N), the follow-up is dated at the last keystroke on **both** feeds, `coalescedFrames` reaches `figma-agent changes`, and a malformed count is stripped on write and on read instead of costing the edit. Two rulings were re-verified by mutation in an extracted copy, not taken on the report's word:

- **C3** — trailing-edge mutant (`fold()` returns `true` on the leading edge, follow-up posted at `frames >= 1`): 21 tests fail across `document-change-capture` + `cowork-waiter`, including the exact ruling test `tests/cowork-waiter.test.ts › a typed description arms the waiter on the FIRST keystroke › the waiter is armed at the first keystroke, and stays armed through the typing`.
- **I1** — `capturedAt: frame.capturedAt` removed from both posts in `postParkedFrame`: exactly one test fails — `tests/document-change-capture.test.ts › the follow-up is stamped with the LAST keystroke's time, on both feeds`.

Ruling → killing test: **C2** `tests/changes-command.test.ts › passes a real count through, strips one that says nothing, and leaves every edit listed` · **I2** `tests/edit-feed.test.ts › a meaningless count is left OFF the frame rather than written, and never costs the edit` + `› admits an edit whose count is meaningless — the edit is never the casualty` + the `coalescedWarnings: 1` assertion in the changes test · **S1** `› a window whose canceller does nothing still posts nothing a second time` (canceller is a genuine no-op `() => {}`; the fired timers are replayed afterwards) · **S2** honest comment at `document-change-capture.ts:112-121`, behaviour pinned by `idleArmed === 2` in the 17-keystroke test · **S4** `› the follow-up carries the page the node is on at flush, on both feeds`, `› a node with no page left at flush keeps the page of its last keystroke, never a guess`, `› a re-read that THROWS at flush keeps the recorded page and counts the refusal` (the mock throws on `parent`, the way an invalidated node really refuses) · **S5** the `executor-ops.ts:134-175` bullets now match the spread order.

Contract checks: the broker honours a payload `capturedAt` for **both** kinds — `cli/src/transport/broker-daemon.ts:346` `captureTimestamp`, used at `:370` (DOC_CHANGE) and `:403` (EDIT_FEED); `shared/protocol.ts:112-118` refuses only a FUTURE stamp (`> now + 60 s`), so a follow-up dated 1.5 s in the past is always honoured. `resolvePageAtFlush` is guarded and its refusal is counted, never swallowed. Counters are symmetric (`intentFramesParked` +1 exactly when a window first holds something, −1 in the single exit `closeSlot`; `intentFramesCoalesced` = N−1). `intent-frame-parking.ts` touches no host global — `Map`/`Set`/`Number`/`Infinity` only, the timer injected.

PR claims vs the diff: `npx vitest list` collects **2534**; the test diff adds **34** `it(` and removes 0; `mutation.md`'s 21/21 is consistent with my two independent reproductions (its #1 and #8 match what I measured). `npm run typecheck` clean, `npm run lint:comments` clean (0 new), and the panel/emitter/freshness tests pass, so the committed bundle carries the feature.

### Risk level: low

### Findings

**Important — 1. A folded follow-up buffered during a broker outage loses the stamp this PR just gave it** (`plugin/src/ui/outbound-buffer.ts:55-61`, called from `plugin/src/ui/ui-relay.ts:168`)

`stampCapturedFrame` writes `{ ...data, capturedAt, replayed: true }`, so an existing `capturedAt` is overwritten by the enqueue time. Before this PR that was harmless: every bufferable frame was enqueued in the same tick it was captured. A parked follow-up is not — it is enqueued at the flush, up to 1.5 s after the keystroke it stands for.

Failing input: broker down; the designer types on node A from `T` to `T+4.0s`, then moves node B at `T+4.2s`; the follow-up is enqueued at `T+5.5s`. Wrong output: on replay the description frame is filed at `T+5.5s` and sorts **after** B's move — the feed-order inversion ruling I1 was raised to remove, and `changes --since T+5s` now reports a description edit that happened at `T+4s`.

Smallest fix — preserve a stamp the producer already set:

```ts
const own = data['capturedAt'];
const ts = typeof own === 'number' && Number.isFinite(own) && own > 0 ? own : capturedAt;
return JSON.stringify({ ...msg, data: { ...data, capturedAt: ts, replayed: true } });
```

Test that fails pre-fix, alongside the existing ones in `tests/replayed-capture-time.test.ts`: "a frame that already carries its capture time keeps it through the buffer". Non-blocking for merge (outage-only, ≤ 1.5 s), but it is a wrong fact with no counter, so it should not stay open as folklore — fix here or file it.

**Suggestion — 2. The flush path is the one exit that can lose a frame after the gauge already said it was gone** (`plugin/src/main/document-change-capture.ts:122-156`, `plugin/src/main/intent-frame-parking.ts:108-114`)

`closeSlot` decrements `intentFramesParked` and then calls `deps.flush`. Inside `postParkedFrame` only `resolvePageAtFlush()` is guarded; `deps.post`, `figma.fileKey`, `figma.root.name` and `deps.armIdle()` are not, and this runs in a bare `setTimeout` callback rather than in Figma's `documentchange` dispatch. `mutation.md` #10 is the evidence that a throw here escapes rather than degrades (removing the `resolvePageAtFlush` try/catch "crashed the test"). A refusal at flush therefore loses the held frame **and** leaves the gauge reading zero — the exact silent-loss shape the rest of this module forecloses. Likelihood is low (all four reads are the same ones the batch path already makes); the fix is a try/catch around the body of `postParkedFrame` routing to `recordCaptureError`, so a refusal is counted the way every other refusal in this pass is.

**Suggestion — 3. The two items the implementer left open, judged**

- *An agent's own description write waiting 1.5 s* — it does not, in the case that matters: a single programmatic write is one batch, so it is a leading edge and posts immediately with its final value. The wait applies only when one agent op produces two or more intent-only batches for the same node (description then annotations, or two writes). Bounded at 1.5 s, the first frame already carries a real value, and both counters are visible. Recommend no change. If it ever bites, the narrowing is one predicate at `document-change-capture.ts:341` (`&& edit.actor !== 'agent'`) — but it would make `intentFramesCoalesced` mean something narrower, in exchange for latency the agent itself controls.
- *The relay re-stamp* — real, and promoted to finding 1 above rather than left as a note, because this PR is what turns it from harmless into a wrong timestamp.

**Suggestion — 4.** `plugin/src/main/intent-frame-parking.ts` is 201 lines against the repo's 200-line guideline, of which ~60 are code. Do not split it: the guideline targets complexity, and cutting the slot table apart from the window would produce exactly the one-off abstraction the house rules warn against. Noted so the next reader does not "fix" it.

### Verdict

**APPROVE.** No blocker. Finding 1 is a small follow-up in a pre-existing file — land it here or as its own issue; findings 2-4 are optional. The live re-type on the owner's file is still owed, as the PR body already says.

PR #130, opened 2026-09-05
fix(broker): reject chunks for admitted requests

### LINKED ISSUE #127: fix(broker): reject request chunks after admission
# Reject post-admission request chunks
## Problem and outcome
The broker fully assembles a chunked request before admission, but its existing-job path still treats later chunks as continuations. Reject those frames before they can reach a plugin or mutate queued request bytes. Preserve legitimate request ownership and the existing raw-file-key mutation gate.

## Scope
- Modify cli/src/transport/broker-daemon.ts and behavior regression coverage in tests/broker-daemon-harness.test.ts.
- Use current retained jobs and parked requests for ownership checks.
- Require the outer chunk ID to match the assembled request ID before admission.
- Keep protocol types, plugin code, queues, job polling and retention unchanged.
- Do not add permanent request-ID history or a new lifetime quota. After existing history expires, a body must fully assemble and pass ordinary admission again.

## Acceptance
1. Valid initial chunked requests still assemble and dispatch once.
2. Later chunks for running, queued, parked or any retained terminal job never forward, append, change ownership or extend a parked deadline. Cover same and different sockets.
3. Same-socket violations leave the original request/result intact, emit a diagnostic, and do not send a misleading correlated terminal reply or close the legitimate owner. Different-socket duplicates receive the existing refusal without affecting that owner.
4. Mismatched outer/assembled IDs create no job, parked request or plugin dispatch. Expired history re-enters normal admission; pause, FIFO and exact-file guards still apply.
5. New regressions fail on the unchanged baseline, pass after the patch; polling, reconnect and full project gates remain green.

## Verification and delivery
Focused: npx vitest run tests/broker-daemon-harness.test.ts.
Then npm run typecheck, npm run build, npm test, npm run lint:comments and git diff --check.
Run in an isolated worktree with isolated daemon state/ports, never the user's running broker.
Independent review and controller verification required before merge. No live Figma mutation is needed for this broker-only repair.

jangtrinh:
After a chunked request was admitted, additional frames with the same request ID could still be forwarded to the plugin or appended to queued work. The broker now refuses chunks for IDs retained by jobs or parked requests before buffering or dispatch. Same-connection duplicates leave the original request and its reply intact; other connections receive an explicit duplicate-ID error. Assembled requests must also match their outer chunk ID.

Regression coverage includes running, queued, parked, terminal and outcome-unknown jobs, both connection ownership cases, envelope mismatch, and ordinary admission after history expiry. Independent review found a test-only watchdog override that failed under slower polling; the exact reproduction passes after its removal.

Validation: typecheck, build, all 2,545 tests across 175 files, comment hygiene, and independent focused re-review pass. The daemon harness uses real loopback WebSockets with simulated plugins; live Figma acceptance is separate.

Closes #127.

### DIFF
--- a/cli/src/transport/broker-daemon.ts
+++ b/cli/src/transport/broker-daemon.ts
@@ -1427,22 +1427,16 @@ export async function runBrokerDaemon(options?: BrokerDaemonOptions): Promise<vo
    * and put false metadata in `status`.
    */
   const admitChunk = (from: WebSocket, id: string, rawText: string, last: boolean): void => {
-    // A chunk belonging to an ALREADY-admitted job (running or queued) is a continuation
-    // of a request already fully classified — never re-admitted as a fresh request. The
-    // `existing.from === from` check (stage-4 fix round, minor 6) guards the theoretical
-    // collision or replay: an id received from a DIFFERENT connection must never be
-    // treated as a continuation of this job — it is a fresh, still-unadmitted request on
-    // its own per-connection buffer, where admission can refuse the duplicate safely.
     const existing = st.jobs.byRequestId(id);
-    if (existing !== undefined && existing.from === from) {
-      if (existing.state === 'running') {
-        const entry = st.registry.getByInstanceId(existing.targetInstanceId);
-        if (entry && entry.ws.readyState === WebSocket.OPEN) {
-          try { entry.ws.send(rawText); } catch { /* the running-job error path below catches this on the NEXT frame or the reply timeout */ }
-        }
+    const parked = st.waiting.find((request) => request.id === id);
+    if (existing !== undefined || parked !== undefined) {
+      const owner = existing?.from ?? parked?.from;
+      if (owner === from) {
+        log(`ignored chunk for request id ${id} already owned by ${existing ? `job ${existing.jobId}` : 'parked request'}`);
       } else {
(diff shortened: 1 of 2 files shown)


jangtrinh (commented):
Reviewed commit `405ccd9d357322ad9277baaa14b17b50b3059fa5` against `e033d0decb76b67875c402022343f9cfa31afcba`. The two remote file blobs match the independently reviewed and locally verified artifacts.

**Summary:** The ownership guard runs before buffering, forwarding or appending later chunks. It preserves the legitimate request's reply and rejects mismatched outer/assembled request identities before admission.

**Risk:** Medium, because this changes transport admission. No Critical or Important finding remains in the reviewed scope. Existing job retention, FIFO, pause and outcome-unknown semantics remain intact; this does not claim replay protection beyond retained ownership.

The initial review found a test-only 150ms watchdog override unrelated to the retained-state matrix. Its failure was reproduced with a 500ms harness polling interval. After removing that override, the same reproduction passes and the independent daemon harness passes all 73 tests. Controller verification after the final edit passes typecheck, build, all 2,545 tests, comment hygiene and whitespace checks.

**Verdict: Approve within this scope.** This is posted as a neutral formal review because the authenticated account owns the PR; it is not a separate GitHub approval. CI and live Figma acceptance remain separately reported gates.

PR #131, opened 2026-09-05
fix(sync): settle reconcile completion once

### LINKED ISSUE #129: fix(sync): settle reconcile completion exactly once
When the reconcile kernel executable cannot be launched, Node emits both `error` and `close`. `runReconcile` currently calls its completion callback for each event, so a single sync request can report two failures. A controlled run with a nonexistent executable reproduced two public `spawnReconcileApply` callbacks: the actionable launch error followed by `ui exited -2`.

Implement one settlement boundary shared by launch failure and process closure in `cli/src/transport/figma-sync-apply.ts`. Preserve the first truthful result and existing successful dry-run → capture → apply behavior.

Acceptance:
- A real missing executable produces exactly one public completion after both process events have had an opportunity to fire.
- Event ordering tests cover synchronous spawn throw, error then close, close then late error, malformed output, and normal successful completion. A controlled executable fixture verifies the normal pipeline executes each stage once without a live Figma file or registry mutation.
- The regression fails against the original implementation because it receives two completions, not because a module is missing.
- Typecheck, build, complete test suite, comment hygiene, and independent review pass after the final edit.

Scope: reconcile settlement and focused tests only. Do not alter subprocess deadlines, output limits, network policy, automatic retry, mirror semantics, result schema, or unrelated child runners. Resource and unknown-outcome work remains in the broader hardening plan.

jangtrinh:
When the reconcile executable could not launch, Node emitted both `error` and `close`, causing a single sync request to deliver two failure results. A per-process settlement guard now preserves the first result and completes each reconcile stage once.

Regression tests cover both event orders, synchronous launch failure, apply-stage failure after a successful preview, and malformed/nonzero output. A real missing executable reproduces the original duplicate callback after its observed `close`; a controlled executable verifies dry-run and apply each run once.

Validation on the current main baseline: typecheck, build, all 2,552 tests across 176 files, comment hygiene, and independent review pass. Tests use controlled subprocesses without a live Figma document or registry mutation. Subprocess deadlines and output limits remain separate work.

Closes #129.

### DIFF
--- a/cli/src/transport/figma-sync-apply.ts
+++ b/cli/src/transport/figma-sync-apply.ts
@@ -53,6 +53,12 @@ function runReconcile(
   extra: string[],
   done: (env: Record<string, unknown> | null, err: string, exit: number) => void,
 ): void {
+  let settled = false;
+  const settle = (env: Record<string, unknown> | null, err: string, exit: number): void => {
+    if (settled) return;
+    settled = true;
+    done(env, err, exit);
+  };
   let child;
   try {
     child = spawn(uiBin(), ['figma', 'reconcile', '--dir', projectDir, '--json', ...extra], {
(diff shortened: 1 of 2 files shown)


jangtrinh (commented):
Reviewed commit `9f8b01d5f912119fe65aca27c46845179698f3b1` against `f04a52a0a0cf6e3eece6974ed6c62a20733ad931`. The two remote blobs match the independent review and controller verification.

**Verdict: Approve within scope. Risk: Medium.** No actionable findings remain. The per-process flag is set before invoking the completion callback, preserves the first result, handles both event orders, and leaves the existing error listener installed for late errors. Public result shape and the successful reconcile pipeline remain unchanged.

Independent focused verification passed 21 tests, including a real missing executable observed through `close` and a controlled successful dry-run/apply sequence. Controller verification on the final main baseline passed typecheck, build, all 2,552 tests, comment hygiene and whitespace checks. PR CI Verify also succeeded.

Posted as a neutral formal review because the authenticated account owns the PR; this is not a separate GitHub approval. Live product-kernel integration, subprocess deadlines and output limits are separate from this settlement repair.

PR #137, opened 2026-09-05
fix(import): cancel oversized image streams

### LINKED ISSUE #134: Enforce the existing image size limit while streaming responses
The image inliner has an existing 8 MiB per-image limit, but currently calls response.arrayBuffer() before checking it. An oversized or chunked response is therefore fully buffered before refusal.

Enforce the same limit while consuming response chunks and cancel reading immediately on overflow. Preserve valid images at the exact limit, current HTML/CSS URL replacement, deduplication, timeout, MIME behavior and warning fallback. Report observed bytes honestly; after cancellation the complete response length is unknown. Do not claim a whole-process memory bound.

Acceptance: test-first proof that an over-limit streaming response is cancelled without reading its remaining chunks; missing or misleading Content-Length cannot bypass enforcement; valid small and exact-limit responses remain byte-identical after base64 encoding; failures and aborts leave the original URL and an actionable warning. Include one real local HTTP streaming fixture using ephemeral ports, with server/socket cleanup. Full project gates and independent review after final edits.

Ownership: cli/src/util/inline-images.ts and focused tests/inline-images.test.ts only, plus a narrowly named helper only if needed. No new network/SSRF policy, concurrency limit, aggregate limit, renderer, dependency, config or broker changes in this repair. Those remain required separate parts of plans/260905-0843-plugin-hardening/phase-05-import-and-child-process-bounds.md.

jangtrinh:
Stop downloading an image as soon as observed response bytes exceed the existing 8 MiB limit. The reader cancels the stream before retaining more data or converting it to base64; the fallback warning explicitly identifies the partial byte observation.

Regression coverage includes exact-limit responses, misleading Content-Length, real PNG bytes, overlapping HTML/CSS URLs, stream failures, overflow socket closure and an in-flight response timeout. All 2,595 tests, typecheck, build and comment lint pass; committed plugin artifacts are unchanged.

This change preserves current URL and MIME behavior. Address policy, aggregate import budgets and concurrency limits remain separate work.

Closes #134.

### DIFF
--- a/cli/src/util/inline-images.ts
+++ b/cli/src/util/inline-images.ts
@@ -32,6 +32,35 @@ function collectImageUrls(html: string): string[] {
   return [...urls];
 }
 
+class ImageTooLargeError extends Error {
+  constructor(readonly observedBytes: number) {
+    super(`observed at least ${observedBytes} bytes, exceeding ${MAX_IMAGE_BYTES}`);
+  }
+}
+
+async function readImageResponse(response: Response): Promise<Buffer> {
+  const reader = response.body?.getReader();
+  if (!reader) return Buffer.alloc(0);
+
+  const chunks: Buffer[] = [];
+  let observedBytes = 0;
+  while (true) {
+    const { done, value } = await reader.read();
+    if (done) break;
+    observedBytes += value.byteLength;
+    if (observedBytes > MAX_IMAGE_BYTES) {
+      try {
+        await reader.cancel();
+      } catch {
+        // The oversized response is already being rejected.
+      }
+      throw new ImageTooLargeError(observedBytes);
+    }
+    chunks.push(Buffer.from(value));
+  }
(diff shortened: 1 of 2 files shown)


jangtrinh (commented):
Independent read-only re-review approved the exact source and test snapshot now committed at 2d33da3412118d0c34390a521b90d2387b5bb72a. Prior findings on ambiguous partial-byte wording and missing durable cases were reproduced and resolved. Six focused tests pass, including real PNG bytes, prefix-overlapping URLs, overflow cancellation and a real stalled response with socket closure. Controller separately verified all 2,595 tests, typecheck, build and comment lint; plugin build artifacts are unchanged.

No blocking finding remains in the two-file scope. Aggregate budgets, concurrency and remote-address policy remain separate work. This comment records the independent local review; it is not a second-account GitHub approval.

PR #143, opened 2026-09-05
fix(transport): acquire private runtime snapshots exclusively

### LINKED ISSUE #141: fix(security): exclusively acquire private runtime temporary files
## Problem

Three atomic runtime writers open predictable temporary paths with truncating writes. Real controlled filesystem probes confirmed that a pre-existing symlink redirects writes into another owned file and is then renamed onto the final advertisement, last-plugin or mutation-gate path. Temporary cleanup may also unlink a path the writer never acquired.

## Repair scope

Use a shared exclusive private-file writer for broker-discovery.ts, last-plugins-log.ts and mutation-admission-gate.ts. Open a new file exclusively with mode0600, write by owned descriptor, close before atomic rename, and clean only owned temporary files. Preserve original write errors and fail-closed mutation gate behavior. Existing files, symlinks, dangling links and directories must remain untouched on collision. Keep current runtime paths, temp naming, schemas and orphan sweep behavior; private-directory migration, safe reads, append logs and auth remain separate work.

Files: new cli/src/transport/private-file-write.ts; those three modules; tests/private-runtime-writes.test.ts and focused existing atomic/gate tests as required; existing operations docs owning local state. No dependency, generated artifact, broker message or owner runtime-file mutation.

## Acceptance

1. Test-first actual-filesystem collision matrix for all three writers: sentinel and original target preserved, pre-existing path not removed, failure surfaced or gate unavailable. Happy path produces a regular0600 file with complete JSON and no temporary residue.
2. Inject partial write, close and rename failures: close owned descriptors, remove only owned temporary files, preserve original failure and old complete target. Keep existing atomic sequence and gate transition/restart tests meaningful.
3. Controlled fresh and replacement writes under permissive umask still produce0600; no claim of protection from same-UID processes or network filesystem semantics.
4. Focused tests, typecheck, build, full tests, comment lint and diff check pass; independent review and root final rerun after all edits.

No live default path write or modification of owner data is needed to verify the fix.

jangtrinh:
Create broker advertisement, last-plugin and mutation-gate temporary files with exclusive descriptor access and mode `0600`. A pre-existing file or symlink now refuses the write instead of truncating a referenced file; successful writes still close before atomic rename. Failed acquisition leaves the collision intact, and failure cleanup removes only a temporary file acquired by the current write.

Real-filesystem regressions cover regular files, directories, symlinks and dangling symlinks, existing-target preservation, permissive umask, partial-write and close failures, and rename failure. The independent review reproduced 77 focused tests. The final controller checks passed all 2,619 tests, typecheck, build and comment lint; generated plugin artifacts are unchanged.

This repair covers the three snapshot writers. Existing parent-directory permissions, read integrity and other runtime stores remain separate work.

Closes #141.

### DIFF
--- a/cli/src/transport/broker-discovery.ts
+++ b/cli/src/transport/broker-discovery.ts
@@ -4,7 +4,7 @@
 // polling until advertised.
 import { spawn } from 'node:child_process';
 import { dirname, basename, join } from 'node:path';
-import { readdirSync, readFileSync, renameSync, statSync, unlinkSync, writeFileSync } from 'node:fs';
+import { readdirSync, readFileSync, renameSync, statSync, unlinkSync } from 'node:fs';
 import { fileURLToPath } from 'node:url';
 import WebSocket from 'ws';
 import {
@@ -17,6 +17,7 @@ import {
 } from '../../../shared/protocol.ts';
 import { CliError, envMs } from './protocol-helpers.ts';
 import { safeCleanup } from '../../../shared/safe-cleanup.ts';
+import { writePrivateFileExclusive } from './private-file-write.ts';
 
 // Mirrors broker-daemon.ts's own `HEARTBEAT_MS` override (same env var, same
 // fallback) — see envMs's doc for why this must be the ONE shared reader, not a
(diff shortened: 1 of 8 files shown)


jangtrinh (commented):
The independent read-only review found no blocking issue in the eight-file change and reproduced all 77 focused filesystem tests. Existing collisions and prior target contents are preserved, successful files are regular `0600` snapshots, and the atomic-write test verifies the actual descriptor sequence through rename. The controller separately reran the full 2,619-test suite, typecheck, build and comment lint against the reviewed file hashes.

The close-failure tests prove the close attempt and error handling; they do not establish that an operating system descriptor is always closed after a real close syscall error. Parent-directory migration and read integrity remain outside this repair.

This comment records the independent review evidence. It is not a separate-account GitHub approval.

Issue #144, opened 2026-09-05
Capture reconcile executable identity and qualify supported kernel selection


jangtrinh:
Bare-name `ui` lookup inherits the broker's PATH and can select a different kernel than the one reported in a plain shell. Under npm, the exact panel-gate dependency `ease-design@0.5.0` provides a local `ui` binary ahead of the global 0.6.0 installation. Reconcile evidence currently omits the executable identity, making compatibility failures easy to misdiagnose.

Controlled reproduction: the explicitly identified local 0.5.0 binary exits 0 after 65,536 bytes of incomplete JSON on each 1k/10k preview/apply fixture. The explicitly identified global `/opt/homebrew/bin/ui` 0.6.0 returns complete valid JSON in all four cases through both independent native Node spawn and the production bounded runner:261,090–8,862,779 bytes. The original report incorrectly attributed the bare-name failure to 0.6.0; executable identity now proves npm PATH shadowing. No package inspection bypass or global install change was needed. The old stdout-drain defect was repaired upstream in https://github.com/jangtrinh/design-os/issues/209.

Scope: resolve and capture the exact executable once per reconcile run using its actual cwd, PATH and existing FIGMA_AGENT_UI_BIN / DESIGN_OS_UI_BIN override precedence; use that same path for preview and apply, and include it in successful and failed SyncApplyEvidence. Preserve existing executable selection policy: do not silently reject all npm-local binaries, rewrite PATH, bump the exact panel dependency, invoke a shell, add a stdout-spooling workaround, or change a global installation. Provide a short supported-install diagnostic and explicit override instruction in owning operations docs.

Acceptance: test-first PATH shadowing and explicit relative/absolute override behavior, missing executable/no-start evidence, same selected executable for both phases, and recorded identity equal to the spawned command. On POSIX honor executable permission and cwd-relative PATH entries; preserve supported platform behavior. Two actual Node-pipe controls already pass for global0.6.0; repeat exact-identity production integration after the source edit. Incomplete preview must not start apply; ambiguous apply retains its capture. Run focused tests, full suite, typecheck, build and comment lint; independent review before merge. Work in a separate plugin worktree based on the merged bounded-child repair. No sibling-product or live owner state mutation.

Issue #140, opened 2026-09-05
fix(sync): prevent surviving reconcile children overlapping after broker restart


jangtrinh:
## Problem

The broker currently exits without tracking reconcile children. A restart resets the in-memory sync lane while an old apply may still write the same registry. Sending a signal or restarting the broker does not prove old writes stopped. This remains a separate failure mode after bounded reconcile settlement.

## Scope and design gate

Scout the bounded child runner, daemon shutdown, per-project sync ownership and existing durable mutation recovery before implementation. Design graceful shutdown cancellation plus crash/restart admission that cannot silently overlap a surviving apply. Avoid stale PID reuse or trusting a PID file as proof; a lease must bind project and process identity. Consider a durable unknown latch requiring evidence-based recovery rather than claiming arbitrary orphan termination is safe. No kernel/registry write implementation moves into this broker.

## Acceptance

Reproduce a controlled child surviving parent termination, then prove a restarted broker refuses another apply until the original is confirmed stopped or a deliberate recovery protocol completes. Cover graceful TERM/KILL, abrupt broker death, PID reuse/stale records, cleanup failure and explicit unknown result. Log every retained/deleted ownership record. Preserve read/poll traffic and model-free operation. Use isolated temporary project/advertisement paths and owned PIDs only. Run focused/full gates plus independent review and controller verification.

This follow-up is mandatory for crash-safe full sync qualification; bounded replies alone do not close it. No owner state may be reset or deleted automatically.

Issue #139, opened 2026-09-05
fix(sync): bound mirror scan child processes after measured live baseline


jangtrinh:
## Problem

`figma-mirror-capture-run.ts` passes a30s request timeout to scan-node, but the parent has no child deadline and buffers stdout/stderr without limits. A hung process after a reply can hold sync indefinitely. This remains after the reconcile-only repair.

## Scope and dependency

After the shared bounded child runner lands, adopt it for mirror scan subprocesses. Own mirror capture runner, its focused tests and owning operations docs. Preserve live scan targeting, pending-first ordering, MAX_SCANS40 and honest dropped/skipped counts. Do not change registry/kernel ownership or cap an unmeasured valid scan reply.

## Acceptance

Measure actual scan replies on a designated scratch Figma file and at least one large representative component before selecting output bounds; record fixture identity and byte count. Use actual source-backed request/connect timing to set a parent deadline, TERM/KILL and closure cleanup. Tests reproduce hang, output flood, timeout, nonzero exit, UTF-8 split, cancel and malformed reply; successful captures remain compatible. Report partial scan outcomes and retained evidence accurately. Align or explicitly document overall batch timing against panel30s stuck indication; no silent automatic retry. All repository gates, independent review, one live scan proof and controller rerun required.

No arbitrary mutation or owner canvas probing is authorized by this issue. A concrete candidate and designated scratch file are prerequisites for the live gate.

PR #146, opened 2026-09-05
fix(sync): bound reconcile children and preserve uncertain outcomes

### LINKED ISSUE #138: fix(sync): bound reconcile children and preserve unknown outcomes
## Problem and outcome

Reconcile currently retains unlimited child output, has no parent deadline, and accepts `ok:true` even when the preview exits nonzero. An actual controlled child reproduced exit7 followed by an apply. Buffer chunks are decoded independently, so split UTF-8 names can be corrupted. Apply failures also delete the capture before uncertainty is classified.

Bound the reconcile child lifecycle while preserving exactly-once completion. A deadline initiates TERM, then KILL; it does not prove the child exited. Never free the broker sync lane for an unconfirmed-live child. Never advance from a failed preview or report applied counts from uncertain apply output.

## Implementation contract

- Add a small shared bounded child runner, adopt it in reconcile, and keep broker changes confined to sync-lane ownership. Track spawn, exit and close separately; pre-spawn failure is known no-start. A post-spawn error alone is not exit evidence. Keep late errors harmless and release timers/listeners without uncaught error events.
- Internal injectable defaults: stdout128MiB; stderr1MiB head/tail;120s deadline;2s TERM grace;1s close grace after exit;5s reap grace after KILL. These are provisional hang/memory safeguards, not workload SLOs. Real installed kernel on controlled10k records emitted8,862,779apply bytes; an8MiB stdout cap would reject that case.
- Count observed bytes independently. Stop retaining stdout beyond its cap but keep draining under the deadline. Decode retained Buffer bytes once, never per chunk. Stderr head/tail keeps both ends and total/truncated counters.
- Success requires confirmed exit0, no signal, no timeout or stdout truncation, and a valid success envelope. Dry nonzero exit must never capture/apply. A possibly started apply with missing or bad evidence is UNKNOWN: no `applied`/`landed`, no automatic retry. Always propagate `childExited:false` before other failure classifications so the broker holds its lane.
- Delete a capture only after a known successful apply or known no-start. Retain it in its existing private temporary directory on failed/unknown apply, especially while the child may still read it; include bounded outcome evidence and capture path. This adds no permanent project store or retention period. Do not recommend restarting as proof that an unkillable child stopped.

## Ownership and non-goals

Allowed: cli/src/transport/bounded-child-process.ts (small cohesive helpers if needed); figma-sync-apply.ts; broker-daemon.ts sync lane only; tests/bounded-child-process.test.ts; tests/figma-sync-apply.test.ts; tests/broker-daemon-harness.test.ts focused sync cases; existing docs/operations.md owning sync text.

No kernel/registry writes, panel redesign, mirror-scan migration, authentication, dependencies, live broker/Figma or unrelated source. Mirror-scan children and broker restart/orphan recovery are separately tracked follow-ups and remain explicit limitations.

## Acceptance

1. RED then GREEN: preview valid JSON + nonzero exit never enters apply; apply valid JSON + nonzero exit returns UNKNOWN without counts; split multibyte UTF-8 preserves names; spawn throw/error+close remains exactly once.
2. Deterministic small injected bounds cover independent byte caps, drain totals, stderr head/tail, timeout, ignored TERM/KILL, false kill return, close without exit, exit without close, late events, unconfirmed exit and timer cleanup. Real controlled children cover hang and TERM refusal; dispose every fixture PID.
3. Capture matrix: success/no-start removes; failed/unknown retains and reports path; unconfirmed-live never removes. Broker harness proves no overlapping sync and lane release only after confirmed settlement. Tests must assert childExited:false survives every classification.
4. Real installed deterministic kernel on a temporary controlled fixture retains dry/apply compatibility; no owner artifacts or live Figma. Existing9MiB valid reply remains accepted under defaults.
5. Focused tests, typecheck, build, full tests, comment lint, diff checks; separate independent review and controller rerun after final edit. No weakened tests or comment-baseline changes.

jangtrinh:
Bound reconcile subprocess output and lifetime, and require reliable exit evidence before accepting a successful JSON envelope. Nonzero previews no longer start apply. Timeouts, truncated output and uncertain apply completion preserve the capture for inspection; an unconfirmed child exit keeps the current broker's sync lane closed. Cleanup failures remain visible without changing a known successful apply result.

The runner counts all drained bytes, retains at most 128 MiB stdout and 1 MiB stderr, and escalates from a 120-second deadline through TERM and KILL. UTF-8 fragments cannot combine into invented characters, and decoded diagnostics stay within their bound. These are provisional engineering limits, not supported-workload promises.

Validation: test-first lifecycle, output, capture and broker-lane regressions; independent read-only review passed 105 focused tests; controller passed all 2,644 tests, typecheck, build and comment lint. Generated plugin artifacts are unchanged.

Compatibility: explicit `/opt/homebrew/bin/ui` 0.6.0 produced complete valid JSON in all four controlled 1k/10k preview/apply cases through both this runner and independent plain Node spawn (261,090–8,862,779 stdout bytes). A negative control identified the npm-local `ui` 0.5.0 binary, which truncates those same outputs at 65,536 bytes. An earlier draft attributed that failure to 0.6.0 without capturing executable identity; this has been corrected. Runtime executable identity and supported-install diagnostics remain #144. Mirror scan adoption and restart/orphan ownership remain #139 and #140. The current sync-lane hold does not survive broker restart.

Closes #138.

### DIFF
--- a/cli/src/transport/bounded-child-output-buffer.ts
+++ b/cli/src/transport/bounded-child-output-buffer.ts
@@ -1,0 +1,104 @@
+import { StringDecoder } from 'node:string_decoder';
+
+export interface BoundedOutputSnapshot {
+  text: string;
+  totalBytes: number;
+  retainedBytes: number;
+  truncated: boolean;
+}
+
+type RetentionMode = 'prefix' | 'head-tail';
+
+/** Omit an incomplete trailing code point without inventing a replacement character. */
+function completePrefix(buffer: Buffer): string {
+  return new StringDecoder('utf8').write(buffer);
+}
+
+function completeTail(buffer: Buffer): string {
+  let start = 0;
+  while (start < buffer.length && (buffer[start]! & 0xc0) === 0x80) start++;
+  return buffer.subarray(start).toString('utf8');
+}
+
+function bytes(chunk: unknown): Buffer {
+  if (Buffer.isBuffer(chunk)) return chunk;
+  if (typeof chunk === 'string') return Buffer.from(chunk);
+  return Buffer.from(String(chunk));
+}
+
+/** Retain either a prefix or non-overlapping head/tail while counting every drained byte. */
+export class BoundedOutputBuffer {
+  private totalBytes = 0;
+  private retainedBytes = 0;
+  private truncated = false;
+  private readonly prefix: Buffer[] = [];
+  private head = Buffer.alloc(0);
+  private tail = Buffer.alloc(0);
+
+  constructor(private readonly limit: number, private readonly mode: RetentionMode) {}
+
+  push(chunk: unknown): void {
+    const buffer = bytes(chunk);
+    const nextBytes = this.totalBytes + buffer.length;
+    if (this.mode === 'prefix') {
+      const available = Math.max(0, this.limit - this.retainedBytes);
+      if (available > 0) {
+        const retained = Buffer.from(buffer.subarray(0, available));
+        this.prefix.push(retained);
+        this.retainedBytes += retained.length;
+      }
+      this.totalBytes = nextBytes;
+      this.truncated = nextBytes > this.limit;
+      return;
+    }
+
+    const headBytes = Math.ceil(this.limit / 2);
+    const tailBytes = Math.floor(this.limit / 2);
+    if (!this.truncated && nextBytes <= this.limit) {
+      const retained = Buffer.from(buffer);
+      this.prefix.push(retained);
+      this.retainedBytes += retained.length;
+    } else if (!this.truncated) {
+      const previous = Buffer.concat(this.prefix);
+      const neededHead = Math.max(0, headBytes - previous.length);
+      this.head = previous.length >= headBytes
+        ? Buffer.from(previous.subarray(0, headBytes))
+        : Buffer.concat([previous, Buffer.from(buffer.subarray(0, neededHead))]);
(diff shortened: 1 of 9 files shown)


jangtrinh (commented):
Independent read-only code review recorded for commit 5f714783dc82cc17fb91b23ac71beed9efa1c678: no blocking finding in the nine-file diff. The reviewer verified the frozen file hashes before and after running 105 focused tests. Controller separately passed all 2,644 tests, typecheck, build and comment lint after the final edit.

The review checked drained-byte versus retained-byte accounting, UTF-8 fragment boundaries, timeout and signal escalation, late errors, confirmed child exit, capture retention and cleanup failures, and the broker lane hold after unconfirmed exit.

This is a review record posted by the PR author's account, not a separate GitHub approval. Compatibility correction after executable identity was captured: the truncating bare-name command was npm-local ui 0.5.0, not global 0.6.0. Controller's absolute-path global 0.6.0 tests now pass all four controlled 1k/10k preview/apply cases through both production bounded runner and native Node spawn, returning complete JSON up to 8,862,779 bytes. A negative control with the identified local 0.5.0 executable reproduces all four 65,536-byte failures. Earlier version attribution was incorrect; PR body and issue144 have been corrected. Runtime executable identity diagnostics remain #144. No live Figma acceptance is claimed. Mirror adoption and restart/orphan recovery remain #139 and #140.

PR #153, opened 2026-09-05
fix: write binding restart cache privately and report failures

### LINKED ISSUE #149: Protect restart binding cache with private atomic writes and visible failures
The broker restart cache contains local project paths but currently writes directly to its destination with the process umask. A failed or interrupted write can replace a complete cache with partial data, and write errors are silently ignored. The cache is reconstructible; project binding markers remain the authority.

Implement a focused repair using the existing exclusive0600 temporary-file writer followed by atomic rename. Preserve the cache location, v1 schema, deduplication, binding precedence, and best-effort semantics. Do not alter project-owned binding markers, live default state, retention policy, authentication, or path identity. Separate the cache owner from project-bind.ts if that keeps each responsibility small; preserve existing exports and imports.

Return a structured cache-write result. The daemon records a bounded warning on failure; the bind command reports that its durable marker succeeded but restart-cache persistence failed, without falsely claiming the binding itself failed. Never silently overwrite or remove an existing temporary collision. On failure preserve the prior destination and only clean a temporary file this invocation acquired. A successful atomic replacement may replace a destination symlink itself but must never write its referent; no parent-directory or same-UID security claim.

Test-first with owned temporary fixtures: permissive umask still0600, complete replacement, existing temporary file, directory, symlink and dangling-symlink collisions; write, close and rename refusal; destination sentinel unchanged on failure; structured warning observable at daemon and CLI boundaries; no stdout pollution of JSON. Existing project-bind and daemon behavior must stay compatible. Run focused tests, typecheck, build, full tests, comment lint, independent read-only review and controller rerun. Record a real filesystem write, read, mode and cleanup run; never touch the default broker cache or owner project data.

jangtrinh:
Binding restart hints now use an exclusively acquired private temporary file and atomic replacement. A colliding temporary is preserved, and write, close or rename failures preserve the prior cache. The cache location, schema and durable project binding marker remain unchanged.

The daemon records a bounded filesystem cause when the restart cache cannot be updated. The bind command keeps its successful binding result and adds a machine-readable warning only on a real cache failure. Existing cache state is not migrated; parent-directory integrity and same-user access are outside this write guarantee.

Validation: 2,677 tests pass after integration with current main, including actual filesystem collision and native failure cases; TypeScript checks, bundle generation, comment lint and whitespace checks pass. An owned Darwin filesystem replay confirms mode0600, replacement, deduplication, collision preservation and cleanup. Independent review covers the complete nine-file change.

Closes #149.

### DIFF
--- a/cli/src/commands/bind.ts
+++ b/cli/src/commands/bind.ts
@@ -110,7 +110,7 @@ export async function run(args: CommandArgs): Promise<unknown> {
   const marker = readBindMarker(projectDir) ?? { v: 1 as const, bindings: [] };
   const entry = upsertBinding(marker, fileName, fileKey);
   writeBindMarker(projectDir, marker);
-  writeBindCache([...readBindCache().projectDirs, projectDir]);
+  const cacheWrite = writeBindCache([...readBindCache().projectDirs, projectDir]);
 
   return {
     projectDir,
@@ -120,5 +120,9 @@ export async function run(args: CommandArgs): Promise<unknown> {
     migratedCount,
     migratedEditCount,
     marker: bindMarkerPath(projectDir),
+    ...(!cacheWrite.ok ? { warnings: [{
+      code: 'E_BIND_CACHE_WRITE', cause: cacheWrite.error.code,
+      message: 'Binding saved, but restart cache was not updated. Inspect the cache directory and reported filesystem cause before retrying bind.',
+    }] } : {}),
   };
 }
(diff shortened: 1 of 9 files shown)


jangtrinh (commented):
Independent local review: APPROVE for the complete nine-file change at9fdd1c110532cdc579c0a5196ffeb861630d5080. No Critical or Important findings. Exclusive private acquisition, collision ownership, preservation on native write/close/rename failures, bounded warnings and existing binding contracts were checked.

The integrated review reran51 focused tests and verified9/9 source hashes before and after. Controller validation separately passed2,677 tests plus TypeScript, bundle generation, comment lint and whitespace gates on the final source. An owned filesystem replay verified mode0600, replacement, deduplication, collision preservation and cleanup. Committed blobs match all9 frozen hashes.

This COMMENT records the independent local review because the PR and review submission share a GitHub account; it is not approval from a second GitHub account. Parent-directory integrity, same-user access, existing-cache migration and read-path hardening remain outside this scoped write guarantee.

Issue #154, opened 2026-09-05
test: investigate resource harness job-state correlation race


jangtrinh:
The retained-resource daemon harness failed once at `tests/broker-resource-inspection-harness.test.ts:161`, observing one queued job where two were expected. Three isolated reruns and a final complete focused rerun passed. The observer-startup failure branch under review was not entered in that scenario.

The helper `sendJob` currently accepts any JOB_STATE with a job ID, so a delayed update from another request may resolve the wrong wait. This is a source-backed hypothesis, not a reproduced root cause. Reproduce the ordering deterministically before changing the helper, correlate the response to the request using real protocol data, and retain every lifecycle assertion. Do not solve the race with sleeps, retries, weaker counts or product behavior changes. Keep production resources, quotas, auth and live Figma out of scope.

Acceptance: a deterministic regression fails on the current helper, the corrected helper accepts only the intended job state, and the focused resource and daemon tests pass without relaxing assertions. This follow-up is recorded for later scheduling; no new implementation is started as part of the current scope review.

PR #155, opened 2026-09-05
fix(broker): close listeners when observer startup fails

### LINKED ISSUE #150: Close owned broker listeners when resource observer registration fails
The optional in-process resource observer is a trusted diagnostics hook. Its registration currently runs after WebSocket listeners bind but before shutdown cleanup is defined. If that callback throws and an embedding process catches the startup rejection, the listeners can remain until process exit. This is a bounded embedding cleanup concern; the observer is not exposed on the wire and default broker startup has no observer.

Before broadening this hook beyond the controlled measurement harness, make observer-registration failure close every listener acquired by this startup attempt, retain the original error, and avoid publishing a successful advertisement. Preserve ordinary no-observer startup and at-most-once teardown. Tests must inject the registration refusal using owned ephemeral listeners, verify rejected startup, closed ports, no advertisement, and no leftover timers. Independent review and repository gates required. Do not change auth, protocol, resource budgets or owner runtime state.

jangtrinh:
When the optional synchronous resource observer throws during broker startup, the broker now closes its acquired WebSocket listeners before rejecting with the original error. Failed startup cannot leave those ports occupied, and does not publish an advertisement or install ordinary lifetime timers and signal handlers.

Validation: 2,678 tests pass after integration with current main, including a regression that failed against the previous implementation. TypeScript checks, bundle generation, comment lint and whitespace checks pass. An owned Darwin replay confirms IPv4 and IPv6 listener closure, exact-port reuse, the original error, and scratch cleanup. Independent review approved the bounded change; an unrelated intermittent resource-harness assertion is tracked separately in #154.

This repair covers the synchronous observer registration boundary. It does not claim rollback of every possible startup failure or cleanup of external resources created by an observer.

Closes #150.

### DIFF
--- a/cli/src/transport/broker-daemon.ts
+++ b/cli/src/transport/broker-daemon.ts
@@ -627,16 +627,28 @@ export async function runBrokerDaemon(options?: BrokerDaemonOptions): Promise<vo
   };
   let resourceObserverTeardown: (() => void) | null = null;
   if (options?.resourceObserver) {
-    const teardown = options.resourceObserver(() => collectBrokerResourceSnapshot({
-      pendingChunks: st.pendingChunks,
-      parkedRequestFrames: st.waiting.map((request) => request.rawText),
-      jobTable: st.jobs.resourceSnapshot(),
-      queues: st.queues,
-      pendingRequestReferenceCount: st.pending.size,
-      dispatchedRequestReferenceCount: st.dispatchedTo.size,
-      dispatchReservationReferenceCount: st.dispatchReservations.size,
-    }));
-    if (typeof teardown === 'function') resourceObserverTeardown = teardown;
+    try {
+      const teardown = options.resourceObserver(() => collectBrokerResourceSnapshot({
+        pendingChunks: st.pendingChunks,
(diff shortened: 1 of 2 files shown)


jangtrinh (commented):
Independent local review: APPROVE for synchronous observer startup cleanup. No Critical or Important source findings. The review checked closure of every acquired listener before preserving the original rejection, the absence of success advertisement and lifetime hooks, and unchanged ordinary startup/teardown behavior.

The controller verified that the reviewed additions and removals are identical after integration with current main, reran all 2,678 tests and the TypeScript/build/comment/whitespace gates, and verified both committed blobs at d19855646eca94f8817cb2f06f639d5470a4d04d against the final frozen hashes. A fresh owned Darwin replay closed and rebound both IPv4 and IPv6 listeners. Cleanup-rejection aggregation and accepted-client termination have source review, but no dedicated runtime fixture.

One unchanged resource-harness assertion failed during initial review and passed four subsequent runs. It is recorded in #154; it is not hidden by a retry setting or weakened assertion.

This COMMENT records an independent local review because the PR and review submission share a GitHub account; it is not approval from a second GitHub account.

PR #158, opened 2026-09-05
fix(broker): validate and release request chunks by connection

### LINKED ISSUE #152: fix: validate request chunks before retention and record cleanup
# Validate request chunk metadata and order before retention

The broker buffers raw CLI chunk frames before validating their sequence or `last` field. A malformed or duplicate sequence can therefore retain data until completion or sweep. Closing a CLI also leaves its partial chunks until the sweep, and abandonment removes retained data without a record.

Preserve the current protocol, request ownership, FIFO, result retention, and timeout behavior. Before retaining a request chunk, require a nonnegative safe-integer sequence, a boolean `last`, a string payload, and sequence equal to the next expected index. Refuse malformed metadata or missing/repeated sequence with a correlated existing error and release only that connection's affected request. Keep the existing ownership guard first so late frames cannot corrupt an admitted request's reply. Release partial chunks immediately on disconnect; record numeric counts and UTF-8 bytes for refusal, disconnect, and existing sweep cleanup. Preserve empty fragments, Unicode, current producer sizes, unrelated requests, and fresh same-id assembly after cleanup.

Scope: broker request admission, a cohesive transport helper if useful, focused pure and daemon tests, and the smallest owning operations documentation. No new quotas, deadlines, id-length constraints, authentication, plugin reply policy, result TTL changes, or new wire error codes. Resource ceilings need separate workload evidence and product decisions.

Acceptance: meaningful tests fail before the fix; tests prove rejection before retention, per-connection isolation, owner guard precedence, exact cleanup accounting, valid Unicode assembly, and continued healthy traffic. Run typecheck, bundle generation, complete tests, comment lint, whitespace checks, independent review, and an owned actual daemon replay with cleanup verified. Never use the active broker or owner Figma documents.

jangtrinh:
Validate request chunk metadata and exact sequence before retaining frames. Malformed or out-of-order input releases only the affected connection/request, while the existing owner guard protects already admitted jobs and valid request/reply routing remains unchanged.

Connection close now immediately releases its partial chunks even when that socket subsequently registered as a plugin. Close and stale-sweep cleanup record numeric request, frame and UTF-8 byte counts without logging payloads or paths. The existing sweep stays wire-silent so a stale partial cannot terminate a legitimate admitted request using the same ID.

Validation includes real daemon regressions for metadata, ordering, Unicode and empty fragments, duplicate ownership, same-ID sweep preservation, ordinary disconnect and role-promotion disconnect. The role-promotion regression failed against the pre-fix implementation. No new quota, retention deadline, authentication policy or wire format is introduced.

Final validation: 2,686 tests pass, along with TypeScript checks, bundle generation, comment lint and whitespace checks. A fresh owned daemon run passed eight fixture/pressure/recovery scenarios and confirmed process, port and scratch cleanup.

Closes #152.

### DIFF
--- a/cli/src/transport/broker-daemon.ts
+++ b/cli/src/transport/broker-daemon.ts
@@ -21,8 +21,9 @@ import {
   cleanupOrphanedAdvertisementTempFiles, isPidAlive, readAdvertisement, selfBuildMtime, writeAdvertisement,
 } from './broker-discovery.ts';
 import {
-  ChunkAssembler, deleteConnectionChunk, envMs, getConnectionChunks, isChunkMsg, isEventMsg, isReplyMsg, isRequestMsg,
-  parseWireMsg, rawToString, sendWireMsg, sweepAbandonedChunks, type ChunkBuffers,
+  ChunkAssembler, deleteConnectionChunk, deleteConnectionChunks, envMs, getConnectionChunks, isChunkMsg, isEventMsg,
+  isReplyMsg, isRequestMsg, parseWireMsg, rawToString, requestChunkCandidateId, sendWireMsg, summarizeChunkCleanups,
+  sweepAbandonedChunks, validateRequestChunkFrame, type ChunkBuffers, type ChunkCleanup,
 } from './protocol-helpers.ts';
 import { PluginRegistry, appReadiness, suspectedZombie, type PluginEntry } from './plugin-registry.ts';
 import { createSessionTallies, type SessionTallies } from './session-tallies.ts';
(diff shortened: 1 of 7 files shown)


jangtrinh (commented):
Independent local review: APPROVE for the final seven-file request-chunk change at ccbe7039aa9676c17322187ec53e093963489428. The final review reran 98 tests across four actual test files, verified every frozen hash, and confirmed that disconnect cleanup covers both ordinary clients and sockets promoted to plugins. No Critical or Important code finding remains.

Controller validation separately passed 2,686 tests and all project gates on the final source, and a fresh eight-case owned daemon replay confirmed functional recovery and process/port/scratch cleanup. The role-promotion regression failed before the repair and passed afterward. The final reviewer did not locate the abbreviated red/green log names; the controller verified the complete artifacts and the pre-fix expected-zero/actual-one failure. No claim is made that this reviewer independently reran the pre-fix state.

All seven committed blobs match the reviewed manifest. This COMMENT records an independent local review because the PR and review submission share a GitHub account; it is not a second-account approval. No quota, authentication or new retention policy is included.

Issue #157, opened 2026-09-05
Make real child timeout fixtures verify startup readiness


jangtrinh:
The integrated full suite observed one failure in `tests/bounded-child-process.test.ts:192`: the real TERM fixture's PID file was absent after the 2,500 ms deadline. The runner reported spawned=true, timedOut=true, exited=true and signal=SIGTERM, with no launch error. The isolated eight-test file subsequently passed unchanged. This test and the bounded-child implementation are unchanged by the renderer work.

The fixture writes the PID from inside Node after the OS spawn event. Under the observed run, timeout/termination preceded that fixture marker. Scheduling/startup delay is a plausible explanation; it has not been reproduced deterministically. The observation alone does not establish a product timeout regression.

Task: make the real TERM/KILL lifecycle fixture verify its startup prerequisite deterministically, while preserving the assertions for actual termination, escalation, exactly-once completion and absence of a live child. Do not increase production deadlines, add blanket retries, remove PID/death checks or weaken the existing tests.

Acceptance: reproduce the missing-prerequisite condition in a controlled test before fixing it; demonstrate TERM and TERM-refusing/KILL behavior with real child processes; run the focused file and full suite. Keep failure artifacts and report limits of the proof.

Observed evidence: full suite 2,733 passed/1 failed; subsequent isolated diagnostic 8/8 passed. This is a separate test-reliability follow-up, not scope added to the personal renderer feature.

Issue #156, opened 2026-09-05
Qualify personal Figma workflows and compare completed-task cost


jangtrinh:
Owner decision 2026-09-05 16:14 (+07): visual demo canceled; do not resume it. The CLI exception question is withdrawn, not granted. Historical browser/MCP proof remains; plugin comparison and renderer live checks remain unrun. Existing broader qualification work is deferred. Local demo server stopped; files and Figma test page retained.

Qualify the personal Figma workflow on an explicitly designated scratch file and compare completed tasks with the official Figma MCP route. The goal is evidence for performance, accuracy and cost, using the existing plugin architecture.

Scope: inspect a frame; edit text/layout; batch repeated edits; components/variables; import/render. Use the same starting document, requested final properties, agent/model, undo requirement and output quality on both routes. Internal batching may differ; record the observed calls. Unsupported overlap must be reported explicitly.

Acceptance:
- Record scratch file identity, source fixture/snapshot and final node/property/reference checks. No mutation of an unrelated working file.
- Qualify the reviewed HTML and shader artifacts in Figma: message source/origin, ordinary import/bake, supported small/large render, error cleanup and actual visual result. Browser tests remain separate evidence.
- Record cold/warm elapsed time, tool calls, payload bytes, every failure/retry, and actual token usage when supplied by the runtime. Missing usage remains unavailable; bytes do not become invented token counts or per-call prices.
- Compare successful tasks with equivalent quality before claiming speed or cost improvements. Do not assert that one route wins universally.
- Restore or remove only test-owned content and preserve the owner's plugin/runtime setup.

[...]

PR #147, opened 2026-09-05
fix(render): isolate HTML extraction in an opaque iframe

jangtrinh:
Move HTML extraction into an opaque `sandbox="allow-scripts"` iframe. Repository JavaScript and runtime CSS remain available inside the child, while parent-document access and forged panel control messages are refused. The panel accepts only the active child's correlated reply and validates its payload before import. Every terminal path releases its timers, listeners and iframe; child-derived errors retain their typed refusal with a 512-character message bound.

The compiled renderer is generated with the panel, including escaped script delimiters. Browser regressions exercise parent access, control spoofing, malformed payloads, repeated replies, setup refusals, timeout cleanup, runtime animation-frame styles and artifact freshness. Actual Tailwind CDN extraction passed with computed padding and color preserved.

Validation: independent code review approved the renderer and corrections; controller passed 2,697 tests, typecheck, build and comment lint after integrating main at 937fe1b12cec75577e609bebee3b82d0d7137eaf. The actual compiled-panel probe returned one bounded typed refusal, zero mutation forwards, and accepted genuine parent messages. This does not provide CPU or process isolation.

Draft gate: source/origin and import behavior in the supported Figma host remains unrun. Browser evidence does not replace live Figma qualification. Shader isolation and authentication are separate work. Keep this PR unmerged until the live host checks pass.

Closes #133.

PR #159, opened 2026-09-05
fix(renderer): isolate shader rendering and validate PNG input

### LINKED ISSUE #142: fix(renderer): isolate shader child and bound PNG admission
## Problem and outcome

The shader renderer creates an unsandboxed iframe and accepts replies by token alone. Imported renderer code can access panel DOM, and a different window with the token can forge a result. PNG data URLs are decoded without encoded/decoded size checks; the main executor reconstructs arbitrary numeric object values before a native image decode. The load guard can also remain armed after other completion paths.

Isolate the shader child, strictly bind replies to that child, and bound/validate image input before allocation and native decoding. Preserve the pinned ShaderGradient renderer and actual visual output.

## Dependency and scope

Stack after the opaque HTML/shared parent guard candidate, issue133. Own plugin/src/ui/gradient-host.ts; new shared gradient image admission helper if needed; plugin/src/main/executor-gradient.ts; relevant focused gradient tests; generated plugin/code.js and plugin/ui.html; existing operations docs. No auth, broker, dependency upgrades, manifest domain expansion, panel redesign, placeholder images or renderer version changes. Other workers are not to be reverted.

## Implementation constraints

- sandbox allow-scripts only, opaque origin. Require actual event.source===child.contentWindow, origin null and exact per-render token; ignore wrong/stale/duplicate replies. Treat child fields as unknown, bound failure text/code at the host. All load/render timers, listeners and iframe resources terminate on success, refusal, timeout and failure.
- Use fixed in-viewport placement at top-left with opacity0 and pointer-events none. Actual prior experiments proved an offscreen opaque iframe stalls requestAnimationFrame despite CDN200; normal and240x44 rail experiments succeed with in-viewport placement. Do not reintroduce allow-same-origin, visibility:hidden or an unsupported settle watchdog.
- Validate finite positive dimensions/scale and existing4096pixel side contract before iframe allocation. Validate PNG data URL prefix/base64 length and canonical encoding before atob; validate observed decoded size and PNG signature/header dimensions before forwarding. Main applies the same bounded contract before converting numeric objects/arrays and before getNodeById or createImage; preserve valid Uint8Array/array/numeric-key legacy representations but refuse gaps, extra keys and invalid byte values. Native Figma decode remains the final image decoder; do not claim a partial header check proves full PNG validity.
- Derive encoded/decoded limits from measured actual canvas/renderer outputs and the supported4096x4096pixel envelope; an8MiB image-fetch cap is not evidence for a shader cap. Record existing480x300 and compact-rail proof plus a representative larger render before selecting final defaults. Do not silently remove supported dimensions or presets.

## Test and delivery gates

Test-first actual browser proof for sandbox isolation, forged sibling/parent/wrong-origin/token replies, bounded error, cancellation/timeout/cleanup and successful PNG. Main tests prove invalid image data cannot call native image creation or modify fills; use valid PNG fixture and realistic refusing mocks. Preserve pinned module URL/config and real CDN checks separately from ordinary CI. Run the actual production host with pinned modules on normal and compact rail viewports, verify nonempty valid output, dimensions and no leftover iframe.

Run focused gradient tests, typecheck, build, full tests, comment lint and diff checks; independent review and controller rerun after final edit. Live Figma candidate proof is a separate pre-merge gate shared with issue133: exact main source, raw file identity, read-only shader self-test and one designated scratch-node bake. No arbitrary owner canvas mutation, no live PASS from Chromium alone.

jangtrinh:
Render ShaderGradient inside an opaque `sandbox="allow-scripts"` child. The host accepts only the active child's matching origin and correlated result, and removes listeners, timers and the iframe after success, failure or timeout. Keeping the hidden child inside the viewport preserves animation-frame rendering in the compact panel.

PNG admission now checks dimensions, byte bounds, canonical base64 and PNG headers before decoding or touching the Figma scene. Typed-array messaging remains direct; valid legacy dense arrays and numeric-keyed objects remain supported. Figma's native image decoder still decides full PNG validity. Renderer dependencies, existing deadlines, presets and undo behavior remain unchanged.

Validation: 2,734 tests and all project gates pass on the final integrated candidate. Browser regressions cover parent isolation, spoofed replies, lifecycle cleanup and malformed images. A fresh production-only CDN replay passed normal and compact 480x300, 2048x2048 and 4096x4096 rendering, with no page errors or leftover iframe. One unrelated child-fixture startup assertion failed in an earlier full run; the unchanged isolated test and subsequent full runs pass, and the observation is tracked in #157.

This PR is stacked on the opaque HTML renderer change in #147. Keep it draft until actual Figma source/origin, self-test, scratch bake and visual checks pass. Browser measurements do not establish live Figma latency, process isolation or a cost advantage over MCP. The shared live workflow comparison is tracked in #156.

Closes #142.

### DIFF
--- a/cli/src/commands/shader-gradient.ts
+++ b/cli/src/commands/shader-gradient.ts
@@ -15,7 +15,7 @@ import { SHADER_GRADIENT_PRESETS } from '../../../shared/shader-gradient-presets
  * Pinned renderer, recorded onto the node so a stale bake stays identifiable.
  * The PUBLISHED version — upstream's package.json at the read revision says 2.4.24,
  * which was never released and 404s everywhere. Keep this in step with
- * RENDERER_VERSION in plugin/src/ui/gradient-host.ts.
+ * RENDERER_VERSION in plugin/src/ui/gradient-render-document.ts.
  */
 const RENDERER = '@shadergradient/react@2.4.20';
 
--- a/docs/operations.md
+++ b/docs/operations.md
@@ -124,6 +124,23 @@ separate headroom for the existing 8 MiB image producer's base64 output. These a
 budgets bound validation and forwarding; they do not isolate arbitrary renderer JavaScript
 or establish a whole-process memory limit.
 
+## Shader renderer boundary
+
+ShaderGradient runs in an opaque `allow-scripts` child so renderer code cannot share the
+panel origin. The boundary and its exact-source lifecycle live in
+[`gradient-host.ts`](../plugin/src/ui/gradient-host.ts); the pinned renderer document stays
+separate in [`gradient-render-document.ts`](../plugin/src/ui/gradient-render-document.ts).
+The child remains in the panel viewport at zero opacity because Chromium suspends animation
+frames for an opaque offscreen child; hiding it with `visibility:hidden` has the same failure.
+
+PNG admission is shared by the UI and main executor in
+[`gradient-image-admission.ts`](../shared/gradient-image-admission.ts). Its shader-specific
+limit is derived from the supported 4096px RGBA envelope plus PNG/zlib and ancillary-chunk
+headroom; it is intentionally separate from HTML image-fetch budgets. Encoded and decoded
+ceilings bound allocation; signature and IHDR checks bound the image shape, while Figma's
+native decoder remains the final decodability check. `figma-agent shader-gradient --self-test` is read-only, but only a live
+supported Figma host can qualify WebGL availability.
+
 ## Runtime snapshot writes
 
 Advertisement, last-plugin and mutation-gate snapshots acquire temporary files exclusively,
--- a/plugin/code.js
+++ b/plugin/code.js
@@ -7344,20 +7344,92 @@
     return { sourceId: source.id, targetId: target.id, traits, applied, skipped };
   }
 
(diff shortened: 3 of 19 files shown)


jangtrinh (commented):
Independent local review: APPROVE for the final 19-file shader change at 51fd1b892a0963e0b094e335cb9c929152c43eae. Earlier review checked opaque-child isolation, exact result correlation, PNG admission, cleanup and generated artifacts with 93 passing focused tests. Final review verified the fixture corrections with 33 passing tests, clean standalone whitespace checks and all 19 matching hashes. No Critical or Important code finding remains.

Controller validation passed 2,734 tests and all project gates after the final edit. Committed blobs match the reviewed manifest. The production-only CDN replay remains applicable because renderer runtime bytes are unchanged; all four cases through 4096x4096 passed. One unchanged child-fixture startup failure during an earlier full run is recorded in #157; subsequent unchanged focused and full runs passed.

Actual Figma source/origin, self-test, scratch bake and visual acceptance remain unrun. Keep this stacked PR draft until those checks pass. Browser evidence does not establish live Figma latency, process isolation or savings over MCP.

This COMMENT records an independent local review because the PR and review submission share a GitHub account; it is not a second-account approval.

